PR #26, opened 2019-06-14
Use Ed25519 signature scheme for identities

Wollac:
- Switch from Secp256k1 to Ed25519:
    - No additional hashing before signing
    - Public key and Signature are added to the message to retain the same functionality
- Remove go-etherium dependency
- Refactor identity handling to be more flexible with the signature scheme


capossele:
Needs further changes. We are working on this as part of the autopeering refactoring.

PR #48, opened 2019-07-24
Dockerfile volume path

nuriel77:
* Use a more pleasant background color for the status screen
* Don't user root directory in the dockerfile. Easier for anyone wanting to use the dockerimage without user root.

### DIFF
--- a/plugins/statusscreen/statusscreen.go
+++ b/plugins/statusscreen/statusscreen.go
@@ -49,7 +49,7 @@ func run(plugin *node.Plugin) {
 	content.SetColumns(0)
 
 	footer := newPrimitive("")
-	footer.SetBackgroundColor(tcell.ColorDarkMagenta)
+	footer.SetBackgroundColor(tcell.ColorDarkBlue)
 	footer.SetTextColor(tcell.ColorWhite)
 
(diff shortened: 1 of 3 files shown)


plugins/statusscreen/statusscreen.go
```diff
@@ -49,7 +49,7 @@ func run(plugin *node.Plugin) {
 	content.SetColumns(0)
 
 	footer := newPrimitive("")
-	footer.SetBackgroundColor(tcell.ColorDarkMagenta)
+	footer.SetBackgroundColor(tcell.ColorDarkBlue)
```
capossele: If you really want to change the colors then it should be something that anybody can change. Maybe try adding the color as a parameter

nuriel77: Ok, but this is as far as my golang experience goes 😂

hmoog: I like the idea of having "themes" - we can just add a public variable for the colors and a 2nd plugin can then just change these Public colors from another plugin

So you can have stattusscreen-dark-theme or sth as a plugin and people can change to whatever they want - for the scope of this PR can you remove the color changes?

PR #50, opened 2019-07-25
Update README.md

martingravel:
For Windows user:

- `shimmer` must be renamed to `shimmer.exe`

- To launch shimmer, `shimmer` instead of `./shimmer`


martingravel:
Sorry, I did a mistake on my comment, the last line must be this one:

To launch shimer, `shimmer` instead of `./shimmer`

Issue #40, opened 2019-07-22
Crashs when using the spammer


Citrullin:
I don't know if this issue is already known, so I open an issue for it. Feel free to close when this is already known.

Environment:
- ARM64 (ARMv8, Scaleway)
- Ubuntu 18.04.2 
- 4.9.93-mainline-rev1
- go version go1.12.7 linux/arm64

Steps to reproduce
1. Build shimmer ```go build -o shimmer```
2. Start shimmer ```./shimmer```
3. Run the spammer in a web-browser ```http://[YOUR_IP]:8080/spammer?cmd=start&tps=10000```
4. Let it run for some minutes. It will crash with:
```github.com/dgraph-io/badger.(*DB).Update(0x4001098700, 0x4032b67eb0, 0x0, 0x0)
                                                                /root/go/pkg/mod/github.com/dgraph-io/badger@v2.0.0-rc.2.0.20190610165348-a804ffa35fb9+incompatible/txn.go:696 +0x80
                    github.com/iotaledger/goshimmer/packages/database.(*databaseImpl).Set(0x40019e4580, 0x4009eaee40, 0x51, 0x0, 0x4000024c00, 0x5dc, 0x5dc, 0x51, 0x403fe67500)
                        /root/go/src/github.com/iotaledger/goshimmer/packages/database/database.go:75 +0x80
                           github.com/iotaledger/goshimmer/plugins/tangle.storeTransactionInDatabase(0x4009e88780, 0x51, 0x400b469800)
                                                        /root/go/src/github.com/iotaledger/goshimmer/plugins/tangle/transaction.go:91 +0xdc
                                                           github.com/iotaledger/goshimmer/plugins/tangle.onEvictTransaction.func1(0x4009e88780)
                                                                        /root/go/src/github.com/iotaledger/goshimmer/plugins/tangle/transaction.go:64 +0x28
                                                                           created by github.com/iotaledger/goshimmer/plugins/tangle.onEvictTransaction
                                                                        /root/go/src/github.com/iotaledger/goshimmer/plugins/tangle/transaction.go:63 +0x74
                                                                           root@```

PR #69, opened 2019-08-01
Improved []byte<->string conversion

Wollac:
- Restored test and benchmarks
- Use more efficient version for `BytesToString`
- Make `StringToBytes` safe under current unsafe rules


Wollac:
@capossele 
Can you still check whether there is no regression with #40 on ARM?

PR #79, opened 2019-08-26
Distance function should use salt

capossele:
This PR fixes a bug in the distance function that was not using the salt


capossele:
Still needs changes. Closing for now

PR #80, opened 2019-08-28
UI and webauth plugins

Evanfeenstra:
adds web UI and authentication plugins. Has one extra dependency: dgrijalva/jwt-go. If this gets merged in I will definitely maintain and add features :)


hmoog:
Can you embed the css files and the additional templates and so on as strings in go files? Then it's possible to still deploy the node as a standalone executable.

Other than that - I like it :D

Wollac:
Thanks a lot for the contribution!

PR #81, opened 2019-09-02
Removed badger import in packages and plugins

muXxer:


muXxer:
I like to keep my external packages separate :cry:
Oh, you did it on your own :+1:

PR #85, opened 2019-11-14
Removes parameter/events packages in favor of hive.go's

luca-moser:
Config:
* All CLI flags have been adjusted to use `pflags` and keys are normalized to fit in a JSON format. 
All default values have been left intact.
* A `config.json` in the root directory contains all node configuration settings.
* Defining a setting via a CLI flag overrides what is defined in the config. 
For example doing `--database.directory=testnetdb`
* Thanks to the config.json, it is now apparent what kind of settings GoShimmer actually has.

Packages:
* The `events` package has been replaced with the one from hive.go
* The `parameter` package has been replaced with the one from hive.go


packages/node/node.go
```diff
@@ -1,6 +1,7 @@
 package node
 
 import (
+	"github.com/iotaledger/hive.go/parameter"
```
Wollac: File is not `goimports`-ed

plugins/bundleprocessor/bundleprocessor_test.go
```diff
@@ -1,12 +1,14 @@
 package bundleprocessor
 
 import (
+	"github.com/iotaledger/hive.go/parameter"
```
Wollac: File is not `goimports`-ed

plugins/autopeering/server/tcp/server.go
```diff
@@ -1,34 +1,37 @@
 package tcp
 
 import (
```
Wollac: File is not `goimports`-ed

plugins/gossip/neighbors.go
```diff
@@ -1,6 +1,6 @@
 package gossip
 
-import (
+import(
```
Wollac:
```suggestion
import (
```

plugins/analysis/webinterface/httpserver/plugin.go
```diff
@@ -1,12 +1,12 @@
 package httpserver
 
-import (
+import(
```
Wollac:
```suggestion
import (
```

plugins/cli/plugin.go
```diff
@@ -3,82 +3,62 @@ package cli
 import (
 	"flag"
```
Wollac: This should probably changed to `pflag` as well, so that the provided `printUsage` is displayed.

plugins/cli/cli.go
```diff
@@ -3,6 +3,7 @@ package cli
 import (
 	"flag"
```
Wollac: Shouldn't that be changed to `pflag`?

plugins/cli/cli.go
```diff
@@ -53,6 +42,6 @@ func printUsage() {
 	)
 	flag.PrintDefaults()
 
-	fmt.Fprintf(os.Stderr, "\nThe following plugins are enabled by default and can be disabled with -%s:\n  %s\n", getFlagName(node.DISABLE_PLUGINS.Name), getList(enabledPlugins))
-	fmt.Fprintf(os.Stderr, "The following plugins are disabled by default and can be enabled with -%s:\n  %s\n\n", getFlagName(node.ENABLE_PLUGINS.Name), getList(disabledPlugins))
+	fmt.Fprintf(os.Stderr, "\nThe following plugins are enabled: %s\n", getList(parameter.NodeConfig.GetStringSlice(node.CFG_ENABLE_PLGUINS)))
```
Wollac: When a plugin is created in the code it takes a parameter whether it is enabled by default. The idea of these two lines in the previous version was to loop through all the plugins registered in GoShimmer and display them here and say whether they are enabled/disabled (to get this information, they are added to `enabledPlugins` `disabledPlugins`). That way you could see all the available plugins and what actually makes sense to enable/disable as a parameter.

I guess it makes sense to have that behavior again. An alternative would be to remove the enabled-flag from the plugin itself completely and only use CFG with default values for that. However, that way, when a new plugin is written that is not added to the defaults, the user couldn't see it and its name in the usage screen.

plugins/autopeering/parameters/parameters.go
```diff
@@ -1,11 +1,21 @@
 package parameters
 
-import "github.com/iotaledger/goshimmer/packages/parameter"
+import (
+	"flag"
```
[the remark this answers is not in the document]
luca-moser: Changed.

packages/node/parameters.go
```diff
@@ -1,10 +1,17 @@
 package node
 
-import "github.com/iotaledger/goshimmer/packages/parameter"
-
-var (
-	LOG_LEVEL = parameter.AddInt("NODE/LOG_LEVEL", LOG_LEVEL_INFO, "controls the log types that are shown")
+import (
+	"flag"
```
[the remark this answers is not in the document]
luca-moser: Fixed.

packages/node/node.go
luca-moser: Fixed.

plugins/bundleprocessor/bundleprocessor_test.go
luca-moser: Fixed.

plugins/cli/cli.go
luca-moser: Changed

plugins/cli/plugin.go
luca-moser: Changed

plugins/cli/cli.go
luca-moser: I've reverted it back to the two old print outs (just with the adjusting flag). Note that because the config file can also hold `node.enablePlugins/node.disablePlugins` but is only loaded after `flag.Parse()` is called, its defined enabled/disabled plugins are not part of the usage output.

Wollac: I guess that is acceptable!

PR #93, opened 2020-01-08
Updates to the new logger in hive.go

luca-moser:
This PR ensures that all loggers are only initialized after `InitGlobalLogger()` was called. The config file, respectively parameters, are now parsed before `node.Run()` is called in order to have the Viper config ready to pass it into `InitGlobalLogger()`. 

It turns out that `InjectWriters()` isn't needed anymore in the status_screen plugin, as it automatically overrides stdout in the background.


Wollac:
Made sure that when the status screen is not running or has crashed, everything is logged to console.

PR #139, opened 2020-01-17
:sparkles: Implement relay-check

### LINKED ISSUE #108: Test whether actual gossiping works between nodes
* Make a program which sends out transactions using the new API and then check after a cooldown whether the transactions are actually query-able on the other non listed nodes. 
* Make the other nodes to check configurable via a simple config.json file
* Make the interval in which the program issues a transaction configurable
* Program code is going to be located in a `tool` folder and is called: `relay-checker`

jkrvivian:
What `relay-check` does:
1. Broadcast txn to target node
2. Wait for a cooldown time
3. Check if target node can find the broadcasted txn
4. Check if other nodes can find the broadcasted txn

Close #108


tool/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "relaycheck": {
+    "targetnode": "http://127.0.0.1:8080",
+    "neighbor": "http://127.0.0.1:8080",
```
capossele: What do you think if instead of only 1 neighbor we allow for multiple ones (they don't even need to be "neighbors")?
In that case, we could automatically check whether txs do get propagated over the network.

tool/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// query target node for broadcasted data
+	_, err := api.GetTransactions([]trinary.Hash{txnHash})
+	if err != nil {
+		fmt.Printf("Relay-Check: get transaction from local failed.\n")
+		return true
+	}
+	return false
+}
+
+func testNeighborGetTransactions(txnHash trinary.Hash) bool {
```
luca-moser: Return an actual error from within this func, then print it out from the caller.

tool/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	client "github.com/iotaledger/goshimmer/client"
+	"github.com/iotaledger/iota.go/trinary"
+)
+
+func testBroadcastData(api *client.GoShimmerAPI) (trinary.Hash, bool) {
```
luca-moser: move the print outs to the caller of the func

tool/main.go
```diff
@@ -0,0 +1,65 @@
...
+	txnHash, err := api.BroadcastData(txnAddr, txnData)
+	if err != nil {
+		fmt.Printf("Relay Check: broadcast data failed.\n %s\n", err)
+		return "", true
+	}
+	fmt.Printf("Relay Check: txnHash: %s\n", txnHash)
+	return txnHash, false
+}
+
+func testTargetGetTransactions(api *client.GoShimmerAPI, txnHash trinary.Hash) bool {
```
luca-moser: same, return an error and move the print outs to the caller

tool/parseConfig.go
```diff
@@ -0,0 +1,46 @@
+package main
```
luca-moser: rename the file to `config.go`

tool/parseConfig.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func LoadConfig() {
+	if err := parameter.FetchConfig(true); err != nil {
+		panic(err)
+	}
+}
+
+func SetConfig() {
+	if parameter.NodeConfig.GetString("relaycheck.targetNode") != "" {
```
luca-moser: invert if case to remove the unnecessary else statement

tool/parseConfig.go
```diff
@@ -0,0 +1,46 @@
...
+		target = parameter.NodeConfig.GetString("relaycheck.targetNode")
+	} else {
+		panic("Set the target node address\n")
+	}
+	if parameter.NodeConfig.GetString("relaycheck.neighbor") != "" {
```
luca-moser: invert if case to remove the unnecessary else statement

tool/parseConfig.go
```diff
@@ -0,0 +1,46 @@
+package main
+
+import (
+	"github.com/iotaledger/goshimmer/packages/parameter"
+)
+
+var (
```
luca-moser: this parameters should be defined as `pflags` settings, as it is done within plugins and packages.

tools/config.go
```diff
@@ -0,0 +1,48 @@
...
+func SetConfig() {
+	if parameter.NodeConfig.GetString("relaycheck.targetNode") == "" {
+		panic("Set the target node address\n")
+	}
+	target = parameter.NodeConfig.GetString("relaycheck.targetNode")
+
+	if len(parameter.NodeConfig.GetStringSlice("relaycheck.nodes")) == 0 {
+		panic("Set node addresses\n")
+	}
+	for _, n := range parameter.NodeConfig.GetStringSlice("relaycheck.nodes") {
```
golangcibot: S1011: should replace loop with `nodes = append(nodes, parameter.NodeConfig.GetStringSlice("relaycheck.nodes")...)` (from `gosimple`)

tools/relay-checker/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	client "github.com/iotaledger/goshimmer/client"
+	"github.com/iotaledger/goshimmer/packages/errors"
+	"github.com/iotaledger/iota.go/trinary"
+)
+
+func testBroadcastData(api *client.GoShimmerAPI) (trinary.Hash, error) {
+	txnHash, err := api.BroadcastData(txnAddr, txnData)
+	if err != nil {
+		return "", errors.Wrapf(err, "Broadcast failed")
```
Wollac: Use `fmt.Errorf`and `%w` to wrap errors

Wollac:
Do we need a super short readme or something or is this mainly just for us?

PR #146, opened 2020-01-20
Removes error package in favor of stdlib's

luca-moser:
* There are no more instances of IdentifiableErrors.
* Made some explicit error vars for errors which were produced inline.
* According to "Whether to wrap" in https://blog.golang.org/go1.13-errors, I adjusted some wrappings to not wrap the origin error, as it comes from an implementation detail (for example from the badger db)
* Replaced `Wrap()` or `Derive()` instances with `fmt.Errorf()` and wrapping the error if it made sense.
* Got rid of any instance of `github.com/pkg/errors` within GoShimmer directly. Note that it is still a dependency.


packages/model/error.go
```diff
@@ -0,0 +1,8 @@
+package model
+
+import "errors"
+
+var (
+	ErrUnmarshalFailed = errors.New("unmarshal failed")
+	ErrMarshalFailed  = errors.New("marshal failed")
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

PR #155, opened 2020-01-20
Migrates to database package of hive.go

### LINKED ISSUE #130: Implement a database version flag
We want to panic if someone starts the node with an incompatible version of the database and instruct the user to delete it. With every breaking change, we adjust the database version flag.

luca-moser:
* Use v2.0.1 of BadgerDB
* Every database now uses a byte prefix which is defined in `prefixes.go` inside the `packages/prefixes.go` file
* Everytime the node starts up, now a flag is checked within the database to ensure that it is compatible with the given version of GoShimmer. If the schema of stored data somehow inflicts a breaking change, the database version can be adjusted to force users to delete their database folder (this is ok, because GoShimmer is just a prototype, people should never assume that their data persists over version jumps). (Old databases which were made pre v0.1.0 are not compatible and therefore also there a crash message instructing the user to delete the database is printed out.)
* Note that the code uses an own function of getting the database instance and not the one defined in hive.go (like Hornet). Later we can perhaps change the function within hive.go and have a common function which can be used for both in a sensible way.

Closes #130


tools/api_works/main.go
```diff
@@ -0,0 +1,82 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/iotaledger/goshimmer/client"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	goshimmer "github.com/iotaledger/goshimmer/client"
```

packages/database/versioning.go
```diff
@@ -0,0 +1,49 @@
...
+// also automatically sets the version if the database is new.
+func checkDatabaseVersion(dbIsNew bool) {
+	dbInstance, err := database.Get(DBPrefixDatabaseVersion, instance)
+	if err != nil {
+		panic(err)
+	}
+
+	if dbIsNew {
+		// store db version for the first time in the new database
+		if err := dbInstance.Set(database.Entry{Key: dbVersionKey, Value: []byte{DBVersion}}); err != nil {
```
golangcibot: shadow: declaration of "err" shadows declaration at line 25 (from `govet`)

packages/autopeering/peer/peerdb.go
```diff
@@ -157,25 +158,25 @@ func parseInt64(blob []byte) int64 {
...
+	return db.db.Set(database.Entry{Key: key, Value: blob, TTL: peerExpiration})
 }
 
 // LocalPrivateKey returns the private key stored in the database or creates a new one.
 func (db *persistentDB) LocalPrivateKey() (PrivateKey, error) {
-	key, err := db.db.Get(localFieldKey(dbLocalKey))
+	entry, err := db.db.Get(localFieldKey(dbLocalKey))
 	if err == database.ErrKeyNotFound {
-		key, err = generatePrivateKey()
+		key, err := generatePrivateKey()
```
golangcibot: shadow: declaration of "err" shadows declaration at line 177 (from `govet`)

packages/autopeering/peer/peerdb.go
```diff
@@ -157,40 +158,39 @@ func parseInt64(blob []byte) int64 {
...
 			err = db.UpdateLocalPrivateKey(key)
 		}
 		return key, err
 	}
 	if err != nil {
 		return nil, err
 	}
-
-	return key, nil
+	return []byte(entry.Value), nil
```
Wollac:
```suggestion
	return PrivateKey(entry.Value), nil
```

packages/database/database.go
```diff
@@ -1,134 +1,65 @@
...
 )
 
 var (
-	ErrKeyNotFound = badger.ErrKeyNotFound
-
-	dbMap = make(map[string]*prefixDb)
-	mu    sync.Mutex
+	instance *badger.DB
+	once     sync.Once
 )
```
Wollac: Add something like the following, so that it is no longer necessary to also include `hive.go/database` each time the database is used in goshimmer:
```Go
type Database = database.Database

func Get(dbPrefix byte, optionalBadger ...*badger.DB) (Database, error) {
	return database.Get(dbPrefix, optionalBadger...)
}
```

packages/database/database.go
```diff
@@ -1,134 +1,65 @@
...
-	return this.forEach(append(this.prefix, prefix...), consumer)
+		// up on the first caller, check whether the version of the database is compatible
+		checkDatabaseVersion(dbDirClear)
+	})
+	return instance
 }
 
-func (this *prefixDb) ForEach(consumer func([]byte, []byte)) error {
-	return this.forEach(this.prefix, consumer)
+func CleanupGoShimmerBadgerInstance() {
```
Wollac: This function is not used. Should it be deleted?

packages/database/versioning.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/iotaledger/hive.go/database"
+)
+
+const (
+	// defines which version of the database schema this version of GoShimmer supports.
```
Wollac:
```suggestion
	// DBVersion defines the version of the database schema this version of GoShimmer supports.
```

packages/database/versioning.go
```diff
@@ -0,0 +1,49 @@
...
+	// everytime there's a breaking change regarding the stored data, this version flag should be adjusted.
+	DBVersion byte = 1
```
Wollac:
```suggestion
	DBVersion = 1
```
Untyped constant should be preferred.

packages/database/database.go
```diff
@@ -1,134 +1,65 @@
...
-		db:     badger,
-		name:   name,
-		prefix: getPrefix(name),
-	}
-
-	dbMap[name] = db
-
-	return db, nil
-}
+func GetGoShimmerBadgerInstance() *badger.DB {
```
Wollac: Consider renaming to `GetBadgerInstance()`

packages/database/database.go
luca-moser: This doesn't work unfortunately really, because I still need to import the `hive.go` database package as I can't type alias the `ErrKeyNotFound` which is virtually used by every file which also initiates a database.

packages/database/database.go
luca-moser: Can't, as aliasing the database package type doesn't work completely. Having the function named like this gives a better indication when the wrong database package is used.

packages/database/database.go
luca-moser: Made a background worker which calls it every 5 minute. See https://github.com/dgraph-io/badger#garbage-collection

packages/database/versioning.go
luca-moser: Fixed

packages/autopeering/peer/peerdb.go
luca-moser: Fixed

packages/database/database.go
Wollac: I don't fully understand, shouldn't the following be sufficient to not import `hive.go/database` throughout goshimmer:
```go
var ErrKeyNotFound = database.ErrKeyNotFound
type (
	Database     = database.Database
	Entry        = database.Entry
	KeyOnlyEntry = database.KeyOnlyEntry
)
func Get(dbPrefix byte, optionalBadger ...*badger.DB) (Database, error) {
	return database.Get(dbPrefix, optionalBadger...)
}
```

packages/database/database.go
Wollac: Where is it called from exactly?

Wollac: Can we somehow replace the `Println` with `log.Info` at least when the logger is initialized?

packages/database/database.go
Wollac: Sure, this should only be renamed when database package is aliased

packages/database/database.go
luca-moser: Yeah, that's how I changed it, the logger is passed into the function.

packages/database/database.go
luca-moser: You're obviously right. I must have tried to use `type ErrKeyNotFound` 🤦‍♂

packages/database/database.go
Wollac: ah sorry, I thought you didn't change that.
Perfect then

packages/database/database.go
luca-moser: Changed

plugins/tangle/solidifier_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/iotaledger/goshimmer/packages/database"
+	goshimmerDB "github.com/iotaledger/goshimmer/packages/database"
```
Wollac: Remove unnecessary import alias

luca-moser: Removed

Wollac: But you didn't push :stuck_out_tongue_winking_eye:

Wollac: I'll fix this so that we can merge

PR #157, opened 2020-01-21
:bug: Get correct plugin names by removing spaces

jkrvivian:
Plugin names contain no spaces, thus we need to replace spaces to get the correct plugin names from `config.json`.

### DIFF
--- a/plugins/cli/plugin.go
+++ b/plugins/cli/plugin.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/iotaledger/goshimmer/packages/parameter"
 	"github.com/iotaledger/hive.go/events"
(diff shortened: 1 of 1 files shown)


jkrvivian:
Yes, exactly, this PR will fix that. :strong:

Wollac:
I am not sure I understand:
- If `webapigttaendpoint` is put in the config.json it can be activated
- But when entering `WebAPI getTrytes Endpoint` id does not work
Is that the your issue, @jonastheis

Wollac:
It used to be that `node.disablePlugins` that was a space separated list, but now its properly comma separated (or json if you use the config.json).
That's probably why we removed all the spaces in the names :sweat_smile: 
If you do `--help` it even gives you all the names of the plugins in that format.

jkrvivian:
> It used to be that `node.disablePlugins` that was a space separated list, but now its properly comma separated (or json if you use the config.json).
> That's probably why we removed all the spaces in the names sweat_smile
> If you do `--help` it even gives you all the names of the plugins in that format.

Ah! okay! Since I input the plugin name which is passed as a parameter in `xxx/plugin.go`, did not aware of `--help` messages. :stuck_out_tongue: 
@Wollac ,then I think we could close this PR?

Wollac:
Well, I agree it is super confusing. As it does not hurt let's go without the spaces.
@jkrvivian maybe just use `node.GetPluginIdentifier` to be consistent and we can merge this.

plugins/cli/plugin.go
```diff
@@ -37,10 +37,10 @@ func init() {
 
 func parseParameters() {
 	for _, pluginName := range parameter.NodeConfig.GetStringSlice(node.CFG_DISABLE_PLUGINS) {
-		node.DisabledPlugins[strings.ToLower(pluginName)] = true
+		node.DisabledPlugins[strings.ToLower(strings.Replace(pluginName, " ", "", -1))] = true
```
Wollac: Use `node.GetPluginIdentifier`

PR #168, opened 2020-01-23
:recycle: removes goshimmer error package

capossele:
Fix the error thing


luca-moser:
Did the same, but with some different error messages: https://github.com/iotaledger/goshimmer/pull/169

PR #180, opened 2020-01-24
Fix: Buffer gossip messages before passing them to ProtoBuf

Wollac:
Introduces buffconn.BufferedConnection to simplify this process.

### DIFF
--- a/packages/netutil/buffconn/buffconn.go
+++ b/packages/netutil/buffconn/buffconn.go
@@ -15,7 +15,13 @@ const (
 	// MaxMessageSize is the maximum message size in bytes.
 	MaxMessageSize = 4096
 	// IOTimeout specifies the timeout for sending and receiving multi packet messages.
-	IOTimeout = 5 * time.Second
+	IOTimeout = 4 * time.Second
+)
+
+// Errors returned by the BufferedConnection.
+var (
+	ErrInvalidHeader      = errors.New("invalid message header")
+	ErrInsufficientBuffer = errors.New("insufficient buffer")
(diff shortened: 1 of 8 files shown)


packages/netutil/buffconn/buffconn.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := c.conn.SetReadDeadline(time.Now().Add(IOTimeout)); err != nil {
+		return 0, fmt.Errorf("error while setting timeout: %w", err)
+	}
+	return c.read(buffer[:msgLength])
+}
+
+func newHeader(msgLength int) []byte {
+	// the header only consists of the message length
+	header := make([]byte, headerSize)
+	binary.BigEndian.PutUint32(header, uint32(msgLength))
```
luca-moser: Since a message can be max 4kb, we can reduce this to an uint16?

Wollac: My reasoning for this was that maybe 65535 bytes could be a bit small in the future. If you reserve 32bits for the size in the header it might be easier to increase the size in the future without the need to change the header.
I don't know. What do you think does it make change to reserve this or not?

luca-moser: Aight, fine by me then, after all it is just 2 bytes more. 
I just didn't really know we'd have single messages over 65kb in the future.

Wollac: was thinking of maybe atomic transactions :stuck_out_tongue_winking_eye:

PR #214, opened 2020-01-30
Adds single-page-application dashboard

luca-moser:
Contains:
* TPS chart
* Memory chart
* Neighbors chart
* Explorer for transactions and addresses

All assets for the SPA are automatically within the compiled binary, so that no additional files need to be shipped. There's a readme under the `spa/frontend` folder describing how to pack the static assets via packr2.

Basically a stripped down version of the Hornet dashboard with a backend adjusted to work with GoShimmer.

image


packages/gossip/manager.go
```diff
@@ -155,6 +148,14 @@ func (m *Manager) getAllNeighbors() []*Neighbor {
...
 }
 
+func (m *Manager) getNeighbors(ids ...peer.ID) []*Neighbor {
+	if len(ids) > 0 {
+		return m.getNeighborsById(ids)
+	}
+	return m.GetAllNeighbors()
+}
+
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

PR #225, opened 2020-02-04
Disable Statusscreen Plugin

centercirclesolutions:
Web Dashboard is now default status screen


luca-moser:
We'll actually remove it completely

PR #228, opened 2020-02-04
Removes status screen plugin

luca-moser:
As we now have the webinterface showing interesting data, we no longer rely on the status screen plugin. Additionally, the logger now writes directly to stdout.


config.json
```diff
@@ -44,6 +44,7 @@
     "disableStacktrace": false,
     "encoding": "console",
     "outputPaths": [
+      "stdout",
       "goshimmer.log"
     ],
     "disableEvents": false
```
Wollac: the status screen was the only plugin using the logging events. Do we want to disable them here and in the default settings?

luca-moser: sure

jonastheis: I will use them for #211 
But this new plugin also shouldn't be enabled by default. So I guess it's fine to do that and when the remote log feature is activated this needs enabling too then.

PR #249, opened 2020-02-11
Fix: keep offline master peers in active peer list

### LINKED ISSUE #246: Do not remove offline master peers
- Even when offline the provided master should never be removed from the peer list.
- When offline, it should not be returned in `verifiedPeers`

jkrvivian:
* Add a `masters` mpeer list to record master peers
* Implement `wrapPeers` to initiate `masters`
* If a master peer is offline
   1. set the `verifiedCount` to 0 => it will not be included in `GetVerifiedPeers`
   2. move it to the front of active list => avoid reverifying the same offline node

Close #246

### DIFF
--- a/packages/autopeering/discover/manager.go
+++ b/packages/autopeering/discover/manager.go
@@ -33,6 +33,7 @@ type manager struct {
 	mutex        sync.Mutex // protects active and replacement
 	active       []*mpeer
 	replacements []*mpeer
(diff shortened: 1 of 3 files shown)


jkrvivian:
@Wollac , I should open this PR in hive.go, right?

jkrvivian:
Due to #250 , this PR will go to hive.go.

Issue #282, opened 2020-03-05
Binary merge inconsistencies


jonastheis:
There are some packages that are conflicting either within Goshimmer or were already migrated to hive.go.

- [x] Node identity vs. local identity: here some more work is necessary
- [x] goshimmer/netutil vs. **hive.go/netutil**
- [x] goshimmer/database vs. **hive.go/database**

PR #283, opened 2020-03-05
Migrate database and netutil package to hive.go

jonastheis:
Relates #282


jonastheis:
Dependency to hive.go needs to be updated once iotaledger/hive.go#73 is merged.

PR #302, opened 2020-03-19
Feat: Use autopeering update

Wollac:
Do be merged after https://github.com/iotaledger/hive.go/pull/79

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -14,8 +14,7 @@
     "entryNodes": [
       "MRCJTGX9x+PPiT3um1DQSHXvALaUg/tCH/oQr6mliGo=@35.246.104.243:14626"
     ],
-    "port": 14626,
-    "version": 1
+    "port": 14626
   },
   "dashboard": {
     "bindAddress": "127.0.0.1:8081",
--- a/packages/gossip/server/server.go
+++ b/packages/gossip/server/server.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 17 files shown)


packages/gossip/server/server.go
```diff
@@ -46,10 +47,10 @@ var dialRetryPolicy = backoff.ConstantBackOff(500 * time.Millisecond).With(backo
 
 // TCP establishes verified incoming and outgoing TCP connections to other peers.
 type TCP struct {
-	local      *peer.Local
-	publicAddr net.Addr
-	listener   *net.TCPListener
-	log        *zap.SugaredLogger
+	local    *peer.Local
+	port     int
```
golangcibot: field `port` is unused (from `unused`)

config.default.json
```diff
@@ -14,8 +14,7 @@
     "entryNodes": [
       "MRCJTGX9x+PPiT3um1DQSHXvALaUg/tCH/oQr6mliGo=@35.246.104.243:14626"
     ],
-    "port": 14626,
-    "version": 1
```
capossele: if we decide to add the networkID (by splitting the current `version`) we could set it here

Wollac: If the version number is connected to some other parameter like network-ID or Coordinator address, then I agree. But right now it would just be a parameter that you have to set to the right version or everything breaks. That was why I've removed it form the config for now.

capossele: I meant something like instead of having 4 bytes for the protocol version we could have a field `networkID` (2 or 4 bytes) and reduce the `version` field to 1-2 bytes

Wollac: Do you think we should maybe remove the `version int` but introduce an `networdID string`?

capossele: the version field could still be there just for defining the protocol version. Maybe 1 byte is enough for that? 256 possible different versions? For the `networkID` I like the idea of using `string`

Wollac: unfortunately, there is no `byte` type in protobuf, so we would have to use `uint32` or `int32`. But it uses variable length encoding, so it should not be much longer than a byte.

Wollac: it now has `uint32 version` and `uint32 networkID`

PR #310, opened 2020-03-26
Value tangle updates

hmoog:


packages/binary/valuetransfer/address/address.go
```diff
@@ -0,0 +1,123 @@
+package address
+
+import (
+	"crypto/rand"
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"fmt"

```

PR #312, opened 2020-03-28
Integration tests

jonastheis:
# Description of change
This PR adds the possibility to easily run a local GoShimmer network via Docker and run integration tests. The Docker network can be run using `docker-compose up --scale peer_replica=x`. It contains following nodes:

- `entry_node`: The entry node that all other peers need to contact in order to discover each other.
- `peer_master`: A special peer that has its web API and SPA exposed to the host system. This is useful for manual testing. 
- `peer_replica`: Peers that are only addressable within the Docker network. 

Additionally, there is a separate `tester` container that joins the same Docker network and waits until nodes are reasonably peered. This `tester` can implement and execute any sort of end-to-end tests in the future.

## Type of change
- Enhancement


.github/workflows/test.yml
```diff
@@ -19,4 +19,48 @@ jobs:
...
+      run: docker-compose -f tools/integration-tests/docker-compose.yml up -d --scale peer_replica=5
+
+    - name: Dispay containers
+      run: docker ps -a
+
+    - name: Run integration tests
+      run: docker-compose -f tools/integration-tests/tester/docker-compose.yml up --exit-code-from tester
+
+    - name: Stop GoShimmer Docker network
+      if: always()
```
Wollac, line 42: This technically means that this is also run even when already the checkout failed, but I am not sure there is a way around this

tools/integration-tests/config.entry_node.json
```diff
@@ -0,0 +1,74 @@
...
+      "port": 1888
+    },
+    "httpServer": {
+      "bindAddress": "0.0.0.0:9000",
+      "dev": false
+    }
+  },
+  "autopeering": {
+    "entryNodes": [
+      "X2cmCzYnZDjmsvdAH90Q7oKmhNeTdwJdj2FX84adLzo=@entry_node:14626"
```
Wollac, line 16: It does not make sense to have oneself as an entry node.

tools/integration-tests/config.entry_node.json
```diff
@@ -0,0 +1,74 @@
+{
```
Wollac, line 1: Having copies of config files here, means that we need to adapt everything on config changes or additions..
Is there a feasible way to avoid this? Like specifying only the changes in a json/flags?

tools/integration-tests/README.md
```diff
@@ -0,0 +1,53 @@
+# Integration tests with Docker
+
+![Integration testing setup](../../images/integration-testing-setup.png)
+
+Running the integration tests spins up a GoShimmer network within Docker as schematically shown in the figure above.
+`N` defines the number of `peer_replicas` and can be specified when running the network.
+The peers can communicate freely within the Docker network and this is exactly how the tests are run using the `tester` container.
+Test can be written in regular Go style while the framework provides convenience functions to access a specific peer's web API or logs.
+
+The visualizer, `master_peer's` dashboard and web API are reachable from the host system on the respective ports.
```
Wollac, line 10: I think we should mention that the git submodules need to be initialized to have glumb visualizer ready.

.github/workflows/test.yml
```diff
@@ -19,4 +19,48 @@ jobs:
...
+    - name: Save logs as artifacts
+      if: always()
+      uses: actions/upload-artifact@v1
+      with:
+        name: container-logs
+        path: tools/integration-tests/logs
+
+    - name: Clean up
+      if: always()
+      run: docker-compose -f tools/integration-tests/docker-compose.yml down --rmi all --remove-orphan --volumes
```
Wollac, line 66: `--volumes`: I don't think we have any named volumes so this flag has no effect

Wollac, line 66: I am not a big fan of `--rmi`. What is the purpose of this?
In an action it does not matter whatsoever, and in the context of `run-tests.sh` should not a `... up --build` have the same effect that we always have an updated image?

tools/integration-tests/README.md
jonastheis: I was referring to the autopeering network visualizer. I'll make that more clear. But we could of course also expose the glumb visualizer. Though that probably needs some work with the new layer approach?

tools/integration-tests/tester/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: "3.7"
+
+services:
+  tester:
+    image: iotaledger/goshimmer_tester
+    container_name: tester
+    build:
+      context: ../../..
+      dockerfile: Dockerfile-tester
```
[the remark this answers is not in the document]
jonastheis: A bit tricky because you can't copy anything into an image outside of the Dockerfile. In combination with the reference to GoShimmer few directories up this is an issue. However, I solved it now with `image: golang:1.14` and made `tester` it's own module as well.

Dockerfile
```diff
@@ -44,4 +42,4 @@ COPY --from=build /go/bin/goshimmer /run/goshimmer
 # Copy the default config
 COPY config.default.json /config.json
 
-ENTRYPOINT ["/run/goshimmer", "--config-dir=/", "--database.directory=/mainnetdb"]
+ENTRYPOINT ["/run/goshimmer", "--config-dir=/", "--database.directory=/tmp/mainnetdb"]
```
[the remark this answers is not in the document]
jonastheis: The base image does not have `/bin/sh` to run `mkdir /mainnetdb`. But as a workaround I used `WORKDIR` to create the directory now.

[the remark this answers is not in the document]
jonastheis: This does, however, only work if there is a mounted volume which has `rw` rights. Inside the container (for the network) this does not work. No permission to write into this folder.

I'll leave it with `/tmp/mainnetdb`. 

Btw which tools are you referring to?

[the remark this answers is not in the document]
Wollac: @nuriel77 what do you think?
IIRC your playbook depends on the `mainnetdb` path. Is it ok to move it to `/tmp/mainnetdb` or, if necessary, do you have any proposals on how to keep it under `/mainnetdb` that are compatible with running the container without any mounts?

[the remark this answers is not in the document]
nuriel77: @Wollac Why does anything have to be hard-coded for the entry point?
For me doesn't matter much because I can also override the default entry point.
Nevertheless, I think it is best practice to avoid hard coding.

Why not use:
```
ENTRYPOINT ["/run/goshimmer", "--config-dir=/", "$@"]
```
Or even:
```
ENTRYPOINT ["/run/goshimmer", "--config-dir=/", "$@"]
```
And set the defaults inside of goshimmer? (i.e. config dir=/ and db dir to /tmp/etc...)?

client/lib.go
```diff
@@ -26,15 +26,15 @@ const (
 
 func NewGoShimmerAPI(node string, httpClient ...http.Client) *GoShimmerAPI {
 	if len(httpClient) > 0 {
-		return &GoShimmerAPI{node: node, httpClient: httpClient[0]}
+		return &GoShimmerAPI{baseUrl: node, httpClient: httpClient[0]}
```
Wollac, line 29: Make `NewGoShimmerAPI` parameter name consistent with the struct field. It should either both be node or baseUrl

client/lib.go
```diff
@@ -62,7 +62,7 @@ func interpretBody(res *http.Response, decodeTo interface{}) error {
 	case http.StatusInternalServerError:
 		return fmt.Errorf("%w: %s", ErrInternalServerError, errRes.Error)
 	case http.StatusNotFound:
-		return fmt.Errorf("%w: %s", ErrNotFound, errRes.Error)
+		return fmt.Errorf("%w: %s", ErrNotFound, res.Request.URL.String())
```
Wollac, line 65: Why was this changed? Is the error in the response not helpful? Not existent?
Is it always assured that `Request` or `URL` be nil and cause a panic?
All the other invalid status codes check for the error returned in the body and this one does not

PR #325, opened 2020-04-08
Refactors FPC packages

luca-moser:
* Rewrites the FPC respectively voter package.
* Adds a PRNG which uses the unix time to act as a substitution for a DRNG.


packages/vote/fpc/fpc.go
```diff
@@ -0,0 +1,267 @@
...
+	for _, opinionGiverToQuery := range opinionGiversToQuery {
+		wg.Add(1)
+		go func(opinionGiverToQuery vote.OpinionGiver) {
+			defer wg.Done()
+
+			queryCtx, cancel := context.WithTimeout(context.Background(), f.paras.QueryTimeout)
+			defer cancel()
+
+			// query
+			opinions, err := opinionGiverToQuery.Query(queryCtx, ids)
```
capossele, line 209: Maybe as a future improvment we might consider to not re-query the same `opinionGiverToQuery` (if sampled multiple times) multiple times and instead re-use its opinion. This should save some network resources.

packages/vote/fpc/fpc.go
```diff
@@ -0,0 +1,267 @@
...
+	f.enqueue()
+	// we can only form opinions when the last round was actually executed successfully
+	if f.lastRoundCompletedSuccessfully {
+		// form opinions by using the random number supplied for this new round
+		f.formOpinions(rand)
+		// clean opinions on vote contexts where an opinion was reached in FinalizationThreshold
+		// number of rounds and clear those who failed to be finalized in MaxRoundsPerVoteContext.
+		f.finalizeOpinions()
+	}
+	// query for opinions on the current vote contexts
```
capossele, line 105: Here we could trigger an event EndRound and include the following info:
* the `rand` used
* the list of `IDs` object of voting
* the list of `opinionGivers` with their opinions
* the formed opinion for all of the `IDs`

packages/vote/fpc/fpc.go
```diff
@@ -0,0 +1,267 @@
...
+func (f *FPC) formOpinions(rand float64) {
+	f.ctxsMu.RLock()
+	defer f.ctxsMu.RUnlock()
+	for _, voteCtx := range f.ctxs {
+		// when the vote context is new there's no opinion to form
+		if voteCtx.IsNew() {
+			continue
+		}
+
+		lowerThreshold := f.paras.SubsequentRoundsLowerBoundThreshold
```
capossele, line 136: Let's fix the `lowerThreshold` to `0.5` so that the range for the subsequent rounds is always between 0.5 and 1-Beta. I know it's different from the FPC white paper [beta - 1-beta] but simulations have shown that FPC performs better in this way.

packages/vote/fpc/parameters.go
```diff
@@ -0,0 +1,42 @@
...
+	MaxRoundsPerVoteContext int
+	// The max amount of time a query is allowed to take.
+	QueryTimeout time.Duration
+}
+
+// DefaultParameters returns the default parameters used in FPC.
+func DefaultParameters() *Parameters {
+	return &Parameters{
+		FirstRoundLowerBoundThreshold:       0.75,
+		FirstRoundUpperBoundThreshold:       0.85,
```
capossele, line 29: Let's initialize both the `FirstRoundLowerBoundThreshold` and the `FirstRoundUpperrBoundThreshold` to the same value of `0.75`

packages/vote/fpc/parameters.go
```diff
@@ -0,0 +1,42 @@
...
+		SubsequentRoundsLowerBoundThreshold: 0.33,
+		QuerySampleSize:                     10,
```
capossele, line 31: Let's increase the `QuerySampleSize` to `21` for now.

packages/vote/fpc/parameters.go
```diff
@@ -0,0 +1,42 @@
...
+		FinalizationThreshold:               2,
+		CoolingOffPeriod:                    2,
```
capossele, line 33: Let's set the `CoolingOffPeriod` to `0`.

packages/vote/fpc/parameters.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// DefaultParameters returns the default parameters used in FPC.
+func DefaultParameters() *Parameters {
+	return &Parameters{
+		FirstRoundLowerBoundThreshold:       0.75,
+		FirstRoundUpperBoundThreshold:       0.85,
+		SubsequentRoundsLowerBoundThreshold: 0.33,
+		QuerySampleSize:                     10,
+		FinalizationThreshold:               2,
```
capossele, line 32: Let's set the `FinalizationThreshold` to 10.

packages/vote/fpc/fpc.go
```diff
@@ -0,0 +1,267 @@
...
+	queueSet map[string]struct{}
+	queueMu  sync.Mutex
+	// contains the set of current vote contexts.
+	ctxs   map[string]*VoteContext
+	ctxsMu sync.RWMutex
+	// parameters to use within FPC.
+	paras *Parameters
+	// indicates whether the last round was performed successfully
+	lastRoundCompletedSuccessfully bool
+	rng                            *rand.Rand
```
capossele, line 59: Should we rename this `rng` to something like `opinionGiverRng` or `sampleRng` so that it is more clear that it's used only for randomizing the opinionGiver sample? wdyt?

packages/vote/net/query.proto
```diff
@@ -0,0 +1,17 @@
...
+service VoterQuery {
+    rpc Opinion (QueryRequest) returns (QueryReply) {}
+}
+
+message QueryRequest {
+    repeated string id = 1;
+}
+
+message QueryReply {
+    repeated int32 opinion = 1;
```
capossele, line 16: That's unfortunate. Doesn't protobuf allow to use something smaller than 4 bytes?

packages/vote/net/query.proto
```diff
@@ -0,0 +1,17 @@
...
+
+service VoterQuery {
+    rpc Opinion (QueryRequest) returns (QueryReply) {}
+}
+
+message QueryRequest {
+    repeated string id = 1;
+}
+
+message QueryReply {
```
capossele, line 15: Just as a reminder, we will eventually need to sign each `QueryReply` and verify its validity on the receiving side.

plugins/fpc/parameters.go
```diff
@@ -0,0 +1,17 @@
...
+const (
+	CfgFPCQuerySampleSize = "fpc.querySampleSize"
+	CfgFPCRoundInterval   = "fpc.roundInterval"
+	CfgFPCBindAddress     = "fpc.bindAddress"
+)
+
+func init() {
+	flag.Int(CfgFPCQuerySampleSize, 3, "Size of the voting quorum (k)")
+	flag.Int(CfgFPCRoundInterval, 5, "FPC round interval [s]")
+	flag.String(CfgFPCBindAddress, "0.0.0.0:14636", "the bind address on which the FPC vote server binds to")
```
capossele, line 16: The default port should be `10895`.

plugins/fpc/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	voter                *fpc.FPC
+	voterOnce            sync.Once
+	voterServer          *votenet.VoterServer
+	roundIntervalSeconds int64 = 5
+)
+
+func Voter() vote.DRNGRoundBasedVoter {
+	voterOnce.Do(func() {
+		// create a function which gets OpinionGivers
+		opinionGiverFunc := func() (givers []vote.OpinionGiver, err error) {
```
capossele, line 42: Is it possbile to return an err != nil from this function? If not maybe let's remove the returned error.

packages/vote/net/query.proto
luca-moser: Unfortunately not, however, over the wire, protobuf truncates the amount of bytes used to the amount required. So this would actually be one byte.

plugins/fpc/plugin.go
luca-moser: I'd leave it for now because there might be an impl. of this function type which does indeed return an error in the future.

packages/vote/fpc/fpc.go
luca-moser: I've added a `RoundExecuted` event which gives you the rand used, the active vote contexts and the opinion givers with their corresponding response opinion. Additionally added a duration field and how many times a given opinion giver's opinion was taken into account.

packages/vote/fpc/fpc.go
luca-moser: I've modified this now to be a set of OpinionGiver so the actual query is done only once.

packages/vote/fpc/parameters.go
luca-moser: Changed

packages/vote/fpc/parameters.go
luca-moser: Changed

packages/vote/fpc/parameters.go
luca-moser: Changed

packages/vote/fpc/parameters.go
luca-moser: Changed

packages/vote/fpc/fpc.go
luca-moser: Renamed to `opinionGiverRng`

plugins/fpc/parameters.go
luca-moser: Changed

PR #334, opened 2020-04-17
Feat/api info

capossele:
# Description of change

Adds the API endpoint `/info` providing:

- Version
- Identity (base58 encoded)
- Public Key (base58 encoded)
- Enabled plugins
- Disabled plugins

## Type of change

### DIFF
--- a/client/info.go
+++ b/client/info.go
@@ -11,11 +11,9 @@ const (
 )
 
 // Info gets the info of the node.
-func (api *GoShimmerAPI) Info(knownPeers bool) (*webapi_info.Response, error) {
+func (api *GoShimmerAPI) Info() (*webapi_info.Response, error) {
 	res := &webapi_info.Response{}
-	if err := api.do(http.MethodGet, func() string {
-		return routeInfo
-	}(), nil, res); err != nil {
+	if err := api.do(http.MethodGet, routeInfo, nil, res); err != nil {
 		return nil, err
 	}
 	return res, nil
--- a/client/info.go
+++ b/client/info.go
@@ -1,0 +1,22 @@
+package client
+
+import (
+	"net/http"
+
+	webapi_info "github.com/iotaledger/goshimmer/plugins/webapi/info"
+)
+
+const (
+	routeInfo = "info"
+)
+
+// Info gets the info of the node.
+func (api *GoShimmerAPI) Info(knownPeers bool) (*webapi_info.Response, error) {
+	res := &webapi_info.Response{}
+	if err := api.do(http.MethodGet, func() string {
+		return routeInfo
+	}(), nil, res); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
--- a/plugins/webapi/info/plugin.go
+++ b/plugins/webapi/info/plugin.go
@@ -1,13 +1,14 @@
 package info
 
 import (
-	"encoding/hex"
 	"net/http"
 
 	"github.com/iotaledger/goshimmer/plugins/autopeering/local"
+	"github.com/iotaledger/goshimmer/plugins/banner"
 	"github.com/iotaledger/goshimmer/plugins/webapi"
 	"github.com/iotaledger/hive.go/node"
 	"github.com/labstack/echo"
+	"github.com/mr-tron/base58"
 )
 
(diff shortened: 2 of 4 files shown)


client/info.go
```diff
@@ -0,0 +1,22 @@
...
+
+	webapi_info "github.com/iotaledger/goshimmer/plugins/webapi/info"
+)
+
+const (
+	routeInfo = "info"
+)
+
+// Info gets the info of the node.
+func (api *GoShimmerAPI) Info(knownPeers bool) (*webapi_info.Response, error) {
```
jonastheis, line 14:
```suggestion
func (api *GoShimmerAPI) Info() (*webapi_info.Response, error) {
```

client/info.go
```diff
@@ -0,0 +1,22 @@
...
+	res := &webapi_info.Response{}
+	if err := api.do(http.MethodGet, func() string {
+		return routeInfo
+	}(), nil, res); err != nil {
```
jonastheis, line 18:
```suggestion
	if err := api.do(http.MethodGet, routeInfo, nil, res); err != nil {
```
No need to call a closure here

plugins/webapi/info/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+			enabledPlugins = append(enabledPlugins, plugin)
+		default:
+			continue
+		}
+	}
+
+	return c.JSON(http.StatusOK, Response{
+		Version:         banner.AppVersion,
+		Identity:        base58.Encode(local.GetInstance().Identity.ID().Bytes()),
+		PublicKey:       base58.Encode(local.GetInstance().PublicKey().Bytes()),
```
jonastheis, line 72: Use the `String()` methods here. Then you also don't need to worry about different representations :)

plugins/webapi/info/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+// 	"disabledlugins":[
+// 		"Graph",
+// 		"RemoteLog",
+// 		"Spammer",
+// 		"WebAPI Auth"
+// 	]
+// }
+func getInfo(c echo.Context) error {
+	enabledPlugins := []string{}
+	disabledPlugins := []string{}
```
jonastheis, line 57:
```suggestion
	var enabledPlugins []string
	var disabledPlugins []string
```
Not entirely sure but I think its more Go idiomatic to declare a nil slice instead of an empty literal declaration.

plugins/webapi/info/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+		DisabledPlugins: disabledPlugins,
+	})
+}
+
+// Response holds the response of the GET request.
+type Response struct {
+	// version of GoShimmer
+	Version string `json:"version,omitempty"`
+	// identity of the node encoded in base58
+	Identity string `json:"identity,omitempty"`
```
jonastheis, line 83: How do we call it in the SPA dashboard? Is it `identity` as well or `ID`? We should try to be consistent.

plugins/webapi/info/plugin.go
capossele: The String() method of the Identity.ID() returns hex.EncodeToString(id[:8])
Should we use that?

plugins/webapi/info/plugin.go
capossele: It's called `Node` in the frontend, but tbh I prefer to use identity. Maybe we should change the SPA frontend to `Node's identity`?

plugins/webapi/info/plugin.go
jonastheis: I think it's also about consistency here. If we use the method everywhere it's easy to change the representation just at one place. I don't mind it being the full ID or just part of it, I just want it to be consistent! :)

capossele: I meant, my point is that if we want to use the String method everywhere we need to change its implementation (from hex encoded to base58 encoded) and I did not want to touch the autopeering.

capossele: Unless of course you are fine with hex encoded for the identity

plugins/webapi/info/plugin.go
capossele: Just to be more precise I've changed it to IdentityID

PR #335, opened 2020-04-17
Enabled golint warnings for missing comments

Wollac:
- Enable `golint` in GolangCI
- Do not excluded warnings about missing comments
- Set confidence level to 0.9 to disable capitalization warnings

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,7 +8,7 @@ linters-settings:
     gofmt:
         simplify: true
     golint:
-        min-confidence: 0.9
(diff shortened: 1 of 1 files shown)


Wollac:
We could set the confidence level to the default `0.8` to also get warnings like:
- _don't use ALL_CAPS in Go names; use CamelCase_
- _var trunkTransactionId should be trunkTransactionID_

luca-moser:
Lets enforce that vars/consts are not all in caps. (we currently have a lot of those)

PR #339, opened 2020-04-20
Flexible integration tests

jonastheis:
# Description of change
Introduces flexible integration tests where each test can spawn its own individual Docker/GoShimmer network. Allows us to test for solidification, eventually. 

## Type of change


tools/integration-tests/tester/framework/network.go
```diff
@@ -0,0 +1,179 @@
...
+func (n *Network) WaitForAutopeering(minimumNeighbors int) {
+	fmt.Printf("Waiting for autopeering...\n")
+	defer fmt.Printf("Waiting for autopeering... done\n")
+
+	maxTries := autopeeringMaxTries
+	for maxTries > 0 {
+
+		for _, p := range n.peers {
+			if resp, err := p.GetNeighbors(false); err != nil {
+				fmt.Printf("request error: %v\n", err)
```
Wollac, line 133: Is it expected behavior that the request fails? And if yes, is it correct to ignore this peer?
Wouldn't it maybe more consistent to cancel WaitForAutopeering with an error in this case

tools/integration-tests/tester/framework/util.go
```diff
@@ -0,0 +1,33 @@
...
+	f, err := os.Create(fmt.Sprintf("%s%s.log", logsDir, name))
+	if err != nil {
+		panic(err)
+	}
+	defer f.Close()
+
+	// remove non-ascii chars at beginning of line
+	scanner := bufio.NewScanner(logs)
+	for scanner.Scan() {
+		bytes := append(scanner.Bytes()[8:], '\n')
```
Wollac, line 28: The `8` seems rather arbitrary, consider defining a properly named const to give it more context.

tools/docker-network/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The different containers (`entry_node`, `peer_master`, `peer_replica`) are based on the same config file 
+and separate config file and modified as necessary, respectively. 
+
+## How to use as development tool
+Using a standalone throwaway Docker network can be really helpful as a development tool. 
+
+Prerequisites: 
+- Docker
+- Docker compose
```
Wollac, line 18: As we already have such a nice README here, I think we should add that we are requiring >= 3.5

tools/integration-tests/tester/framework/network.go
jonastheis: It can happen that a container and its GoShimmer instance is not yet fully started once this is called. This obviously depends on the underlying system and I don't think we have another way for checking this except maybe [Docker's healthcheck feature](https://docs.docker.com/compose/compose-file/#healthcheck). But this would make it unnecessarily complicated.

tools/integration-tests/tester/framework/network.go
```diff
@@ -0,0 +1,179 @@
...
+		if min >= minimumNeighbors {
+			fmt.Printf("Neighbors: min=%d avg=%.2f\n", min, float64(total)/float64(len(n.peers)))
+			return
+		}
+
+		fmt.Println("Not done yet. Try again in 5 seconds...")
+		time.Sleep(5 * time.Second)
+		maxTries--
+	}
+	panic("Peering not successful.")
```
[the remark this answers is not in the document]
jonastheis: I was thinking about error handling in general. While developing I just put a panic everywhere to make it easier and faster to test. How should we handle errors emerging from Docker in general?

tools/integration-tests/tester/framework/network.go
Wollac: OK, then let's keep it like this!

tools/integration-tests/tester/framework/network.go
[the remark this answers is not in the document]
Wollac: I think everything in framework etc should rather return errors instead of panics. And then in the actual tests we would check for those with `require.noError` or something.
If someone uses the framework outside of tests, they are supposed to do their own error handling

tools/integration-tests/tester/framework/framework.go
```diff
@@ -1,99 +1,85 @@
...
+	dockerClient *client.Client
 }
 
-// New creates a new instance of Framework, gets all available peers within the Docker network and
-// waits for them to autopeer.
-// Panics if no peer is found.
-func New() *Framework {
-	fmt.Printf("Finding available peers...\n")
+// Instance returns the singleton Framework instance.
+func Instance() (*Framework, error) {
```
Wollac, line 27: Do not make `instanceErr` a global variable, instead use named return values for `Instance` for the same effect.

PR #340, opened 2020-04-21
:bug: adds len check

capossele:
# Description of change

Adds a check of the length of the payload before parsing it.

## Type of change

### DIFF
--- a/plugins/drng/plugin.go
+++ b/plugins/drng/plugin.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/iotaledger/goshimmer/packages/binary/drng"
 	"github.com/iotaledger/goshimmer/packages/binary/drng/payload"
+	"github.com/iotaledger/goshimmer/packages/binary/drng/payload/header"
 	"github.com/iotaledger/goshimmer/packages/binary/drng/state"
 	cbPayload "github.com/iotaledger/goshimmer/packages/binary/drng/subtypes/collectiveBeacon/payload"
 	"github.com/iotaledger/goshimmer/packages/binary/messagelayer/message"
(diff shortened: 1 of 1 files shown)


plugins/drng/plugin.go
```diff
@@ -68,6 +69,9 @@ func configureEvents() {
 		cachedMessageMetadata.Release()
 
 		cachedMessage.Consume(func(msg *message.Message) {
+			if len(msg.Payload().Bytes()) < header.Length {
```
jonastheis, line 72: We should probably come up with a standard template how to handle messages in a solidification event of the base tangle. Does the following make sense?

- check `msg.Payload().Type()`. This is not a guarantee since anybody could send a message with your type but already a good and fast filter
- the payload interface should provide a `Size` method. Since this information is given in the message and can be verified this is also a quite fast check. Correct? @hmoog 
- then do any additional application specific checks

capossele: Good points. I've added the Type() check. 
For the Size, I'm not sure. Should we trust whatever written in the size field of the message? Using len(payload.Bytes()) should be more safe. Wdyt?

jonastheis: I think the base layer could verify it and store it. Not sure if it actually does this currently. That's why I linked Hans. What happens when the payload is shorter than the bytes in its header?

capossele: Nothing critical, just the error you noticed: 
`tried to read 5 bytes from 4 bytes input`
that happens while parsing the message

hmoog: I usually simply "Parse" the payload and try to transform it into an object of the corresponding type. If the parsing fails, then we know its an invalid message and it gets dropped.

capossele: Yeah I do the same (line 76). Should I then just remove the subsequent log?

jonastheis: Maybe a debug if it's useful? But I don't really mind since it shouldn't appear anymore on every message anyway.

PR #344, opened 2020-04-21
Normalize global plugin variables

luca-moser:
* `PLUGIN` -> `Plugin`
* +const `PluginName` per plugin


luca-moser:
Yeah sure, I'll resolve the conflict then.

PR #346, opened 2020-04-21
Synchronization/solidification integration test

jonastheis:
# Description of change

Adds an integration test `TestNodeSynchronization` that checks whether messages are synchronized by a peer that joined the network later and initially missed some messages.

Relates to #329


tools/integration-tests/tester/tests/synchronization_test.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// TestNodeSynchronization checks whether messages are synchronized by a peer that joined the network later
+// and initially missed some messages.
+func TestNodeSynchronization(t *testing.T) {
+	n := f.CreateNetwork("TestNodeSynchronization", 4, 2)
+	defer n.Shutdown()
+
+	numMessages := 100
+	ids := make([]string, numMessages)
```
luca-moser, line 19: Consider using a map as a set so you don't have to loop over to find whether an id exists in `checkForMessageIds`

jonastheis: The problem here is that we can't trust the output of `resp.Messages`. Even if `len(resp.Messages)==len(ids)` it doesn't mean it's all the right `ids`. There could be duplicates or even all the same messages. When checking with a map this would still succeed. Therefore, we need to explicitly check for every `id` within `resp.Messages`.

luca-moser: What I was referring to is that in `checkForMessageIds` you have a for loop with an inner for loop to check against resp.Messages whether the given ID was one of the ones from the issued messages. If you store the IDs of the messages you've issued into a set, you can simply iterate over resp.Message and check the set whether the given ID exists.

This:
```
	outer:
		for _, id := range ids {
			for _, msg := range resp.Messages {
				// if message found skip to next
				if msg.Id == id {
					continue outer
				}
			}

			t.Errorf("MessageId=%s, issuer=%s not found in peer %s.", id, idsMap[id].issuerPublicKey, peer.String())
		}
```

would become this:
```
for _, msg := range resp.Messages {
           if , has := ids[msg.Id]; !has {
               t.Errorf("MessageId=%s, issuer=%s not found in peer %s.", id, idsMap[id].issuerPublicKey, peer.String())
           }
}
```

jonastheis: Yes but iterating over `resp.Message` has exactly the problems I stated above. We can't trust this as this could be x times the same message and then the second version would still succeed. But it should not.

luca-moser: When you're adding the id to the set after issuing a message, you simply check whether the id is already in the set and if so, you error out early since a duplicated id was returned from a newly issued message.

tools/integration-tests/tester/tests/synchronization_test.go
```diff
@@ -0,0 +1,122 @@
...
+		for _, id := range ids {
+			for _, msg := range resp.Messages {
+				// if message found skip to next
+				if msg.ID == id {
+					continue outer
+				}
+			}
+
+			t.Errorf("MessageId=%s, issuer=%s not found in peer %s.", id, idsMap[id].issuerPublicKey, peer.String())
+		}
```
Wollac, line 111:
```suggestion
		respIDs := make(map[string]struct{}, len(resp.Messages))
		for _, msg := range resp.Messages {
			respIDs[msg.ID] = struct{}{}
		}
		for _, id := range ids {
			assert.Containsf(t, respIDs, id, "MessageId=%s, issuer=%s not found in peer %s.", id, idsMap[id].issuerPublicKey, peer.String())
		}
```
I guess this is also what @luca-moser was referring to earlier.
IMHO it looks a bit nicer (by avoiding the label) and hash linear runtime instead of quadratic (not that this matters much in this test, though)

tools/integration-tests/tester/tests/synchronization_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		// check for general information
+		for _, msg := range resp.Messages {
+			msgSent := idsMap[msg.ID]
+
+			assert.Equal(t, msgSent.issuerPublicKey, msg.IssuerPublicKey, "MessageId=%s, issuer=%s not correct issuer in %s.", msgSent.id, msgSent.issuerPublicKey, peer.String())
```
Wollac, line 117: I find it usually a bit nicer to use the f version when a format string is used (so assert.Equalf), as this is also how functions in stdlib are named. However, internally it is just an alias

tools/integration-tests/tester/tests/synchronization_test.go
jonastheis: I changed it to a similar approach but at the same time making sure that `resp.Messages` also contains no additional `ids`.

tools/integration-tests/tester/tests/synchronization_test.go
jonastheis: Changed it to use a map.

tools/integration-tests/tester/tests/synchronization_test.go
Wollac: also in the above you already do that with `assert.Equal(t, numMessages, len(resp.Messages))`

Wollac: There is another option if you want to assure that in one go:
```go
respIDs := make([]string, len(resp.Messages))
for i, msg := range resp.Messages {
	respIDs[i] = msg.ID
}
assert.ElementsMatch(t, ids, respIDs)
```
That is definitely the shortest version and we let `assert` handle all the interesting stuff.

jonastheis: Definitely the most elegant solution. However, it doesn't yield as much information if a message is missing. But I think this can be inferred from the logs anyway.

PR #347, opened 2020-04-21
Cleanup gossip

### LINKED ISSUE #243: 9 Neighbors allowed
# Issue
After running for 24 hours Node shows 9 neighbors

# Environment
Hertzner CX41 Ubuntu 18.04 Goshimmer 1.1

[Stack Dump](https://github.com/iotaledger/goshimmer/files/4174938/goroutine.zip)
9 Neighbors

Wollac:
- Make `AddInbound`, `AddOutbound`, `DropNeighbor` atomic to assure that adding the peer to the map is not interrupted
- Trigger the `RemoveNeighbor` event only when the connection is closed.
- Add test with many drops/connects

Fixes #243


packages/gossip/manager.go
```diff
@@ -146,8 +132,9 @@ func (m *Manager) SendTransaction(txData []byte, to ...identity.ID) {
 }
 
 func (m *Manager) GetAllNeighbors() []*Neighbor {
```
luca-moser, line 134: Missing function comment. Consider renaming to `AllNeighbors()`

packages/gossip/manager_test.go
```diff
@@ -372,6 +372,46 @@ func TestTxRequest(t *testing.T) {
...
+	mgrA, closeA, peerA := newTestManager(t, "A")
+	defer closeA()
+	mgrB, closeB, peerB := newTestManager(t, "B")
+	defer closeB()
+
+	connect := func() {
+		var wg sync.WaitGroup
+		closure := events.NewClosure(func(_ *Neighbor) { wg.Done() })
+		wg.Add(2)
+		Events.NeighborAdded.Attach(closure)
```
luca-moser, line 385: I believe that `Events` should be part of the actual neighbor manager struct instead of being solely defined in the gossip package.

packages/gossip/manager.go
Wollac: done

packages/gossip/manager_test.go
Wollac: I agree 100%, but I think this is out of the scope of this simple PR.
I'll add an issue to do that for gossip and autopeering.

Wollac:
This will probably conflict with the changes from #331 so we should merge #331 first.

packages/gossip/manager_test.go
Wollac: #349

PR #358, opened 2020-04-23
Fix SIGSEGV when drng plugin is disabled

capossele:
# Description of change

- Fix SIGSEGV when drng plugin is disabled by checking that the `drng` plugin status.
- Makes plugin instance private and implements Instance() instead.

## Type of change

### DIFF
--- a/plugins/dashboard/drng_livefeed.go
+++ b/plugins/dashboard/drng_livefeed.go
@@ -42,6 +42,9 @@ func runDrngLiveFeed() {
 	})
 
 	daemon.BackgroundWorker("Dashboard[DRNGUpdater]", func(shutdownSignal <-chan struct{}) {
+		if drng.Instance == nil {
+			return
+		}
 		drng.Instance.Events.Randomness.Attach(notifyNewRandomness)
 		drngLiveFeedWorkerPool.Start()
 		<-shutdownSignal
--- a/plugins/dashboard/drng_livefeed.go
+++ b/plugins/dashboard/drng_livefeed.go
@@ -21,8 +21,8 @@ func configureDrngLiveFeed() {
 		newRandomness := task.Param(0).(state.Randomness)
 
 		sendToAllWSClient(&wsmsg{MsgTypeDrng, &drngMsg{
-			Instance:      drng.Instance.State.Committee().InstanceID,
-			DistributedPK: hex.EncodeToString(drng.Instance.State.Committee().DistributedPK),
+			Instance:      drng.Instance().State.Committee().InstanceID,
+			DistributedPK: hex.EncodeToString(drng.Instance().State.Committee().DistributedPK),
 			Round:         newRandomness.Round,
 			Randomness:    hex.EncodeToString(newRandomness.Randomness[:32]),
 			Timestamp:     newRandomness.Timestamp.Format("2 Jan 2006 15:04:05")}})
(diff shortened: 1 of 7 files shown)


plugins/drng/drng.go
```diff
@@ -14,6 +18,43 @@ var (
 	ErrParsingCommitteeMember = errors.New("cannot parse committee member")
 )
 
+func configureDRNG() *drng.DRNG {
+	// parse identities of the committee members
+	committeeMembers, err := parseCommitteeMembers()
+	if err != nil {
+		log.Warnf("Invalid %s: %s", CfgDRNGCommitteeMembers, err)
```
Wollac, line 25: `configureDRNG` uses `log`, but the logger is not initialized in `Instance()`. This will fail if `Instance()` is called before `configure`.

plugins/drng/plugin.go
```diff
@@ -24,42 +21,14 @@ const PluginName = "DRNG"
...
-	committeeConf := &state.Committee{
-		InstanceID:    config.Node.GetUint32(CfgDRNGInstanceID),
-		Threshold:     uint8(config.Node.GetUint32(CfgDRNGThreshold)),
-		DistributedPK: dpk,
-		Identities:    committeeMembers,
-	}
-
-	Instance = drng.New(state.SetCommittee(committeeConf))
-
+	instance = Instance()
```
Wollac, line 31: This is an ineffectual assignment, as `Instance()` already sets the global variable instance.
Probably, the cleanest solution to assure that we have an instance in `configureEvents` is to do
```go
instance := Instance()
```
in that function. This makes no assumptions on `Instance()` and even works for non-singletons.

plugins/drng/plugin.go
```diff
@@ -83,13 +52,17 @@ func configureEvents() {
...
 			}
-			log.Info(Instance.State.Randomness())
+			log.Info(instance.State.Randomness())
 		})
 	}))
+}
 
+// Enabled returns the enabled status of the plugin
+func Enabled() bool {
+	return Plugin.Status != node.Disabled
```
Wollac, line 67: This check is not correct as it does not check for dynamically disabled plugins.
Remove this function and use `node.IsSkipped` from `github.com/iotaledger/hive.go/node` where needed.

plugins/dashboard/drng_livefeed.go
```diff
@@ -42,11 +42,14 @@ func runDrngLiveFeed() {
 	})
 
 	daemon.BackgroundWorker("Dashboard[DRNGUpdater]", func(shutdownSignal <-chan struct{}) {
-		drng.Instance.Events.Randomness.Attach(notifyNewRandomness)
+		if !drng.Enabled() {
```
Wollac, line 45: Consider not running `BackgroundWorker` at all, if the plugin is disabled.

Wollac, line 45: Use `node.IsSkipped`

plugins/dashboard/drng_livefeed.go
```diff
@@ -42,11 +42,14 @@ func runDrngLiveFeed() {
...
+			return
+		}
+		drng.Instance().Events.Randomness.Attach(notifyNewRandomness)
 		drngLiveFeedWorkerPool.Start()
 		<-shutdownSignal
 		log.Info("Stopping Dashboard[DRNGUpdater] ...")
-		drng.Instance.Events.Randomness.Detach(notifyNewRandomness)
+		drng.Instance().Events.Randomness.Detach(notifyNewRandomness)
 		newMsgRateLimiter.Stop()
```
Wollac, line 53: The stop should be deferred, right after `NewTicker`.
Also consider deferring the other statements to assure cleanup in case of panic.

PR #365, opened 2020-04-23
Adds graph visualizer to dashboard

luca-moser:
Adds a visualizer like glumb but embedded into the dashboard showing vertices, tips, solid/unsolid, approvers/approvees and trunk/branch of the message layer.

image


plugins/dashboard/visualizer.go
```diff
@@ -0,0 +1,96 @@
...
+		visualizerWorkerPool.TrySubmit(messageId, true)
+	})
+
+	notifyDeletedTip := events.NewClosure(func(messageId message.Id) {
+		visualizerWorkerPool.TrySubmit(messageId, false)
+	})
+
+	daemon.BackgroundWorker("Dashboard[Visualizer]", func(shutdownSignal <-chan struct{}) {
+		messagelayer.Tangle.Events.MessageAttached.Attach(notifyNewMsg)
+		messagelayer.Tangle.Events.MessageSolid.Attach(notifyNewMsg)
```
Wollac, line 84: Is there a particular reason, why `MessageSolid` is the only event to detached?

I am usually a fan of deferring right after the Attach to make it easier glanceable that all the cleanups have indeed happened. But this is also fine with me...

plugins/dashboard/ws.go
```diff
@@ -0,0 +1,146 @@
...
+	}, workerpool.WorkerCount(wsSendWorkerCount), workerpool.QueueSize(wsSendWorkerQueueSize))
+}
+
+func runWebSocketStreams() {
+	updateStatus := events.NewClosure(func(mps uint64) {
+		wsSendWorkerPool.TrySubmit(mps)
+	})
+
+	daemon.BackgroundWorker("Dashboard[StatusUpdate]", func(shutdownSignal <-chan struct{}) {
+		metrics.Events.ReceivedMPSUpdated.Attach(updateStatus)
```
Wollac, line 62: If the Echo server stops due to an error (maybe the bind address is already used), we are still doing all these updates for no reason. Is there an easy fix for that? If not let's just leave it like this.

plugins/dashboard/visualizer.go
luca-moser: Changed it to use defer right after the Attach()

plugins/dashboard/ws.go
luca-moser: I believe GoShimmer should shutdown in that case automatically but for that we'd need some new mechanism to do that. Perhaps we can expose `gracefulStop` from the `gracefulshutdown` plugin in order to make it shutdown properly.

plugins/dashboard/visualizer.go
Wollac: With this it could theoretically happen that any of the events is triggered _after_ the worker pool has been stopped, but _before_ the closure has been detached. Could this cause any issues? Probably not, since `TrySubmit` should handle this gracefully.

luca-moser: Yeah nothing happens because if the pool is shutdown the submission is discarded immediately.

Wollac: :ok_hand:

plugins/dashboard/ws.go
Wollac: Currently, the only option we would have for this is to panic or exit, which is indeed not very nice.
Had a quick look at the `gracefulshutdown` plugin and it seems to be broken as `daemon.GetRunningBackgroundWorkers` is locked during shutdown. So, it does not do more than just calling `daemon.Shutdown()`.
Maybe we could just create a util function somewhere that calls daemon.Shutdown with and prints an error. That could be used for all the errors that should lead to a graceful shutdown...

luca-moser: https://github.com/iotaledger/goshimmer/issues/385 to track this

PR #378, opened 2020-04-27
Do not limit the number of issues per linter

Wollac:
By default `GolangCI-Lint` limits the number of issues per linter to 50. This is not enough.


Wollac:
:see_no_evil: this is how it should have looked:
https://github.com/iotaledger/goshimmer/pull/379

PR #403, opened 2020-04-30
Refactors analysis plugins

luca-moser:
* Makes the analysis functionality 3 plugins: `Analysis-Client`, `Analysis-Server` and `Analysis-WebInterface`. Before it was all one plugin which weirdly would enable server/client functionalities by looking whether ports were defined in the config or not.
*  Changes the heartbeat messages to not include a counter for inbound ids, since it can be derived from the size of the packet. 
* Adds tests for serializing and deserializing heartbeat messages.
* `CfgAnalysisServerBindAddress ` defines the bind address of the `Analysis-Server`.
* `CfgAnalysisWebInterfaceBindAddress` defines the bind address of the `Analysis-WebInterface`.


config.default.json
```diff
@@ -4,9 +4,9 @@
       "serverAddress": "node1.goshimmer.dev:188"
     },
     "server": {
-      "port": 0
+      "bindAddress": "0.0.0.0:16178"
```
Wollac, line 7: @jonastheis do we need to update any test config files after that? Or does that not matter because it's disabled anyways?

plugins/analysis/client/plugin.go
```diff
@@ -8,24 +8,41 @@ import (
...
+)
 
-// Run runs the plugin.
-func Run(plugin *node.Plugin) {
-	log = logger.NewLogger("Analysis-Client")
-	daemon.BackgroundWorker("Analysis Client", func(shutdownSignal <-chan struct{}) {
-		ticker := time.NewTicker(ReportInterval * time.Second)
+func run(_ *node.Plugin) {
+	log = logger.NewLogger(PluginName)
+	_ = daemon.BackgroundWorker(PluginName, func(shutdownSignal <-chan struct{}) {
```
Wollac, line 44: Instead of ignoring potential errors from `BackgroundWorker` consider logging them

plugins/analysis/server/plugin.go
```diff
@@ -2,60 +2,81 @@ package server
...
+}
+
 var (
-	// ErrInvalidPackageHeader defines an invalid package header error.
-	ErrInvalidPackageHeader = errors.New("invalid package header")
-	// ErrExpectedInitialAddNodePackage defines an expected initial add node package error.
-	ErrExpectedInitialAddNodePackage = errors.New("expected initial add node package")
-	server                           *tcp.TCPServer
-	log                              *logger.Logger
+	// ErrInvalidPacketHeader defines an invalid package header error.
```
Wollac, line 39:
```suggestion
	// ErrInvalidPacketHeader defines an invalid packet header error.
```

plugins/analysis/server/plugin.go
```diff
@@ -2,60 +2,81 @@ package server
...
+	addr, portStr, err := net.SplitHostPort(bindAddr)
+	if err != nil {
+		log.Fatal("invalid bind address in %s: %s", CfgAnalysisServerBindAddress, err)
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		log.Fatal("invalid port in %s: %s", CfgAnalysisServerBindAddress, err)
+	}
+
+	_ = daemon.BackgroundWorker(PluginName, func(shutdownSignal <-chan struct{}) {
```
Wollac, line 68: Consider logging the error.

plugins/analysis/webinterface/plugin.go
```diff
@@ -1,19 +1,97 @@
...
+		}
+		engine.GET("/", index)
+	}
+
+	engine.GET("/datastream", echo.WrapHandler(websocket.Handler(dataStream)))
+	configureEventsRecording(plugin)
+}
+
+func run(_ *node.Plugin) {
+	log.Infof("Starting %s ...", PluginName)
```
Wollac, line 54: In the other plugin.run we are not saying `Starting ...`, but I guess it can't hurt :shrug:

tools/entry-node/docker-compose.yml
```diff
@@ -18,8 +18,9 @@ services:
       --autopeering.seed=${SEED}
       --autopeering.entryNodes=
       --analysis.client.serverAddress=
```
Wollac, line 20: What about this, shouldn't we disable the analysis.client plugin for the entrynode?

plugins/analysis/client/plugin.go
luca-moser: Made it panic.

plugins/analysis/server/plugin.go
luca-moser: ditto

plugins/analysis/webinterface/plugin.go
luca-moser: Removed

tools/entry-node/docker-compose.yml
luca-moser: Disabled analysis-client

PR #407, opened 2020-05-06
`PutSignature` method to the value transaction structure. Bug fix with NewOutputs

lunfardo314:
# Description of change
1. added new method `PutSignature` to the value transaction structure. It is needed in situations, when signature is constructed not by signing and `Sign` method is not enough. 
For example in threshold signature scenario signature is aggregated from partial signatures without knowing the private key
2. exported `signaturescheme.NewBLSSignature` function

## Type of change
Enhancement 

Added corresponding unit tests to the `transaction` package

## Change checklist


lunfardo314:
- fixed bug with non-deterministic `NewOutputs` function in value transaction package
- added function to sort slice of addresses

Issue #208, opened 2020-01-30
Integrate URTS TSA with opinion based local modifier


luca-moser:
* Include the like state of the transaction
* Taking into account the opinion of the tips

PR #417, opened 2020-05-10
TipManager & ValueObject factory

jonastheis:
Adds TipManager and ValueObject factory. The TipManager maintains a simple flat map of all currently liked tips and gets filled by the value transfers plugins that listens to `value object liked` and `value object disliked` events. 

Relates to #208


dapps/valuetransfers/packages/tipmanager/tipmanager.go
```diff
@@ -0,0 +1,83 @@
...
+	// select 2 distinct tips if there's more than 1 tip available
+	for parent1ObjectID == parent2ObjectID && t.tips.Size() > 1 {
+		parent2ObjectID = t.tips.RandomEntry().(payload.ID)
+	}
+
+	return
+}
+
+// TipCount returns the total liked tips.
+func (t *TipManager) TipCount() int {
```
hmoog, line 81: I would maybe rename TipCount() to Size() as that's less repetitive than **tip**Manager.**Tip**Count() and also the go idiomatic name for containers.

PR #421, opened 2020-05-11
Use a background worker to handle missing messages on the base layer

luca-moser:
The Tangle of the base layer would spawn separate goroutines per missing message which would check for a given time frame whether a message became available or not. If it doesn't become available, the message and its future cone (approver references) is deleted.

This PR fixes the spawning of a new goroutine per missing message by using a single background worker which is registered in the message layer plugin. Additionally this fixes the problem of the storage being accessed after the corresponding ObjectStorages have been shutdown, since the priority of the "missing mesages monitor" is higher than that of the Tangle's.

### DIFF
--- a/packages/binary/messagelayer/tangle/tangle.go
+++ b/packages/binary/messagelayer/tangle/tangle.go
@@ -34,7 +34,7 @@ type Tangle struct {
 
 	storeMessageWorkerPool async.WorkerPool
 	solidifierWorkerPool   async.WorkerPool
-	cleanupWorkerPool      async.WorkerPool
+	shutdown               chan struct{}
 }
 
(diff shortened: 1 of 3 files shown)


hmoog:
I really don't understand why you hate the time.AfterFunc so much :P This code is so horribly inefficient. A prefix iteration without a set prefix requires the DB to load the whole LSM tree level into memory, which means gigabytes of data in ram for badger for example. And even an index based db engine will be extremely slow.

We already know which missing keys have to be checked and we just need to schedule checks in recurring intervals. Why would you want to scan the whole database to find entries that can just be scheduled using a simple time.AfterFunc?

Sometimes I really don't understand you Luca :P

PR #422, opened 2020-05-13
Fix entropy of prng

capossele:
This PR fixes th entropy of the prng based on the Unix timestamp by hashing the timepoint before converting it to a float64

### DIFF
--- a/packages/prng/unix_ts_rng.go
+++ b/packages/prng/unix_ts_rng.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/binary"
 	"time"
+
+	"github.com/dchest/blake2b"
 )
 
 // TimeSourceFunc is a function which gets an understanding of time in seconds resolution back.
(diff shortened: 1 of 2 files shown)


packages/prng/unix_ts_rng.go
```diff
@@ -62,12 +64,17 @@ func (utrng *UnixTimestampPrng) send() {
...
 	if err := binary.Write(buf, binary.LittleEndian, timePoint); err != nil {
 		panic(err)
 	}
-	pseudoR := float64(binary.BigEndian.Uint64(buf.Bytes()[:8])>>11) / (1 << 53)
+
+	// add entropy
+	h := blake2b.Sum256(buf.Bytes())
+
+	// convert to float64
+	pseudoR := float64(binary.BigEndian.Uint64(h[:8])>>11) / (1 << 53)
```
Wollac, line 77: How about `pseudoR := rand.New(rand.NewSource(timePoint)).Float64()`?
Shouldn't that have the the same effect?

packages/prng/unix_ts_rng.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/binary"
 	"time"
+
+	"github.com/dchest/blake2b"
```
Wollac, line 8: Why aren't we using the default `"golang.org/x/crypto/blake2b"`?

packages/prng/unix_ts_rng.go
Wollac: Well, not _exactly_ the same, but you know what I mean :stuck_out_tongue:

capossele: good call, I've updated the code to do that

packages/prng/unix_ts_rng.go
capossele: Now we don't need that dependency anymore :P

PR #423, opened 2020-05-13
Speedy Docker network

jonastheis:
Significantly speeds up building time for the Docker network especially when developing and a recurring rebuild is often necessary. A separate builder container makes uses of the Go cache and also caches dependencies whereas the actual containers of the network mount the same named volume aka cache and execute the binary with the given config, respectively.


jkrvivian:
lgtm!

PR #429, opened 2020-05-18
Docker aarch64

antonionardella:
Added /docker folder with 3 Dockerfiles, based on the underlying architecture:
- amd64
- arm64
- armhf

Retained the hash of the golang container for every arch.
The arm64 container builds correctly on ubuntu 20.04 on Raspberry Pi 4/4GB RAM

Big thank you to @nuriel77 for his support and the goshimmer-playbook!


luca-moser:
Hey Antonio, instead of having different Dockerfiles with pretty much the same content except the arch during the compilation step, we could simply use https://mirailabs.io/blog/multiarch-docker-with-buildx/ to build the images via `buildx` which builds the appropriate images for the different archs.

nuriel77:
While buildx is nice, I think having multiple Dockerfiles is the easiest for users, sparing them the need to follow the instructions to enable buildx properly.

Also, the current goshimmer Dockerfiles point to specific docker images by noting their hash tag, those point to images of specific architectures.
I am not sure that it will work that way with buildx, as you're supposed to say `FROM golang:alpine AS builder` rather than point to a specific image hash tag.

nuriel77:
As explained earlier, the reason for a Dockerfile per platform is to make it easier for the users so they don't need to pass a build argument containing the digest of the image they require for their platform.

Alternatively, I thought of another idea that might be a middle way, but still requires --build-arg on the build command:

Maintain a file with the digests in the docker directory (e.g. `image_digests`):
```
arm64=944405641f9fb0f322be1dfc4685b916df2de3df54525cf80822f8a0529f636f
armhf=f21086d6d97214e2997d927a8f131b0eb520f935e7a092abc1730edeafd02255
amd64=fc7e7c9c4b0f6d2d5e8611ee73b9d1d3132750108878517bbf988aa772359ae4
```

Then, for all architectures other than amd64, the build command will have to be:
```
docker build -t goshimmer:test --build-arg IMAGE_TAG=$(grep armhf image_digests | cut -d= -f2) .
```

For default (amd64) it will be the simple command without any build args because we can set a default ARG in the Dockerfile to point to the arm64 digest, then simply run:
```
docker build -t goshimmer:test .
```

The added complexity comes because of the image digest pinning, which has been done for security purposes.

luca-moser:
Ok, we are fine with having the `Dockerfile.amd64` and `Dockerfile.arm64` files. However, could you please remove the `armhf` one since to our knowledge, there's no 64-bit OS for Raspberries < 4?

luca-moser:
Any updates?

luca-moser:
If you think we should provide Dockerfiles for the other archs then please reopen the issue or fix what we've mentioned in here.

antonionardella:
Sorry, I was not notified by GH.
armhf has been removed

PR #444, opened 2020-05-30
Add panic if backgroundWorker fails

capossele:
This PR adds a check on each error returned from backgroundWorker and panic if != nil


capossele:
> Now, we use for `daemon.BackgroundWorker`:
> 
> * `panic`
> * `log.Panic`
> * `log.Error`
> 
> I think this needs to be cleaned up.
> If it is something that should really trigger a shutdown of the node we should use `log.Panic` or if it is just a very prominent error we should use `log.Error` all the time.

How about just using `log.Panicf` everywhere?

PR #449, opened 2020-06-04
sort enabled/disabled plugins list in info api

### LINKED ISSUE #446: Fix the enabled/disabled plugins info API
Screenshot 2020-06-04 at 12 34 44

acha-bill:
Sorts `enabledPlugins` and `disabledPlugins` list in `/info` endpoint output.

Fixes #446


acha-bill:
Then we may need to update `hive.go@node/node.go#plugins` from `map[string]int`  to `map[string]*Plugin` in order to expose the Plugin.

Then fix code in goshimmer and hornet that will be broken by that change.
Is this something we want to do?

jonastheis:
As discussed this is fine.

PR #469, opened 2020-06-12
Add PoW package

Wollac:
This PR provides generic PoW functionality:
- the nonce will be appended to the provided byte slice message
- any `hash.Hash` function can be used
- mining until the targeted number of leading zeros is reached
- parallel mining using go routines

### DIFF
--- a/packages/pow/pow.go
+++ b/packages/pow/pow.go
@@ -133,7 +133,7 @@ func (w *Worker) worker(msg []byte, startNonce uint64, target int, done *uint32,
 		}
 		asAnInt.SetBytes(digest)
 		leadingZeros := 8*w.hash.Size() - asAnInt.BitLen()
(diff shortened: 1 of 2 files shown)


packages/pow/pow.go
```diff
@@ -0,0 +1,151 @@
...
+		// write nonce in the buffer
+		binary.BigEndian.PutUint64(buf[len(msg):], nonce)
+
+		digest, err := w.sum(buf)
+		if err != nil {
+			return 0, err
+		}
+		asAnInt.SetBytes(digest)
+		leadingZeros := 8*w.hash.Size() - asAnInt.BitLen()
+		if leadingZeros > target {
```
luca-moser, line 136:
```suggestion
		if leadingZeros >= target {
```

PR #470, opened 2020-06-12
Fix enabled/disabled plugins spelling

capossele:


plugins/webapi/info/plugin.go
```diff
@@ -92,9 +92,9 @@ type Response struct {
 	// public key of the node encoded in base58
 	PublicKey string `json:"publickey,omitempty"`
```
Wollac, line 93: Shouldn't it also be `json:"publicKey,omitempty"` if everything else is small camelCase?

PR #471, opened 2020-06-12
Feat: Added Signature checks for value transfers to MessageParser

hmoog:


dapps/valuetransfers/packages/tangle/signature_filter.go
```diff
@@ -0,0 +1,90 @@
...
+	result = &SignatureFilter{}
+
+	return
+}
+
+// Filter get's called whenever a new message is received. It first checks if the message contains a value Payload and
+// then verifies the signature.
+func (filter *SignatureFilter) Filter(message *message.Message, peer *peer.Peer) {
+	filter.workerPool.Submit(func() {
+		if valuePayload := message.Payload(); valuePayload.Type() == payload.Type {
```
jonastheis, line 36: I think this can be more go idiomatic and much simpler to read from top to bottom if you don't nest the if statements but simply return with an error message. If all checks are passed `filter.getAcceptCallback()(message, peer)` is called at the very end of the function.

PR #472, opened 2020-06-13
dApp network delay

jonastheis:
This PR introduces a little dApp that will help us to understand actual network delay and clock skew. It listens for messages of its type and sends information about receiving time (local clock) to our remote log functionality.

It can also serve as an example how to write a dApp with GoShimmer.


m-delvalle:
> network delay and cock skew

"clock" 😉

jonastheis:
Oops..! That's a rather unfortunate typo 😆 
Fixed it.

PR #476, opened 2020-06-15
feat: Add inclusion states to webapis

jkrvivian:


plugins/webapi/value/utils/transaction_handler.go
```diff
@@ -64,6 +64,7 @@ type Balance struct {
 
 // InclusionState represents the different states of an OutputID
 type InclusionState struct {
+	Solid     bool `json:"confirmed,omitempty"`
```
jonastheis, line 67:
```suggestion
	Solid     bool `json:"solid,omitempty"`
```

plugins/webapi/value/unspentoutputs/handler.go
```diff
@@ -47,9 +46,10 @@ func Handler(c echo.Context) error {
 					ID:       id.String(),
 					Balances: b,
 					InclusionState: utils.InclusionState{
-						Confirmed: true,
-						Conflict:  false,
-						Liked:     true,
+						Solid:     output.Solid(),
+						Confirmed: output.Confirmed(),
+						Conflict:  output.Rejected(),
```
jonastheis, line 51: I'm not sure if `conflict` is a good choice here. A transaction can be conflicted but liked and confirmed if we voted on it. So maybe just stick to `rejected`?

plugins/webapi/value/utils/transaction_handler.go
jkrvivian: :stuck_out_tongue:

PR #511, opened 2020-06-21
Feat: Add PoW for messages

Wollac:
- Add `nonce` to the message
- Handle PoW in the `MessageFactory`
- Messages are created statically, signature, PoW must be computed before creation, e.g. using `MessageFactory`
- Add plugin for dynamic PoW configuration
- Default PoW difficulty is set to `22`, or `4` for the tests


Wollac:
> The PoW filter should not reside in the plugin but in a package, so it is possible to instantiate different filters with different settings for tests and for spinning up "virtual nodes" in simulations.

If we really want to expose the `PowFilter` it must not only take the difficulty, but also the pow Worker. In that case this would also (implicitly) make the hash configurable.

hmoog:
> > The PoW filter should not reside in the plugin but in a package, so it is possible to instantiate different filters with different settings for tests and for spinning up "virtual nodes" in simulations.
> 
> If we really want to expose the `PowFilter` it must not only take the difficulty, but also the pow Worker. In that case this would also (implicitly) make the hash configurable.

Yeah, I guess that would be good.

Wollac:
rebased on current `develop`

plugins/pow/parameters.go
```diff
@@ -0,0 +1,22 @@
+package pow
+
+import (
+	"time"
+
+	flag "github.com/spf13/pflag"
+)
+
+const (
+	// CfgPOWDifficulty defines the config flag of the PoW difficulty
```
luca-moser, line 10:
```suggestion
	// CfgPOWDifficulty defines the config flag of the PoW difficulty.
```

plugins/pow/parameters.go
```diff
@@ -0,0 +1,22 @@
...
+	CfgPOWDifficulty = "pow.difficulty"
+	// CfgPOWNumThreads defines the config flag of the number of threads used to do the PoW
```
luca-moser, line 12:
```suggestion
	// CfgPOWNumThreads defines the config flag of the number of threads used to do the PoW.
```

plugins/pow/parameters.go
```diff
@@ -0,0 +1,22 @@
...
+	CfgPOWNumThreads = "pow.numThreads"
+	// CfgPOWTimeout defines the config flag for the PoW timeout
```
luca-moser, line 14:
```suggestion
	// CfgPOWTimeout defines the config flag for the PoW timeout.
```

plugins/pow/plugin.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/iotaledger/goshimmer/plugins/messagelayer"
+	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/hive.go/node"
+)
+
+// PluginName is the name of the PoW plugin.
+const PluginName = "PoW"
+
+var (
+	// Plugin is the plugin instance of the port check plugin.
```
luca-moser, line 15:
```suggestion
	// Plugin is the plugin instance of the PoW plugin.
```

plugins/pow/pow.go
```diff
@@ -0,0 +1,81 @@
...
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/iotaledger/goshimmer/packages/pow"
+	"github.com/iotaledger/goshimmer/plugins/config"
+	"github.com/iotaledger/hive.go/logger"
+	_ "golang.org/x/crypto/blake2b" // required by crypto.BLAKE2b_512
+)
+
```
luca-moser, line 16: Consider making a constant denoting the amount of bytes the nonce takes then use that in the tests where you cut off message bytes to only the PoW relevant bytes.

PR #526, opened 2020-06-22
Removes periodic badgerdb gc & adds dirty/health flag to db

### LINKED ISSUE #520: Only do BadgerDB garbage collection on startup/shutdown
So we don't get random warnings in the log during runtime.

### LINKED ISSUE #521: Instruct the user to delete the DB if BadgerDB can't be opened
If for whatever reason we're unable to instantiate the BadgerDB, then instruct the user on the console to delete the database folder.

### LINKED ISSUE #522: Introduce a 'dirty' flag to the database
When the node starts up, it flags a boolean as dirty and clears it up on shutdown. This makes it possible to see whether a node was shutdown correctly. If a node does not shutdown correctly, data in the caches might not have been flushed correctly, leading to inconsistent state in the database.

In case the dirty flag is set to true, the user should be instructed to delete the database.

Have a look at HORNET where we do this exact thing.

luca-moser:
* When the node starts up it sets a key in the database marking it as dirty.
* When the node shuts down it deletes the key and therefore marks the node as successfully shutdown.
* Run database garbage collection on the configured store up on configure() and shutdown right before the close()
* Tell the user to delete the db folder if the store can't even be initialized.

Closes #520
Closes #521
Closes #522

### DIFF
--- a/plugins/database/plugin.go
+++ b/plugins/database/plugin.go
@@ -62,11 +62,10 @@ func configure(_ *node.Plugin) {
 	store := Store()
 	configureHealthStore(store)
 
-	err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion}))
-	if errors.Is(err, ErrDBVersionIncompatible) {
-		log.Panicf("The database scheme was updated. Please delete the database folder.\n%s", err)
-	}
-	if err != nil {
+	if err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion})); err != nil {
+		if errors.Is(err, ErrDBVersionIncompatible) {
+			log.Panicf("The database scheme was updated. Please delete the database folder.\n%s", err)
+		}
 		log.Panicf("Failed to check database version: %s", err)
 	}
 
@@ -84,7 +83,7 @@ func configure(_ *node.Plugin) {
 	}
 
 	// we open the database in the configure, so we must also make sure it's closed here
-	if err = daemon.BackgroundWorker(PluginName, closeDB, shutdown.PriorityDatabase); err != nil {
+	if err := daemon.BackgroundWorker(PluginName, closeDB, shutdown.PriorityDatabase); err != nil {
 		log.Panicf("Failed to start as daemon: %s", err)
 	}
 
--- a/plugins/database/plugin.go
+++ b/plugins/database/plugin.go
@@ -3,12 +3,10 @@ package database
 
 import (
 	"errors"
-	"fmt"
 	"sync"
 	"time"
 
 	"github.com/iotaledger/goshimmer/packages/database"
-	"github.com/iotaledger/goshimmer/packages/database/prefix"
 	"github.com/iotaledger/goshimmer/packages/shutdown"
 	"github.com/iotaledger/goshimmer/plugins/config"
 	"github.com/iotaledger/hive.go/daemon"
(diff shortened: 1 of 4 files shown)


plugins/database/plugin.go
```diff
@@ -52,7 +52,7 @@ func createStore() {
 		db, err = database.NewDB(dbDir)
 	}
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal(fmt.Sprintf("Unable to open the database, please delete the database folder. Error: %s", err))
```
Wollac, line 55: Use `log.Fatalf`

plugins/database/plugin.go
```diff
@@ -61,45 +61,57 @@ func createStore() {
...
+	configureHealthStore(store)
 
-	err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion}))
-	if errors.Is(err, ErrDBVersionIncompatible) {
-		log.Panicf("The database scheme was updated. Please delete the database folder.\n%s", err)
-	}
-	if err != nil {
+	if err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion})); err != nil {
+		if errors.Is(err, ErrDBVersionIncompatible) {
+			log.Panicf("The database scheme was updated. Please delete the database folder.\n%s", err)
```
Wollac, line 68: Above we use `Fatal` and here `Panic` is there a particular reason for that?

plugins/database/plugin.go
```diff
@@ -61,45 +61,57 @@ func createStore() {
...
+		MarkDatabaseUnhealthy()
+	}); err != nil {
 		log.Panicf("Failed to start as daemon: %s", err)
 	}
-}
 
-func run(_ *node.Plugin) {
-	if err := daemon.BackgroundWorker(PluginName+"[GC]", runGC, shutdown.PriorityBadgerGarbageCollection); err != nil {
+	// we open the database in the configure, so we must also make sure it's closed here
+	if err := daemon.BackgroundWorker(PluginName, closeDB, shutdown.PriorityDatabase); err != nil {
```
Wollac, line 87: Can we join the two `BackgroundWorker`?

plugins/database/plugin.go
Wollac: Also the has a `\n`

plugins/database/plugin.go
```diff
@@ -61,45 +61,57 @@ func createStore() {
...
 	// assure that the store is initialized
 	store := Store()
+	configureHealthStore(store)
 
-	err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion}))
-	if errors.Is(err, ErrDBVersionIncompatible) {
-		log.Panicf("The database scheme was updated. Please delete the database folder.\n%s", err)
-	}
-	if err != nil {
+	if err := checkDatabaseVersion(store.WithRealm([]byte{prefix.DBPrefixDatabaseVersion})); err != nil {
```
Wollac, line 66: Should we also use the Health realm for the database?

plugins/database/plugin.go
luca-moser: Changed to log.Fatalf

plugins/database/plugin.go
luca-moser: Made it use Fatal everywhere

plugins/database/plugin.go
luca-moser: Joined them

plugins/database/plugin.go
luca-moser: Renamed the DatabaseVersion prefix to Health

PR #527, opened 2020-06-22
Fix graceful shutdown

### LINKED ISSUE #504: FPC voter server doesn't shuttdown cleanly sometimes
The FPC voter server sometimes is unable to shutdown cleanly within a one minute time period.

capossele:
This PR fixes #504 .
It was actually the client-analysis not shutting-down if it was not able to connect to the server-analysis.

### DIFF
--- a/packages/vote/net/server.go
+++ b/packages/vote/net/server.go
@@ -3,6 +3,7 @@ package net
 import (
 	"context"
 	"net"
+	"sync"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/iotaledger/goshimmer/packages/metrics"
@@ -33,6 +34,7 @@ type VoterServer struct {
 	opnRetriever       OpinionRetriever
 	bindAddr           string
 	grpcServer         *grpc.Server
+	grpcServerMutex    sync.RWMutex
 	netRxEvent         *events.Event
 	netTxEvent         *events.Event
 	queryReceivedEvent *events.Event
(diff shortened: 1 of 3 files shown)


packages/vote/net/server.go
```diff
@@ -71,12 +73,16 @@ func (vs *VoterServer) Run() error {
 		return err
 	}
 
+	vs.grpcServerMutex.Lock()
 	vs.grpcServer = grpc.NewServer()
+	vs.grpcServerMutex.Unlock()
```
Wollac, line 78: i am not 100% sure how grpc works, but I guess the `GracefulStop` only makes sense after a `Serve`. Thus, I would just defer the `Unlock` to also protect the register and serve.

packages/vote/net/server.go
```diff
@@ -71,12 +73,16 @@ func (vs *VoterServer) Run() error {
...
 	RegisterVoterQueryServer(vs.grpcServer, vs)
 
 	return vs.grpcServer.Serve(listener)
 }
 
 func (vs *VoterServer) Shutdown() {
+	vs.grpcServerMutex.RLock()
+	defer vs.grpcServerMutex.RUnlock()
 	vs.grpcServer.GracefulStop()
```
Wollac, line 87:
```suggestion
	if vs.grpcServer != nil {
		vs.grpcServer.GracefulStop()
	}
```

packages/vote/net/server.go
capossele: Actually what needs to be protected here is only the assignment, because it might happen that the Shutdown() is getting called by a different goroutine, while we are assigning the pointer.

Wollac: But what would happen now when both are called at the same time? `GracefulStop` could be completely executed and then we reach the `vs.grpcServer.Serve()` in `Run()`. Is then the server running although we called `Shutdown` at the same time?

capossele: Ok, I've removed the initial goroutine, so that the shutdown gets called only after the assignment.

PR #548, opened 2020-06-25
Reduce solidifier worker pool capacity

### LINKED ISSUE #525: Investigate synchronization errors
After starting a node and letting it sync in our little test net it does not change its status to synchronized even though the MPS rate indicates that it is finished with the sync (only 2 MPS if the bootstrapping nodes are shown).

capossele:
Fixes #525

### DIFF
--- a/packages/binary/messagelayer/tangle/tangle.go
+++ b/packages/binary/messagelayer/tangle/tangle.go
@@ -66,7 +66,7 @@ func New(store kvstore.KVStore) (result *Tangle) {
 		Events: *newEvents(),
 	}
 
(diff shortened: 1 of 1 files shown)


Wollac:
How does this relate to https://github.com/iotaledger/goshimmer/pull/549?
Is this one still needed? Does it need to be merged first?

capossele:
> How does this relate to #549?
> Is this one still needed? Does it need to be merged first?

I don't think that the order matters. If you prefer I can include this change into #549

Wollac:
> I don't think that the order matters. If you prefer I can include this change into #549
Yeah, I think that will be better, than we have one PR doing all this including upgrading hive.go

PR #553, opened 2020-06-25
Create view on the dashboard showing information about an address

### LINKED ISSUE #550: Create view on the dashboard showing information about an address
We currently have a route which is supposed to display address information in the explorer, however, the server-side implementation isn't actually there, so linking to any `/explorer/address/<x>` will result in a 404.

The page should show appropriate 404 errors and if the address exists the outputs with their value, balance and spent state per address.

capossele:
Fixes #550 
Screenshot 2020-06-25 at 20 11 27


plugins/dashboard/explorer_routes.go
```diff
@@ -46,8 +50,13 @@ func createExplorerMessage(msg *message.Message) (*ExplorerMessage, error) {
 
 // ExplorerAddress defines the struct of the ExplorerAddress.
 type ExplorerAddress struct {
-	// Messagess hold the list of *ExplorerMessage.
-	Messages []*ExplorerMessage `json:"message"`
+	Address   string           `json:"address"`
+	OutputIDs []ExplorerOutput `json:"output_ids"`
+}
+
+type ExplorerOutput struct {
```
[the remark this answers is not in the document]
capossele: Who is a bad dog?! Have a look at the previous commit!

plugins/dashboard/explorer_routes.go
```diff
@@ -99,7 +109,7 @@ func setupExplorerRoutes(routeGroup *echo.Group) {
 		search := c.Param("search")
 		result := &SearchResult{}
 
-		if len(search) < 81 {
+		if len(search) < 33 {
```
luca-moser, line 112: Since addresses and messages have different lengths of IDs, there's actually no reason to spin up two goroutines to query for them, since the search parameter is for one or the other.

Also consider using the `message.IdLength` or `address.Length` respectively to check for the length and what action to take.

plugins/dashboard/explorer_routes.go
```diff
@@ -142,8 +152,60 @@ func findMessage(messageID message.Id) (explorerMsg *ExplorerMessage, err error)
...
+func findAddress(strAddress string) (*ExplorerAddress, error) {
+
+	address, err := address.FromBase58(strAddress)
+	if err != nil {
+		return nil, fmt.Errorf("%w: address %s", ErrNotFound, strAddress)
+	}
+
+	outputids := make([]ExplorerOutput, 0)
+	// get outputids by address
+	for id, cachedOutput := range valuetransfers.Tangle().OutputsOnAddress(address) {
```
luca-moser, line 164: Consider using `valuetransfers.Tangle().OutputsOnAddress(<bla>).Consume()` so you don't have to do `defer`s inside a for loop.

Issue #538, opened 2020-06-24
Improve the message explorer view in the dashboard


Wollac:
![Dashboard-Message](https://user-images.githubusercontent.com/4930426/85533353-be7fcd00-b610-11ea-8910-7402f443f4e4.png)

Errors:
- The `Time` is always zero. This field should be removed or set to the `issuingTime` of the message.

Improvements:
- A message has the following additional fields, consider whether they should be displayed:
  - issuerPublicKey
  - issuingTime
  - sequenceNumber
  - (signature)

PR #554, opened 2020-06-25
Enriches the message dashboard page with additional fields

luca-moser:
* Adds issuer public key, issuance time, solidification time, signature and sequence number to the message dashboard view
* Fixes weird background worker name for the message layer
* Fixes a message getting loaded via the explorer routes not releasing the msg metadata and thus blocking shutdown
* Terminate the bootstrapper plugin first
* Adds missing port to the default bind address config value for the dashboard bind address

image

Closs #538


plugins/dashboard/frontend/src/app/components/ExplorerAddressResult.tsx
```diff
@@ -48,7 +48,7 @@ export class ExplorerAddressQueryResult extends React.Component<Props, any> {
                 msgsEle.push(
                     <ListGroup.Item key={msg.id}>
                         <small>
-                            {dateformat(new Date(msg.timestamp * 1000), "dd.mm.yyyy HH:MM:ss")} {' '}
+                            {dateformat(new Date(msg.solidification_timestamp * 1000), "dd.mm.yyyy HH:MM:ss")} {' '}
```
capossele, line 51: I don't think we need any solidification timestamp in the ExplorerAddress no?

luca-moser: That's the messages residing on the address but whatever, this code is going to be removed by your changes anyway.

PR #584, opened 2020-06-29
Bumps network and db versions

luca-moser:


plugins/autopeering/autopeering.go
```diff
@@ -26,7 +26,7 @@ import (
 // autopeering constants
 const (
 	ProtocolVersion = 0      // update on protocol changes
-	NetworkVersion  = "v0.2" // update on network changes
```
Wollac, line 29: :thinking:  Initially it was the idea to have this consistent with the goshimmer version, but now we want to have a new network for each DB reset.

Should we maybe add a suffix to `v0.2`? Otherwise we could also remove all this `v0.` and use an integer

luca-moser: I'll just change them both to '0.2.0'

luca-moser: Actually no, I'll just make it a number

plugins/autopeering/autopeering.go
```diff
@@ -25,16 +24,16 @@ import (
...
+	NetworkVersion  = 2 // update on network changes
 )
 
 var (
 	// ErrParsingMasterNode is returned for an invalid master node.
 	ErrParsingMasterNode = errors.New("cannot parse master node")
 
 	// networkID specifies the autopeering network identifier.
-	networkID = hash32([]byte(banner.AppVersion + NetworkVersion))
+	networkID = hash32([]byte{NetworkVersion})
```
Wollac, line 36: That is now a bit strange, we just need a unit32 for the networkID.
If the networkVersion is already a number we don't need to hash it.
Do we want to add something like `-develop` or `-snapshot` in the future?

PR #606, opened 2020-07-01
Double spend check

### LINKED ISSUE #603: Make it impossible to issue a double spend by issuing txs simultaneously via one node
The web API handler for issuing a transaction should be synchronized and block until the given transaction is booked. It should check in `ValidateTransactionToAttach` whether the consumer count of any input consumed is higher than 0. This prevents people from double spending using the same node.

### LINKED ISSUE #605: Quick fix value objects attaching to their conflicting value objects
In `IssueTransaction` check whether the returned value object tip's transactions are conflicting with the to be issued transaction.

hmoog:
Closes #603 
Closes #605


dapps/faucet/packages/faucet.go
```diff
@@ -80,7 +80,10 @@ func (f *Faucet) SendFunds(msg *message.Message) (m *message.Message, txID strin
 	}
 
 	// prepare value payload with value factory
-	payload := valuetransfers.ValueObjectFactory().IssueTransaction(tx)
+	payload, err := valuetransfers.ValueObjectFactory().IssueTransaction(tx)
+	if err != nil {
+		return nil, "", err
```
Wollac, line 85:
```suggestion
		return nil, "", fmt.Errorf("failed to issue transaction: %w", err)
```

PR #607, opened 2020-07-01
prevent double spend from same node

acha-bill:
Prevents double spend from the same node.
Synchronizes `sendTransaction` API handler and adds conflict validation in `ValidateTransactionToAttach`

#603


hmoog:
We already have a PR for this: https://github.com/iotaledger/goshimmer/pull/606 And the checks need to happen after the tip selection - we discussed this in voice

acha-bill:
ok

PR #629, opened 2020-07-06
adds a cli to override database dirty flag

### LINKED ISSUE #599: Add a CLI flag to override the dirty flag of the database
To be used in special situations where you need to unflag the dirty flag.

acha-bill:
# Description of change

Adds a CLI flag `database.dirty` that accepts a boolean (passed as a string) and sets the database dirty flag to the value specified.
If the string passed to the `database.dirty` cannot be converted to boolean, a warning is printed and the override is ignored.

Fixes #599 

## Type of change

Tested manually

## Change checklist

### DIFF
--- a/plugins/database/plugin.go
+++ b/plugins/database/plugin.go
@@ -79,10 +79,10 @@ func configure(_ *node.Plugin) {
 	}
 
 	if str := config.Node().GetString(CfgDatabaseDirty); str != "" {
-		val ,err := strconv.ParseBool(str)
+		val, err := strconv.ParseBool(str)
 		if err != nil {
 			log.Warnf("Invalid %s: %s", CfgDatabaseDirty, err)
-		} else if val  {
+		} else if val {
 			MarkDatabaseUnhealthy()
 		} else {
 			MarkDatabaseHealthy()
--- a/plugins/database/plugin.go
+++ b/plugins/database/plugin.go
@@ -3,6 +3,7 @@ package database
 
 import (
 	"errors"
+	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


plugins/database/plugin.go
```diff
@@ -77,6 +78,17 @@ func configure(_ *node.Plugin) {
 		log.Fatalf("Failed to check database version: %s", err)
 	}
 
+	if str := config.Node().GetString(CfgDatabaseDirty); str != "" {
+		val ,err := strconv.ParseBool(str)
```
jonastheis, line 82: The dog is right. It's not `gofmt`-ed. Best set up `goimports` as a file watcher in your IDE and this shouldn't happen at all.

plugins/database/plugin.go
```diff
@@ -77,6 +78,17 @@ func configure(_ *node.Plugin) {
...
+		if err != nil {
+			log.Warnf("Invalid %s: %s", CfgDatabaseDirty, err)
+		} else if val  {
+			MarkDatabaseUnhealthy()
+		} else {
+			MarkDatabaseHealthy()
+		}
```
jonastheis, line 89: Maybe this could be nested. Makes it a bit easier to read.

acha-bill: I think what is there currently is better than this nested version.

````
if err != nil {
    log.Warnf("Invalid %s: %s", CfgDatabaseDirty, err)
} else  {
    if val  {
        MarkDatabaseUnhealthy()
    } else {
        MarkDatabaseHealthy()
    }
}

PR #641, opened 2020-07-10
Add sendTransactionByJson API

### LINKED ISSUE #637: The sendTransaction endpoint should use a proper JSON object not serialized binary
## Description

The `sendTransaction` endpoint currently has a payload of `txn_bytes: base64 string` which is a serialized version of the data, this is not very RESTful.

## Motivation

The endpoint should receive a standard JSON payload.

## Requirements

The JSON payload should contain the same contents as the serialized binary

e.g.
```js
{
   "inputs": string[],
   "outputs": {
      "id": string,
      "balances": {
          "value": number,
          "color": string
      }[];
    }[],
    "data": string,
    "signatures": {
       "version": number,
       "publicKey": string,
       "signature": string
      }[]
}
```

capossele:
This PR fixes #637 by adding the API endpoint `sendTransactionByJson` with a POST request such as:

```
{
   "inputs": string[],
   "outputs": {
      "address": string,
      "balances": {
          "value": int,
          "color": string
      }[];
    }[],
    "data": string,
    "signatures": {
       "version": byte,
       "publicKey": string,
       "signature": string
      }[]
}
```

where each string MUST be `base58` encoded (with the exception for the `IOTA` color).


plugins/webapi/value/sendtransactionbyjson/handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+		outputs[address] = balances
+	}
+
+	// prepare transaction
+	tx := transaction.New(transaction.NewInputs(inputs...), transaction.NewOutputs(outputs))
+
+	// add data payload
+	if request.Data != "" {
+		data, err := base58.Decode(request.Data)
```
Wollac, line 127: Let's not overdo it with the base58 usage. Base58 is really ill-suited for long data as it requires BigInteger math over the entire data. The `Data` field should just be a `[]byte` then go will automatically decode/encode this as base64 which is much better for general data.

plugins/webapi/value/sendtransactionbyjson/handler.go
```diff
@@ -0,0 +1,234 @@
...
+	// ErrMalformedData defines a malformed data error.
+	ErrMalformedData = fmt.Errorf("malformed data")
+	// ErrMalformedColor defines a malformed color error.
+	ErrMalformedColor = fmt.Errorf("malformed color")
+	// ErrMalformedPublicKey defines a malformed publicKey error.
+	ErrMalformedPublicKey = fmt.Errorf("malformed publicKey")
+	// ErrMalformedSignature defines a malformed signature error.
+	ErrMalformedSignature = fmt.Errorf("malformed signature")
+	// ErrWrongSignature defines a wrong signature error.
+	ErrWrongSignature = fmt.Errorf("wropng signature")
```
jkrvivian, line 39: Typo, `wrong`

PR #644, opened 2020-07-13
sync beacon plugin

### LINKED ISSUE #639: Sync Beacon Plugin
- Create a plugin that uses the heartbeat messages of configurable nodes to figure out the sync status.
- Follow more nodes, and derive synced "percentage" from the array of sync statuses. (2/3 needs to be synced to be considered in sync)
- Check messages with configurable node public keys => if issuing time is recent (configurable time window), we are in sync to that node.
- Add an option to enable plugin (community sync beacon) but should be disabled by default.

In future: compare sync status to high mana nodes.

acha-bill:
# Description

Adds a sync beacon plugin where a node uses a list of beacon nodes to determine if it is synced or not.

A node becomes a sync beacon if it enables `--syncbeacon.primary = true`. Then it'll periodically emit the sync beacon to the message layer.

A node can follow a list of nodes as beacons. e.g `--syncbeacon.followNodes ="pubicKey1,publickKey2"`. The node will handle sync beacon payloads sent from these nodes. All others are discarded.

A node can be primary and also does not follow any other nodes. `--syncbeacon.primary = true --syncbeacon.followNodes=""`
Such a node will broadcast its sync status but will not depend on any other node.

If a node receives a sync beacon payload that is older than a configurable max time window, we consider the beacon to be desynced.
If the beacon is solidified, then the node is considered to be synced wrt the issuer of the beacon.
However, if not synced with more than 2/3 of its following nodes, its said to be desynced.

At every `syncbeacon.clearInterval`, nodes that have not been updated within a configurable time frame are set to be desynced.

This plugin is disabled by default.

Fixes #639

### DIFF
--- a/plugins/syncbeacon/payload.go
+++ b/plugins/syncbeacon/payload.go
@@ -108,17 +108,12 @@ func IsSyncBeaconPayload(p *Payload) bool {
 }
 
 func init() {
-	payload.RegisterType(Type, ObjectName, GenericPayloadUnmarshalerFactory(Type))
-}
-
-// GenericPayloadUnmarshalerFactory sets the generic unmarshaler.
-func GenericPayloadUnmarshalerFactory(payloadType payload.Type) payload.Unmarshaler {
-	return func(data []byte) (payload payload.Payload, err error) {
+	payload.RegisterType(Type, ObjectName, func(data []byte) (payload payload.Payload, err error) {
 		payload = &Payload{
-			payloadType: payloadType,
+			payloadType: Type,
 		}
 		err = payload.Unmarshal(data)
 
 		return
-	}
+	})
 }
--- a/plugins/syncbeacon/payload.go
+++ b/plugins/syncbeacon/payload.go
@@ -1,7 +1,6 @@
 package syncbeacon
 
 import (
-	"github.com/iotaledger/goshimmer/packages/binary/messagelayer/message"
 	"github.com/iotaledger/goshimmer/packages/binary/messagelayer/payload"
 	"github.com/iotaledger/hive.go/marshalutil"
 	"github.com/iotaledger/hive.go/stringify"
@@ -12,6 +11,7 @@ const (
 	ObjectName = "syncbeacon"
 )
 
+// Type is the type of the syncbeacon payload.
 var Type = payload.Type(200)
 
 // Payload represents the syncbeacon payload
@@ -71,7 +71,7 @@ func (p *Payload) Type() payload.Type {
 	return p.payloadType
 }
 
-// Status returns the sync sync status of the  Payload.
+// SyncStatus returns the sync sync status of the  Payload.
 func (p *Payload) SyncStatus() bool {
 	return p.syncStatus
 }
@@ -110,8 +110,8 @@ func (p *Payload) String() string {
 }
 
 // IsSyncBeaconPayload checks if the message is sync beacon payload.
-func IsSyncBeaconPayload(msg *message.Message) bool {
-	return msg.Payload().Type() == Type
+func IsSyncBeaconPayload(p *Payload) bool {
+	return p.Type() == Type
 }
 
 func init() {
--- a/plugins/syncbeacon/payload.go
+++ b/plugins/syncbeacon/payload.go
@@ -49,6 +49,10 @@ func FromBytes(bytes []byte, optionalTargetObject ...*Payload) (result *Payload,
 	if err != nil {
 		return
 	}
+	_, err = marshalUtil.ReadUint32()
+	if err != nil {
+		return
+	}
 	result.sentTime, err = marshalUtil.ReadInt64()
 	if err != nil {
 		return
@@ -74,9 +78,11 @@ func (p *Payload) SentTime() int64 {
 func (p *Payload) Bytes() []byte {
 	// initialize helper
 	marshalUtil := marshalutil.New()
+	objectLength := marshalutil.INT64_SIZE
 
 	// marshal the p specific information
-	marshalUtil.WriteUint32(p.payloadType)
+	marshalUtil.WriteUint32(Type)
+	marshalUtil.WriteUint32(uint32(objectLength))
 	marshalUtil.WriteInt64(p.sentTime)
 
 	// return result
@@ -99,15 +105,6 @@ func (p *Payload) String() string {
 // IsSyncBeaconPayload checks if the message is sync beacon payload.
 func IsSyncBeaconPayload(p *Payload) bool {
 	return p.Type() == Type
-}
-
-func init() {
-	payload.RegisterType(Type, ObjectName, func(data []byte) (payload payload.Payload, err error) {
-		payload = &Payload{}
-		err = payload.Unmarshal(data)
-
-		return
-	})
 }
 
 func init() {
--- a/plugins/syncbeacon/payload.go
+++ b/plugins/syncbeacon/payload.go
@@ -109,9 +109,7 @@ func IsSyncBeaconPayload(p *Payload) bool {
 
 func init() {
 	payload.RegisterType(Type, ObjectName, func(data []byte) (payload payload.Payload, err error) {
-		payload = &Payload{
-			payloadType: Type,
-		}
+		payload = &Payload{}
 		err = payload.Unmarshal(data)
 
 		return
--- a/plugins/syncbeacon/payload.go
+++ b/plugins/syncbeacon/payload.go
@@ -17,15 +17,13 @@ var Type = payload.Type(200)
 // Payload represents the syncbeacon payload
 type Payload struct {
 	payloadType payload.Type
-	syncStatus  bool
 	sentTime    int64
 }
 
 // NewSyncBeaconPayload creates a new syncbeacon payload
-func NewSyncBeaconPayload(status bool, sentTime int64) *Payload {
+func NewSyncBeaconPayload(sentTime int64) *Payload {
 	return &Payload{
 		payloadType: Type,
-		syncStatus:  status,
 		sentTime:    sentTime,
 	}
 }
@@ -51,10 +49,6 @@ func FromBytes(bytes []byte, optionalTargetObject ...*Payload) (result *Payload,
 	if err != nil {
 		return
 	}
-	result.syncStatus, err = marshalUtil.ReadBool()
-	if err != nil {
-		return
-	}
 	result.sentTime, err = marshalUtil.ReadInt64()
 	if err != nil {
 		return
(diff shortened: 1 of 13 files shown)


plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,197 @@
...
+	} else {
+		sync.MarkDesynced()
+	}
+}
+
+// broadcastSyncBeaconPayload broadcasts its sync status to its neighbors.
+func broadcastSyncBeaconPayload() {
+	syncBeaconPayload := NewSyncBeaconPayload(sync.Synced(), time.Now().UnixNano())
+	//msg := message.New(message.EmptyId, message.EmptyId, time.Now(), local.GetInstance().LocalIdentity().PublicKey(), 0, syncBeaconPayload, 0, ed25519.Signature{})
+	mgr.SendMessage(syncBeaconPayload.Bytes())
```
capossele, line 150: This is conceptually wrong. If you don't write the beacon into the message Tangle, then, nodes following this node will not be able to solidify anything from this beacon, thus, their sync status would be meaningless.

plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,217 @@
...
+		})
+	}))
+}
+
+// handlePayload handles the received payload. It does the following checks:
+// It checks that the issuer of the payload is a followed node.
+// The time that payload was sent is not greater than  CfgSyncBeaconMaxTimeWindowSec. If the duration is longer than CfgSyncBeaconMaxTimeWindowSec, we consider that beacon to be out of sync till we receive a newer payload.
+// More than syncPercentage of followed nodes are also synced, the node is set to synced. Otherwise, its set as desynced.
+func handlePayload(syncBeaconPayload *Payload, issuerPublicKey ed25519.PublicKey) {
+	issuerPublicKeyStr := issuerPublicKey.String()
```
jonastheis, line 113: You can save this conversion and avoid the following loop as I suggested in the other comment.

There's no real need for `beaconNodesPublicKeys` as a slice. You could create a `map[string][]byte` mapping the public key from string to its byte representation and then directly compare the bytes. As this map is only created/modified on startup it doesn't need to be locked on access. So the time holding a lock can be minimized.

plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,217 @@
...
+// broadcastSyncBeaconPayload broadcasts sends a sync beacon to the message layer
+func broadcastSyncBeaconPayload() {
+	syncBeaconPayload := NewSyncBeaconPayload(time.Now().UnixNano())
+	messagelayer.MessageFactory().IssuePayload(syncBeaconPayload)
+	log.Info("issued sync beacon")
+}
+
+// cleanupFollowNodes cleans up offline nodes by setting their sync status to false after a configurable time window.
+func cleanupFollowNodes() {
+	if len(beaconNodesPublicKeys) == 0 {
```
jonastheis, line 170: The background worker should not even be started in `run` if there is no beacon nodes configured. 
Similarly, the plugin should not attach to the message tangle if there's no beacon nodes configured. 

If you split the plugin this is a configuration error, i.e. if the plugin is enabled and then there should always be at least 1 beacon node configured for the sync checking part. So this could maybe even panic.

plugins/syncbeacon/payload.go
```diff
@@ -0,0 +1,124 @@
...
+	return
+}
+
+// Type returns the type of the Payload.
+func (p *Payload) Type() payload.Type {
+	return p.payloadType
+}
+
+// SentTime returns the time that payload was sent.
+func (p *Payload) SentTime() int64 {
```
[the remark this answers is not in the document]
acha-bill: I feel strongly that we should rely on the beacons own `sentTime`.

plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,217 @@
...
+				case <-ticker.C:
+					broadcastSyncBeaconPayload()
+				case <-shutdownSignal:
+					return
+				}
+			}
+		}, shutdown.PrioritySynchronization); err != nil {
+			log.Panicf("Failed to start as daemon: %s", err)
+		}
+	}
```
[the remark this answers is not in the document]
acha-bill: moved to another plugin.

plugins/syncbeacon/payload.go
```diff
@@ -0,0 +1,119 @@
...
+
+// IsSyncBeaconPayload checks if the message is sync beacon payload.
+func IsSyncBeaconPayload(p *Payload) bool {
+	return p.Type() == Type
+}
+
+func init() {
+	payload.RegisterType(Type, ObjectName, func(data []byte) (payload payload.Payload, err error) {
+		payload = &Payload{
+			payloadType: Type,
```
jonastheis, line 113: No need to set the `payloadType` here. Since you override it in line 48 in `FromBytes` anyway.

plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,69 @@
...
+	})
+	return plugin
+}
+
+// configure events
+func configure(_ *node.Plugin) {
+	log = logger.NewLogger(PluginName)
+}
+
+// broadcastSyncBeaconPayload broadcasts sends a sync beacon to the message layer
```
jonastheis, line 47:
```suggestion
// broadcastSyncBeaconPayload broadcasts a sync beacon via communication layer
```

plugins/syncbeacon/plugin.go
```diff
@@ -0,0 +1,69 @@
...
+func broadcastSyncBeaconPayload() {
+	syncBeaconPayload := NewSyncBeaconPayload(time.Now().UnixNano())
+	messagelayer.MessageFactory().IssuePayload(syncBeaconPayload)
```
jonastheis, line 50: It should check for an error while issuing the payload. If there's an error log it as `info`. If successful you can do something like this `log.Infof("issued sync beacon %s", msg.ID())`

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+// beaconSync represents the beacon payload and the sync status.
+type beaconSync struct {
+	payload *syncbeacon.Payload
+	synced  bool
+}
+
+func init() {
+	flag.StringSlice(CfgSyncBeaconFollowNodes, []string{"Gm7W191NDnqyF7KJycZqK7V6ENLwqxTwoKQN4SmpkB24", "9DB3j9cWYSuEEtkvanrzqkzCQMdH1FGv3TawJdVbDxkd"}, "list of trusted nodes to follow their sync status")
+	flag.Int(CfgSyncBeaconMaxTimeWindowSec, 10, "the maximum time window for which a sync payload would be considerable")
+	flag.Int(CfgSyncBeaconMaxTimeOfflineSec, 40, "the maximum time the node should stay without receiving updates")
```
jonastheis, line 51: not sure but meaning this?
```suggestion
	flag.Int(CfgSyncBeaconMaxTimeOfflineSec, 40, "the maximum time the node should stay synced without receiving updates")
```

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+
+// configure events
+func configure(_ *node.Plugin) {
+	pubKeys := config.Node().GetStringSlice(CfgSyncBeaconFollowNodes)
+
+	if len(pubKeys) == 0 {
+		log.Panicf("Follow node list cannot be empty: %w", ErrMissingFollowNodes)
+	}
+
+	log = logger.NewLogger(PluginName)
```
jonastheis, line 85: This line should be moved before line 82.

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+// More than syncPercentage of followed nodes are also synced, the node is set to synced. Otherwise, its set as desynced.
+func handlePayload(syncBeaconPayload *syncbeacon.Payload, issuerPublicKey ed25519.PublicKey) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	issuerPublicKeyStr := issuerPublicKey.String()
+	//check if issuer is in configured beacon follow list
+	if _, ok := currentBeacons[issuerPublicKeyStr]; !ok {
+		return
+	}
```
jonastheis, line 125: This should be moved out of the `handlePayload` function. 

In configure create an additional map: `publicKey(string)->publicKey([]byte)`
Then you don't need to grab the lock to do this check and save the expensive conversion to `base58` on every beacon message.

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+
+	_beaconSync := beaconSync{
+		payload: syncBeaconPayload,
+		synced:  true,
+	}
+
+	dur := time.Since(time.Unix(0, syncBeaconPayload.SentTime()))
+	if dur.Seconds() > float64(config.Node().GetInt(CfgSyncBeaconMaxTimeWindowSec)) {
```
jonastheis, line 133: Consider setting a local var once in configure for `float64(config.Node().GetInt(CfgSyncBeaconMaxTimeWindowSec))`

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+		log.Infof("syncbeacon received from %s is too old", issuerPublicKey)
```
jonastheis, line 134: Maybe it'd be useful to print the `msgID` here, too.

plugins/syncbeaconfollower/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	issuerPublicKeyStr := issuerPublicKey.String()
+	//check if issuer is in configured beacon follow list
+	if _, ok := currentBeacons[issuerPublicKeyStr]; !ok {
+		return
+	}
+
+	_beaconSync := beaconSync{
```
jonastheis, line 127: This unnecessarily allocates a new struct every time it is called. You can simply update the struct in the map. Then it's just a pointer assignment and setting a boolean.

`currentBeacons[issuerPublicKeyStr].payload = syncBeaconPayload`

plugins/syncbeacon/plugin.go
acha-bill: `messagelayer.MessageFactory().IssuePayload` does not return error. But yeah, we should log the ID

plugins/syncbeaconfollower/plugin.go
acha-bill: I blame my tea. :)

Issue #648, opened 2020-07-17
Reevaluate worker pool usages


jonastheis:
@capossele and me investigated the latest occurring deadlock and went down several rabbit holes. Finally, we discovered that our frequent usage of [`async.workerpool`](https://github.com/iotaledger/hive.go/blob/master/async/workerpool.go) in its default setting with number of workers equals number of CPU cores is fundamentally flawed. 

## Deadlock
The deadlock is caused by a neighbor dropping, getting the `gossip lock` and then waiting forever on the waitgroup in `Neighbor.Close()`. This waitgroup can't finish because the `readLoop` can't terminate because it is stuck submitting something into the `gossip inboxWorkerpool`. All the workers in `inboxWorkerpool` are stuck because they are waiting to submit something into other `workerpools` and the whole thing starts piling up through several layers of worker pools. All these workers are eventually stuck at the gossip lock leading to the overall deadlock. 

The deadlock has been fixed with #650

## Usage of worker pools
We should distinguish between 2 types of worker pools:

### 1. [`async.workerpool`](https://github.com/iotaledger/hive.go/blob/master/async/workerpool.go) 
This is based on [`ants`](https://github.com/panjf2000/ants) and is more like having a bunch of recyclable goroutines but seemingly without task queue. According to the documentation it looks like this
![image](https://user-images.githubusercontent.com/4181434/87758695-38b90100-c80d-11ea-9eae-bd116d077d81.png)

We are using ants (almost everywhere) in **blocking** mode and with only **`runtime.GOMAXPROCS(0)`**. This means that every `Submit()` into a workerpool potentially blocks because there are only a handful of goroutines working in the pool. This becomes especially evident on hardware with only 2CPU cores and fast bandwidth where there's much to handle but not enough routines to do so. This explains also the problems we were seeing on slower machines.

### 2. [`workerpool.Workerpool`](https://github.com/iotaledger/hive.go/blob/master/workerpool/workerpool.go)
This looks more like a traditional [producer-consumer implementation](https://en.wikipedia.org/wiki/Producer%E2%80%93consumer_problem) with fixed number of goroutines and an additional queue for tasks. This could be better suited for some types of actions.

## Conclusion
Usage of the ants worker pools should be reevaluated because too few goroutines lead to backpressure due to other goroutines trying to submit to the pool. Additionally, usage of non-blocking pools should be considered especially in the case of the `inboxWorkerpool` so that a connection to a neighbor can **always** be closed.

## TODOs

[...]

PR #650, opened 2020-07-17
Various sync related fixes

capossele:
- Adds workerpools load on Prometheus for debugging
- Sets the solidifier and store message workerpools to 1024
- Split the inboxWorker pool to message and MessageRequest
- Remove the filter workerpools
- Changes the order of the mutex unlock in the MessageRequester
- Increase the docker shutdown timeout

Relates to #648

Screenshot 2020-07-17 at 16 21 13

### DIFF
--- a/packages/binary/messagelayer/messagerequester/messagerequester.go
+++ b/packages/binary/messagelayer/messagerequester/messagerequester.go
@@ -33,17 +33,18 @@ func New(optionalOptions ...Option) *MessageRequester {
 // StartRequest initiates a regular triggering of the StartRequest event until it has been stopped using StopRequest.
 func (requester *MessageRequester) StartRequest(id message.Id) {
 	requester.scheduledRequestsMutex.Lock()
-	defer requester.scheduledRequestsMutex.Unlock()
 
 	// ignore already scheduled requests
 	if _, exists := requester.scheduledRequests[id]; exists {
+		requester.scheduledRequestsMutex.Unlock()
 		return
 	}
 
 	// trigger the event and schedule the next request
 	// make this atomic to be sure that a successive call of StartRequest does not trigger again
-	requester.Events.SendRequest.Trigger(id)
 	requester.scheduledRequests[id] = time.AfterFunc(requester.options.retryInterval, func() { requester.reRequest(id) })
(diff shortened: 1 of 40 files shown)


packages/binary/messagelayer/messagerequester/messagerequester.go
```diff
@@ -33,17 +33,18 @@ func New(optionalOptions ...Option) *MessageRequester {
...
-	defer requester.scheduledRequestsMutex.Unlock()
 
 	// ignore already scheduled requests
 	if _, exists := requester.scheduledRequests[id]; exists {
+		requester.scheduledRequestsMutex.Unlock()
 		return
 	}
 
 	// trigger the event and schedule the next request
 	// make this atomic to be sure that a successive call of StartRequest does not trigger again
```
Wollac, line 44: These comments are no longer correct, as now the lock is released after the schedule. They are no longer "atomic".

capossele: BTW, do you think that it was better before?

Wollac: Ideally,
- the event should not be triggered in the lock,
- the reschedule should happen after the trigger (to account for potentially longer running events,
- should use `defer` for cleaner code and panic safety.

With this PR, it achieves the first point, previously it only achieved the last two :shrug: 
I thought about a solution for all points, but that seemed rather extensive for something that is going to be replaced anyways soon.

Honestly, I was so in the `defer` mindset that I didn't even consider this version :grin: 
I guess it makes sense, the reschedule is not relevant and replacing `defer` for parallel triggers seems reasonable...

packages/gossip/manager.go
```diff
@@ -3,23 +3,32 @@ package gossip
...
 
 const (
 	maxPacketSize = 64 * 1024
 )
 
+var (
+	messageWorkerCount     = runtime.GOMAXPROCS(0) * 4
+	messageWorkerQueueSize = 1000
+
+	messageRequestWorkerCount     = 1
```
jonastheis, line 28: I was thinking it might be smarter to have more than 1 worker here since a message might be loaded from disk which would block the whole thing for that time.

PR #687, opened 2020-07-29
Fix all linter issues

### LINKED ISSUE #678: Fix all linter warnings
Install reviewdog locally (https://github.com/reviewdog/reviewdog) and let it run with the same rules as defined in the `.golangci.yml` file (in the repo root).

acha-bill:
# Description of change

Lint the whole thing.
Ran `golangci-lint run` and fixed all errors.

`golangci-lint` version: v1.29.0
Also bumps `reviewdog/action-golangci-lint` to  v1.9
## Type of change

`golangci-lint run` does not return any issues.

Fixes #678


plugins/autopeering/autopeering.go
```diff
@@ -163,15 +162,6 @@ func start(shutdownSignal <-chan struct{}) {
 	lPeer.Database().Close()
 }
 
-func hash32(b []byte) uint32 {
```
lzpap, line 166: I suppose this was unused.

acha-bill: yes

packages/binary/datastructure/queue/queue.go
```diff
@@ -13,27 +14,32 @@ type Queue struct {
...
 
+// Capacity returns the capacity of the queue.
 func (queue *Queue) Capacity() int {
 	queue.mutex.Lock()
 	defer queue.mutex.Unlock()
 
 	return queue.capacity
 }
 
+// Offer adds an element ot the queue and returns true.
```
acha-bill, line 41:
```suggestion
// Offer adds an element to the queue and returns true.
```

PR #690, opened 2020-08-04
Static peering support

### LINKED ISSUE #547: Add static peering support
Enables us to write faster integration tests.

1. We could leverage the fact that autopeering and gossip are on separate layers by introducing a manual peered neighbors in the gossip manager. If the autopeering tries to connect with one of them, we can make the gossip to drop that connection and communicate this drop to the autopeering (clearly by keeping up the manual peered connection)
2. These 2 separated neighborhoods should be transparent from the communication primitive perspective
3. We could decouple the neighbor selection part in Goshimmer so that we could only enable the peer discovery and the manual peering to achieve both deterministic topology as well as network knowledge for FPC

Implementation details:
1. Make it possible to turn off the neighbors selection in the autopeering plugin via configuration (`configureEvents()` function)
2. Add a separate list for manual neighbors to the gossip manager, replicate all logic from autopeering neighbors list (existing) to the manual neighbors list (new). The only exception is events from the autopeering layer don't affect the manual neighbors list (new). 
3. Make it possible to set manual peers in the gossip manager via configuration.

acha-bill:
Adds static peering support via API
 - `addPeers`: Accepts a list of nodes to peer with. Returns the number of nodes added.
 - `removePeers`: Accepts a list of nodes to disconnect. Returns the number of nodes removed.

Fixes #547

### DIFF
--- a/pluginmgr/webapi/plugins.go
+++ b/pluginmgr/webapi/plugins.go
@@ -10,6 +10,7 @@ import (
 	"github.com/iotaledger/goshimmer/plugins/webapi/info"
 	"github.com/iotaledger/goshimmer/plugins/webapi/message"
 	"github.com/iotaledger/goshimmer/plugins/webapi/spammer"
+	"github.com/iotaledger/goshimmer/plugins/webapi/staticpeering"
 	"github.com/iotaledger/goshimmer/plugins/webapi/tools"
 	"github.com/iotaledger/goshimmer/plugins/webapi/value"
 	"github.com/iotaledger/goshimmer/plugins/webauth"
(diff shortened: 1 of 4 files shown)


capossele:
Few questions:

- What happens if I set node A as neighbor via manual peering and then the autopeering tells me to drop this neighbor?
- Are we considering different lists of neighbors based on manual vs auto?
- More in general, how's the interaction between autopeering and manual peering being designed here?

acha-bill:
Good questions.

Ideally, we would not want autopeering to interfere with guys we have decided to statically peer with.
We would then need to maintain those 2 lists as you say; auto and manual.
```
// The Manager handles the connected neighbors.
type Manager struct {
 autoNeighbors   map[identity.Id]*Neighbor
 staticNeighbors map[identity.Id]*Neighbor
  ...
}
```

Autopeering `events` for a peer could then be checked against the manual list and if found, ignored.
Then add `static/neighbors` end point to comlement `autopeering/neighbors` or join combine them.

PR #701, opened 2020-08-12
Make autopeering.NetworkVersion configurable

capossele:
Makes autopeering network version configurable and adds the network version to the output of the API info

### DIFF
--- a/plugins/autopeering/autopeering.go
+++ b/plugins/autopeering/autopeering.go
@@ -195,6 +195,7 @@ func parseEntryNodes() (result []*peer.Peer, err error) {
 	return result, nil
 }
 
+// NetworkVersion returns the network version of the autopeering.
 func NetworkVersion() uint32 {
 	return networkVersion
 }
--- a/plugins/autopeering/autopeering.go
+++ b/plugins/autopeering/autopeering.go
@@ -24,7 +24,6 @@ import (
 // autopeering constants
 const (
 	ProtocolVersion = 0 // update on protocol changes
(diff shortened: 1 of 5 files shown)


plugins/portcheck/plugin.go
```diff
@@ -59,7 +60,7 @@ func checkAutopeeringConnection() {
 	defer conn.Close()
 
 	// create a new discovery server for the port check
-	disc := discover.New(local.GetInstance(), autopeering.ProtocolVersion, autopeering.NetworkVersion, discover.Logger(log))
+	disc := discover.New(local.GetInstance(), autopeering.ProtocolVersion, config.Node().GetUint32(autopeering.CfgNetworkVersion), discover.Logger(log))
```
jonastheis, line 63: I think it would be cleaner to define a variable `autopeering.NetworkVersion` (or even function) then you don't need to read from the config here.

plugins/autopeering/autopeering.go
```diff
@@ -92,7 +91,7 @@ func createPeerDisc() {
 	}
 	log.Debugf("Master peers: %v", masterPeers)
 
-	peerDisc = discover.New(local.GetInstance(), ProtocolVersion, NetworkVersion,
+	peerDisc = discover.New(local.GetInstance(), ProtocolVersion, config.Node().GetUint32(CfgNetworkVersion),
```
jonastheis, line 94: I think it would be cleaner to define a variable `autopeering.NetworkVersion` (or even function) then you don't need to read from the config here.

PR #702, opened 2020-08-12
Fix: missing message race condition

hmoog:
This PR fixes the logical race condition which caused the MissingMessageReceived event to not get triggered in some rare edge cases.


capossele (approved):
fantastic!

PR #704, opened 2020-08-17
Change event parameters to struct

### LINKED ISSUE #478: Change event parameters to struct
Define a struct for each event, which essentially is the parameter with which we call it. Then if we need to add additional parameters to the event, we can just add a field to the struct without breaking  the code.

Example:
Instead of:
```
func OpinionCaller(handler interface{}, params ...interface{}) {
	handler.(func(id string, opinion Opinion))(params[0].(string), params[1].(Opinion))
}
```
Do this:
```
type OpinionEvent struct {
    ID string
    Opinion Opinion
    ...
}

func OpinionCaller(handler interface{}, params ...interface{}) {
	handler.(func(ev *OpinionEvent))(params[0].(*OpinionEvent))
}
```

jkrvivian:
# Description of change
Fix #478 

## Type of change
Enhancement (a non-breaking change which adds functionality)

## Change checklist

### DIFF
--- a/dapps/faucet/dapp.go
+++ b/dapps/faucet/dapp.go
@@ -124,11 +124,11 @@ func run(*node.Plugin) {
 }
 
 func configureEvents() {
-	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *tangle.CachedMessageEvent) {
-		defer cachedMessage.Message.Release()
-		defer cachedMessage.MessageMetadata.Release()
+	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessageEvent *tangle.CachedMessageEvent) {
+		defer cachedMessageEvent.Message.Release()
+		defer cachedMessageEvent.MessageMetadata.Release()
 
-		msg := cachedMessage.Message.Unwrap()
+		msg := cachedMessageEvent.Message.Unwrap()
 		if msg == nil || !faucetpayload.IsFaucetReq(msg) {
 			return
 		}
--- a/dapps/faucet/dapp.go
+++ b/dapps/faucet/dapp.go
@@ -124,11 +124,11 @@ func run(*node.Plugin) {
 }
 
 func configureEvents() {
-	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *message.CachedMessage, cachedMessageMetadata *tangle.CachedMessageMetadata) {
-		defer cachedMessage.Release()
-		defer cachedMessageMetadata.Release()
+	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *tangle.CachedMessage) {
+		defer cachedMessage.Message.Release()
+		defer cachedMessage.MessageMetadata.Release()
 
-		msg := cachedMessage.Unwrap()
+		msg := cachedMessage.Message.Unwrap()
 		if msg == nil || !faucetpayload.IsFaucetReq(msg) {
 			return
 		}
--- a/dapps/faucet/dapp.go
+++ b/dapps/faucet/dapp.go
@@ -124,7 +124,7 @@ func run(*node.Plugin) {
 }
 
 func configureEvents() {
-	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *tangle.CachedMessage) {
+	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *tangle.CachedMessageEvent) {
 		defer cachedMessage.Message.Release()
 		defer cachedMessage.MessageMetadata.Release()
 
--- a/dapps/valuetransfers/packages/tangle/events.go
+++ b/dapps/valuetransfers/packages/tangle/events.go
@@ -69,6 +69,35 @@ func (eventSource EventSource) String() string {
 	return [...]string{"EventSourceTangle", "EventSourceBranchManager"}[eventSource]
 }
 
+type CachedPayloadEvent struct {
+	Payload         *payload.CachedPayload
+	PayloadMetadata *CachedPayloadMetadata
+}
+
+type CachedTxnEvent struct {
+	Txn         *transaction.CachedTransaction
+	TxnMetadata *CachedTransactionMetadata
+}
+
+type CachedTxnBookEvent struct {
+	Txn         *transaction.CachedTransaction
+	TxnMetadata *CachedTransactionMetadata
+	Pending     bool
+}
+
+type ForkEvent struct {
+	Txn         *transaction.CachedTransaction
+	TxnMetadata *CachedTransactionMetadata
+	Branch      *branchmanager.CachedBranch
+	OutputIDs   []transaction.OutputID
+}
+
+type CachedAttachmentsEvent struct {
+	Txn         *transaction.CachedTransaction
+	TxnMetadata *CachedTransactionMetadata
+	Attachments *CachedAttachment
+}
+
 func newEvents() *Events {
 	return &Events{
 		PayloadAttached:        events.NewEvent(cachedPayloadEvent),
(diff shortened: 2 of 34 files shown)


dapps/faucet/dapp.go
```diff
@@ -124,11 +124,11 @@ func run(*node.Plugin) {
 }
 
 func configureEvents() {
-	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *message.CachedMessage, cachedMessageMetadata *tangle.CachedMessageMetadata) {
-		defer cachedMessage.Release()
-		defer cachedMessageMetadata.Release()
+	messagelayer.Tangle().Events.MessageSolid.Attach(events.NewClosure(func(cachedMessage *tangle.CachedMessageEvent) {
```
lzpap, line 127: I think it is a bit odd to call it `cachedMessage` if it is of type `*CachedMessageEvent`, that has a `cachedMessage` and a `cachedMessageMetadata` field. I would rather call it `cachedMessageEvent`, `eventStruct`, `eventParameter` or any combination of these to signal that it is a struct only for this particular event.

jkrvivian:
> What do you think about we have a single `cachedMsgEvent.Release()` that releases all cached props in the event?

That's a great idea!
Otherwise, we need to add `.Release()` in every Trigger function if new object storage is added to the event structure.

jkrvivian:
@acha-bill , consider using  `Consume` for the object storage, it will be double freed and cause panic if we use such release function. So let's stick to releasing them in each attached function.

packages/binary/messagelayer/messageparser/events.go
```diff
@@ -11,3 +15,41 @@ type Events struct {
...
+		MessageRejected: events.NewEvent(messageRejectedEvent),
+	}
+}
+
+func messageParsedEvent(handler interface{}, params ...interface{}) {
+	handler.(func(*MessageParsedEvent))(params[0].(*MessageParsedEvent))
+}
+
+func bytesRejectedEvent(handler interface{}, params ...interface{}) {
+	handler.(func(*BytesRejectedEvent, error))(params[0].(*BytesRejectedEvent), params[1].(error))
```
jonastheis, line 50: Just wondering about the design choice to not include the error in the event. Is there a reason to not do it? 
I've seen it before as well. Whatever the decision is it needs to be consistent.

packages/binary/messagelayer/messageparser/message_parser.go
```diff
@@ -28,17 +27,7 @@ func New() (result *MessageParser) {
...
-				handler.(func(*message.Message, *peer.Peer))(params[0].(*message.Message), params[1].(*peer.Peer))
-			}),
-			BytesRejected: events.NewEvent(func(handler interface{}, params ...interface{}) {
-				handler.(func([]byte, error, *peer.Peer))(params[0].([]byte), params[1].(error), params[2].(*peer.Peer))
-			}),
-			MessageRejected: events.NewEvent(func(handler interface{}, params ...interface{}) {
-				handler.(func(*message.Message, error, *peer.Peer))(params[0].(*message.Message), params[1].(error), params[2].(*peer.Peer))
-			}),
-		},
+		Events:         *newEvents(),
```
jonastheis, line 30: Looking at the `DRNG` or `ValueTangle` the `Events` are a pointer. Maybe that should be the case here, too?

packages/binary/messagelayer/messagerequester/messagerequester.go
```diff
@@ -28,14 +27,7 @@ func New(missingMessages []message.ID, optionalOptions ...Option) *MessageReques
...
 		options:           newOptions(optionalOptions),
-		Events: Events{
-			SendRequest: events.NewEvent(func(handler interface{}, params ...interface{}) {
-				handler.(func(message.ID))(params[0].(message.ID))
-			}),
-			MissingMessageAppeared: events.NewEvent(func(handler interface{}, params ...interface{}) {
-				handler.(func(message.ID))(params[0].(message.ID))
-			}),
-		},
+		Events:            *newEvents(),
```
jonastheis, line 30: Okay it seems that sometimes `Events` are a pointer and sometimes not... Maybe make it consistently a pointer?

packages/binary/messagelayer/tangle/tangle.go
```diff
@@ -206,10 +206,14 @@ func (tangle *Tangle) storeMessageWorker(msg *message.Message) {
...
+			MessageMetadata: cachedMsgMetadata})
 	}
 
-	tangle.Events.MessageAttached.Trigger(cachedMessage, cachedMsgMetadata)
+	tangle.Events.MessageAttached.Trigger(&CachedMessageEvent{
+		Message:         cachedMessage,
+		MessageMetadata: cachedMsgMetadata})
 
 	// check message solidity
 	tangle.solidifierWorkerPool.Submit(func() {
```
jonastheis, line 219: I can't comment on the actual line.

But `tangle.Events.MessageMissing.Trigger(messageID)` needs to be updated. Please make sure to search the full codebase for `.Trigger(` to check if every event trigger is update.

plugins/messagelayer/plugin.go
```diff
@@ -110,28 +109,28 @@ func configure(*node.Plugin) {
...
-		cachedMessageMetadata.Release()
-		cachedMessage.Consume(tipSelector.AddTip)
+	_tangle.Events.MessageSolid.Attach(events.NewClosure(func(cachedMsgEvent *tangle.CachedMessageEvent) {
+		cachedMsgEvent.MessageMetadata.Release()
+		cachedMsgEvent.Message.Consume(tipSelector.AddTip)
 	}))
 
-	MessageRequester().Events.MissingMessageAppeared.Attach(events.NewClosure(func(id message.ID) {
-		_tangle.DeleteMissingMessage(id)
+	MessageRequester().Events.MissingMessageAppeared.Attach(events.NewClosure(func(missingMessageAppeared messagerequester.MissingMessageAppearedEvent) {
```
jonastheis, line 132: This event might not be used anyway but shouldn't it be a pointer?

packages/binary/messagelayer/tipselector/events.go
```diff
@@ -14,6 +14,13 @@ type Events struct {
...
+func newEvents() *Events {
+	return &Events{
+		TipAdded:   events.NewEvent(messageIDEvent),
+		TipRemoved: events.NewEvent(messageIDEvent),
+	}
+}
+
 func messageIDEvent(handler interface{}, params ...interface{}) {
 	handler.(func(message.ID))(params[0].(message.ID))
 }
```
jonastheis, line 26: Should that be embedded in a struct? Or does it overcomplicate things?

jkrvivian:
@jonastheis I've searched the code with `.Trigger(` and `.Attach(`, but still, miss some of them. :stuck_out_tongue: 
Thanks for reviewing it so detailed!

packages/binary/messagelayer/messageparser/events.go
jkrvivian: I just have a feeling that the error shouldn't be the member of the struct, it should be the external checkpoint before we access the data fields.

dapps/valuetransfers/packages/tangle/events.go
```diff
@@ -69,6 +69,40 @@ func (eventSource EventSource) String() string {
...
 }
 
+// CachedPayloadEvent represents the parameters of cachedPayloadEvent
+type CachedPayloadEvent struct {
+	Payload         *payload.CachedPayload
+	PayloadMetadata *CachedPayloadMetadata
+}
+
+// CachedTxnEvent represents the parameters of cachedTxnEvent
+type CachedTxnEvent struct {
```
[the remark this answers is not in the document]
jkrvivian: Let's go for `transaction`!

packages/binary/messagelayer/tipselector/events.go
jkrvivian: yeah, for those events that are rather simple, like `messageIDEvent` should and only contains `message.ID`, we could just keep the way it was.

jonastheis (approved):

PR #707, opened 2020-08-21
Mana Implementation Proposal

lzpap:
# Description

This PR adds `Mana Implementation Proposal` document to the repo. It is intended as  high level overview and a starting point of discussion about the first version of mana in GoShimmer.


docs/mana_proposal.md
```diff
@@ -0,0 +1,100 @@
...
+Mana should be implemented as a separate plugin in GoShimmer.
+
+The `mana plugin` is responsible for:
+ - calculating mana from value transactions,
+ - keeping a log of the different mana values of all nodes,
+ - updating mana values in a timely fashion,
+ - respond to mana related queries from other modules,
+ - allowing certain modules to overwrite mana of nodes (punishing bad behavior).
+
+ The proposed mana plugin in should keep track of the different mana values of nodes and handle calculation and timely
```
acha-bill, line 48:
```suggestion
 The proposed mana plugin should keep track of the different mana values of nodes and handle calculation and timely
```

docs/mana_proposal.md
```diff
@@ -0,0 +1,100 @@
...
+ updates. Mana values are mapped to `nodeID`s and stored in a `map` data structure:
+ ```go
+baseMana1Map := map[<nodeID>[]byte]<baseMana1-value>float64, and so on...
+```
+ Base Mana 2, EMAs, Consensus and Access Mana records are updated every `t` interval with tickers in a dedicated `BackgrounWorker`.
```
acha-bill, line 53: From mana spec
>The Base Mana Snapshot State will only be updated when transactions are snapshotted
This is, therefore, a function of `snapshot` instead of a timer.

Or are we sticking to a timer for this first iteration?

docs/mana_proposal.md
```diff
@@ -0,0 +1,100 @@
...
+ - calculating mana from value transactions,
+ - keeping a log of the different mana values of all nodes,
+ - updating mana values in a timely fashion,
+ - respond to mana related queries from other modules,
+ - allowing certain modules to overwrite mana of nodes (punishing bad behavior).
+
+ The proposed mana plugin in should keep track of the different mana values of nodes and handle calculation and timely
+ updates. Mana values are mapped to `nodeID`s and stored in a `map` data structure:
+ ```go
+baseMana1Map := map[<nodeID>[]byte]<baseMana1-value>float64, and so on...
```
acha-bill, line 51: What do you think about we have a single map for all manas?
```
type Mana struct {
 BaseMana1 float
 etc..
}
manaMap := map[nodeId]Mana
````

File: docs/mana_proposal.md [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+
+ The mana plugin should expose utility funtions to other modules for:
+  1. Obtaining the list of N highest mana nodes.
+  2. Obtaining the mana value of a particular node.
+  3. Obtaining a list of currently known peers + their mana, sorted. Useful for knowing which high mana nodes are online.
+  4. Obtaining a list of neighbors and their mana.
+  5. Decreasing (overwriting) some form of mana because of detected misbehaving.
+  6. Obtaining the full mana maps for comparison with the perception of other nodes.
+
+  +1, consider adding new API routes for mana related infos.
```
acha-bill, line 66: Nice :)

docs/mana_proposal.md
lzpap: Base Mana Snapshot State concerns Base Mana 1 and 2 values and the time these values were last updated, right? In the first iteration we will not have a Mana Snapshot State as it implies you have the message/transaction `snapshotting` in place, which we obviously don't. Anyway, even when we will have this, the decaying parts plus the EMAs should be updated at regular time intervals.
So, mana snapshot file will be updated upon `snapshotting`, but derived values will have to be updated regularly regardless of `snapshotting`. That is why we need timers/tickers/whatever.

docs/mana_proposal.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Introduction
+
+Mana is a a reputation system for nodes within the IOTA network.
+
+Reputation is gained by contributing to the network, i.e. creating value transfers.
+Reputation is lost when not cooperating with the network: using up all the available bandwidth, advocating for
+"bad" transfers during voting, etc..
+
+As time passes, earned mana of a node decays to encourage keeping up the good behavior.
```
oliviasaa: "earned mana of a node decays to encourage keeping up the good behaviour" - this is only true for base mana 2, base mana 1 does not decay. So maybe this sentence can be misleading to some people.

docs/mana_proposal.md
```diff
@@ -0,0 +1,100 @@
...
+## Mana Calculation
+
+Mana is esentially the reputation score of a node in the IOTA network. Mana is calculated locally in each node, as a
+function that takes value transactions as input and produces the Base Mana Vector as output.
+It consists of Base Mana 1 and Base Mana 2. Given a value transaction, these are determined as follows:
+ 1. Base Mana 1 is revoked from the node that created the output(s) used as input(s) in the transaction, and is pledged to
+    the node creating the new output(s).
+ 2. Base Mana 2 is freshly created at the issuance time of the transaction, awarded to the node, but decays with time.
+
+Access and Consensus Mana (mana used by other node modules) are derived from Base Mana 1 and 2 and "smoothened" through
```
oliviasaa: At least to me, it's easier to think like this:

Each Eff Base Mana is the EMA of a Base Mana
Access and Consensus Mana are derived from Eff Base Mana 1 and 2

acha-bill (commented):
Why do we have to have 2 `BaseManaVector`s?
```
accessMana BaseManaVector
consensusMana BaseManaVector
```
And do the same booking operations in both. i.e
```
on TransactionConfirmed(tx):
   value = getValue(tx)
   bookAccessMana(nodeId,value)   -- updates base mana1
   bookConsensusMana(nodeId,value) -- updates base mana1
```
Effectively rendering them to be the same.
From my understanding, we only calculate access and consensus mana at the point when they are requested and so storing outdated mana is irrelevant.

I've also made some edits that you can find here: https://hackmd.io/@djOmrag0QyiRMeuLnDw7mA/Sy92YVm7P/edit
Restructured the document to be a bit easier to assimilate top-down and tried to be a bit more concrete in the design.

docs/mana_proposal.md
lzpap: good point, I will update it.

lzpap:
> Why do we have to have 2 `BaseManaVector`s?
> 
> ```
> accessMana BaseManaVector
> consensusMana BaseManaVector
> ```
> 
> And do the same booking operations in both. i.e
> 
> ```
> on TransactionConfirmed(tx):
>    value = getValue(tx)
>    bookAccessMana(nodeId,value)   -- updates base mana1
>    bookConsensusMana(nodeId,value) -- updates base mana1
> ```
> 
> Effectively rendering them to be the same.
> From my understanding, we only calculate access and consensus mana at the point when they are requested and so storing outdated mana is irrelevant.
> 
> I've also made some edits that you can find here: https://hackmd.io/@djOmrag0QyiRMeuLnDw7mA/Sy92YVm7P/edit
> Restructured the document to be a bit easier to assimilate top-down and tried to be a bit more concrete in the design.

Thanks @acha-bill , great stuff!
The reason why we have two `Base Vectors` is that a transaction can book one type of mana to node A, but book the other type to Node B. In this case, the two vectors will differ.
You are right about the `bookMana` functions, they are the same. Let's define them as a `bookMana` method on a `BaseManaVector`. The `getMana` functions on the other hand should be different, because we want to allow experimenting with the combination of EBM1 & EBM2 for Access and Consensus Mana.
We could also define `BaseManaVector` as an interface, and implement Access and Consensus Base Mana Vectors as types that implement this interface. Then we hide this calculation from outside of the package and the plugin doesn't need to know which mana vector it is dealing with.
What do you think?

oliviasaa:
Well, obviously the implementation is up to you guys, but this is not how I imagined when writing the specs. To me, the Base Mana Vector should consist on the 4 values of Base Mana for all the nodes. So if you book one type of Mana to node A and the other to node B, it's still the same Base Mana Vector, you just update it accordingly.

oliviasaa:
So if I book 1 Base Mana 1 to node A and  1 Base Mana 2 to node B, the vector would look like (suppose that both nodes had 0 Mana before that):
                           Node A               Node B               .... 
BM1                        1                           0                   ....
BM2                        0                           1                   ....
EBM1                     smth < 1                0                   ....
EBM2                        0                 smth <   1             ....

lzpap:
> Well, obviously the implementation is up to you guys, but this is not how I imagined when writing the specs. To me, the Base Mana Vector should consist on the 4 values of Base Mana for all the nodes. So if you book one type of Mana to node A and the other to node B, it's still the same Base Mana Vector, you just update it accordingly.

@oliviasaa okay, then I didn't understand it correctly.
So the two mana fields `accessMana` and `consensusMana` of a transaction are essentially `Base Mana 1` and `Base Mana 2` that this transaction generates and pledges to nodes?

acha-bill:
@lzpap yep. I second we use an interface there.
@oliviasaa true. But the node can book 1 base1 of access Mana to nodeB and 1 base1 of consensus Mana to another. So we need the 2 vectors.

oliviasaa:
Yes, you're right... The thing is, I wrote the specs thinking that both manas (access and consensus) would be always pledged to the same node. 
Ok, so now we have two Base Mana Vectors, one for each (access and consensus) mana

lzpap:
> Yes, you're right... The thing is, I wrote the specs thinking that both manas (access and consensus) would be always pledged to the same node.
> Ok, so now we have two Base Mana Vectors, one for each (access and consensus) mana

@oliviasaa and @acha-bill, just to confirm, so it is like on Figure I, right?
mana_png

oliviasaa:
Yes, if you pledge access and consensus mana to different nodes instead of pledging base mana 1 and 2, it has to be like in figure 1. 

But give me 1 day and I'll try to push this discussion to the research team, because these details were actually never mentioned. Maybe we can simplify something.

acha-bill:
From my understanding,  its as Figure 1.
We said something in the meeting like a node could pledge its access mana to another node to help bootstrap it in the network.
But it wouldn't necessarily want to loose its concensus mana/voting power.

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,469 @@
...
+    n := t - bm.LastUpdated
+    // first, update EBM1, BM2 and EBM2 until `t`
+    bm.updateEBM1(n)
+    bm.updateBM2(n)
+    bm.updateEBM2(n)
+
+    bm.LastUpdated = t
+
+    bm.BaseMana1 += amount
+    bm.BaseMana2 += amount
```
oliviasaa, line 250: The base Mana 2 does not get updated by the whole amount, it uses the "pending mana", it's a formula that depends also on how long the funds sat on an address. So it would look like this:

for input := range tx.inputs {
         // search for the timestamp of the UTXO that generated the input
         t_inp :=  input.timestamp (??? This is probably wrong, but you know what I mean...)
         bm.BaseMana2 += scale/decay*input.balance*(1-math.Pow( math.E,-decay*(tx.timestamp-t_inp) ) ) 
     }

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,469 @@
...
+    amount := sumBalances(tx.outputs)
+
+    for input := range tx.inputs {
+        // search for the nodeID that the input's tx pledged its mana to
+        inputNodeID := loadPledgedNodeIDFromInput(input)
+        bmv[inputNodeID].revokeBaseMana1(input.balance, t)
+        Events.ManaRevoked.Trigger(&ManaRevokedEvent{inputNodeID, input.balance, bmv[inputNodeID], AccessManaType})
+    }
+
+    bmv[pledgedNodeID].pledgeAndUpdate(amount, time)
```
oliviasaa, line 217: Shouldn't this be    bmv[pledgedNodeID].pledgeAndUpdate(amount, t) ?

oliviasaa: I mean, t instead of time

oliviasaa:
Just a general comment:
the event that triggers the Access Mana booking is not that same as the one that triggers the Consensus Mana booking, so the two Base Mana Vectors should be updated independently. 
Then, the Mana used for the dRNG (that is a form of consensus mana) is completely independent of these two Base Mana Vectors. The node that gets dRNG consensus mana is the same that gets regular consensus mana. It will be something that will be used only once per day, so we probably don't even need to store it, depending on the frequency of the snapshots. The dRNG only needs to get this value (initiating the algorithm directly from the snapshot file, maybe?), without updating the Base Mana Vectors, since it will probably be calculated somewhere in the past of both Base Mana Vectors (meaning that a transaction is qualified to be taken into account for dRNG Mana even later than to regular Consensus Mana).

acha-bill:
@oliviasaa 
> the event that triggers the Access Mana booking is not that same as the one that triggers the Consensus Mana booking, so the two Base Mana Vectors should be updated independently.

Can you throw more light on the different triggers?

I understood that mana is booked when a transaction is confirmed. Access or Consensus mana.
So basically
```
when tx confirmed
    book access mana - updates access mana vector
    book consensus mana - updates consensus man vector
```

docs/001-mana_proposal.md
lzpap: Good point! I didn't take into account the time varying nature of pending mana. In your comment, you assume that `alpha` = 1, right? Then if I plot the graph with the `gamma` value from the specs, I get this function:

image

So it's far to say, that pending mana with these parameters reaches its upper bound in appr. 3000 seconds, which is 50 minutes?

lzpap:
> Just a general comment:
> the event that triggers the Access Mana booking is not that same as the one that triggers the Consensus Mana booking, so the two Base Mana Vectors should be updated independently.
> Then, the Mana used for the dRNG (that is a form of consensus mana) is completely independent of these two Base Mana Vectors. The node that gets dRNG consensus mana is the same that gets regular consensus mana. It will be something that will be used only once per day, so we probably don't even need to store it, depending on the frequency of the snapshots. The dRNG only needs to get this value (initiating the algorithm directly from the snapshot file, maybe?), without updating the Base Mana Vectors, since it will probably be calculated somewhere in the past of both Base Mana Vectors (meaning that a transaction is qualified to be taken into account for dRNG Mana even later than to regular Consensus Mana).

Of course, the two Base Mana Vectors are updated independently, but what do you mean by different triggers? As @acha-bill said, isn't the trigger a transaction confirmation?

We should discuss more in detail the DRNG mana thingy, as that is not included in this proposal yet.

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,536 @@
...
+		for input := range tx.inputs {
+			// search for the timestamp of the UTXO that generated the input
+			t_inp := LoadTxTimestampFromOutputID(input)
+			bm2Add := 1/decay * input.balance *(1-math.Pow( math.E,-decay*(t-t_inp) ) ) * math.Pow(math.E, -decay*n)
+			bm.BaseMana2 += bm2Add
+			bm2Pledged += bm2Add
+		}
+		// update EBM1 and EBM2 to `bm.LastUpdated`
+		bm.EffectiveBaseMana1 += amount*(1-math.Pow(math.E,-EMA_coeff*n))
+		bm.EffectiveBaseMana2 += (bm.BaseMana2-oldMana2) * (**additional term**)
```
oliviasaa, line 316: additional term = EMA_coeff*(math.Pow(math.E,-decay*n)-math.Pow(math.E,-EMA_coeff*n))/(EMA_coeff-decay)

lzpap: thanks, I will rename them.

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,538 @@
...
+		bm.EffectiveBaseMana1 += amount*(1-math.Pow(math.E,-EMA_coeff_1*n))
+		bm.EffectiveBaseMana2 += (bm.BaseMana2-oldMana2)*EMA_coeff_2*(math.Pow(math.E,-decay*n)-math.Pow(math.E,-EMA_coeff_2*n))/(EMA_coeff_2-decay)
+	}
+	return bm1Pledged, bm2Pledged
+}
+```
+
+```go
+func (bm *BaseMana) updateEBM1(n time.Duration) {
+    bm.EffectiveBaseMana1 = math.Pow(1-EMA_coeff_1, n) * bm.EffectiveBaseMana1 +
```
oliviasaa, line 324: There is an inconsistency between the formulas (my fault, not yours). This should be: 

bm.EffectiveBaseMana1 = EMA_coeff_1 * math.Pow(math.E, -EMA_coeff_1 * n) * bm.EffectiveBaseMana1 +  (1-math.Pow(math.E, -EMA_coeff_1 * n)) * bm.BaseMana1

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,538 @@
...
+                                 (1-math.Pow(1-EMA_coeff_1, n)) * bm.BaseMana1
+}
+```
+```go
+func (bm *BaseMana) updateBM2(n time.Duration) {
+    bm.BaseMana2 = bm.BaseMana2 * math.Pow(math.E, -decay*n)
+}
+```
+```go
+func (bm *BaseMana) updateEBM2(n time.Duration) {
```
oliviasaa, line 334: There is an inconsistency between the formulas (my fault, not yours). This should be:

bm.EffectiveBaseMana2 = EMA_coeff_2  * math.Pow(math.E, -EMA_coeff_2 * n) * bm.EffectiveBaseMana2 +
                                 ( math.Pow(math.E, -decay * n)- math.Pow(math.E, -EMA_coeff_2 * n)) /
                                 (EMA_coeff_2 - decay) * EMA_coeff_2 * bm.BaseMana2

docs/001-mana_proposal.md
lzpap: updated now.

docs/001-mana_proposal.md
lzpap: updated now.

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,538 @@
...
+```
+
+- `ManaRevoked` when mana (`BM1`) was revoked from a node.
+
+```go
+type ManaRevokedEvent struct {
+    NodeID []bytes
+    AmountBM1 int
+    Time time.Time
+    Type ManaType // access or consensus
```
jkrvivian, line 385: Missing curl bracket: `}`

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+![](https://i.imgur.com/LjfCTm9.png)
+
+The reason for having two separate `Base Mana Vectors` is the fact, that `accessMana` and `consensusMana` can be pledged
+to different nodes.
+
+The exact mathematical formulas, and their respective parameters will be determined later.
+
+## Challenges
+
+### Dependency on Value Tangle
```
capossele, line 75: For now that's ok, but this will need to be changed to just Tangle

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+
+When a transaction is `confirmed`, it is a sufficient indicator that it will not be orphaned. However, in current
+GoShimmer implementation, confirmation is not yet a properly defined concept. This issue will be addressed in a separate
+module.
+
+The Mana module assumes, that the value tangle's `TransactionConfirmed` event is the trigger condition to update the
+mana state machine (base mana vectors for access and consensus mana). Once the concept of transaction finality is
+introduced for the value tangle, the trigger conditions for access and consensus mana calculations can be adjusted.
+
+### Value Transaction Layout
```
capossele, line 89:
```suggestion
### Transaction Layout
```

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+The first implementation of mana in GoShimmer will:
+  - not have voted timestamps on value transactions,
+  - lack proper `TransactionConfirmed` mechanism to trigger mana update,
+  - lack integration into rate control/autopeering/fpc/etc.
+  
+## Detailed Design
+
+In this section, detailed GoShimmer implementation design considerations will be outlined about the mana module.
+In short, changes can be classified into 3 categories:
+ 1. Value Tangle (Transaction) related changes,
```
capossele, line 106:
```suggestion
 1. Transaction related changes,
```

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+ 2. Mana module functionality,
+ 3. and related tools/utilities, such as API, visualization, analytics.
+
+### Value Tangle
```
capossele, line 110:
```suggestion
### Transaction
```

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+nodes to increase their mana.
+
+### Mana Package
+
+The functionality of the mana module should be implemented in a `mana` package. Then, a `mana plugin` can use the package
+structs and methods to connect the dots, for example execute `BookMana` when `TransactionConfirmed` event is triggered
+in the value tangle.
+
+`BaseMana` is a struct that holds the different mana values for a given node.
+Note that except for `Base Mana 1` calculation, we need the time the `BaseMana` values were updated, so we store it in the struct:
```
capossele, line 175:
```suggestion
Note that except for `Base Mana 1` calculation, we need the time when `BaseMana` values were updated, so we store it in the struct:
```

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+`BaseManaVector` is a data structure that maps `nodeID`s to `BaseMana`.
+```go
+type BaseManaVector map[nodeID]BaseMana
+```
+
+#### Methods
+`BaseManaVector` should have the following methods:
+ - `BookMana(transaction)`: Book mana of a transaction. Trigger `ManaBooked` event. Note, that this method updates
+   `BaseMana` with respect to time and to new `Base Mana 1` and `Base Mana 2` values.
+ - `GetTotalMana(nodeID) mana`: Return `Effective Base Mana 1` + `Effective Base Mana 2` of a particular node. Note, that
```
capossele, line 195: Why we would need the sum of Mana1 and Mana2?

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+ - `value/allowedManaPledge`: Endpoint that clients can query to determine which nodeIDs are allowed as part of
+   `accessMana` and `consensusMana` fields in a transaction.
+ - `value/sendTransactionByJson`: Add `accessMana`, `consensusMana` and `timestamp` fields to the JSON request.
+
+Add a new `mana` endpoint route:
+ - `/mana`: Return access and consensus mana of the node.
+ - `/mana/all`: Return whole mana map (mana perception of the node).
+ - `/mana/access/nhighest`: Return `n` highest access mana holder `nodeIDs` and their access mana values.
+ - `/mana/consensus/nhighest`: Return `n` highest consensus mana holder `nodeIDs` and their consensus mana values.
+ - `/mana/rank`: Return the top percentile the node belongs to relative to the network. For example, if there are 100 nodes in the
```
capossele, line 486: I'm afraid this naming could create confusion with `ManaRank` used within the autopeering

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+   top 12%.
+
+#### Metrics collection
+To study the mana module, following metrics could be gathered:
+
+ - Amount of consensus and access mana present in the network. (amount varies because of `Base Mana 2`).
+ - Amount of mana each node holds.
+ - Number of (and amount of mana) a node was pledged with mana in the last `t` interval.
+ - Mana development of a particular node over time.
+ - Mana rank development of a node over time.
```
capossele, line 497: Same issue as above

docs/001-mana_proposal.md
```diff
@@ -0,0 +1,539 @@
...
+
+There are two ways to visualize mana in GoShimmer:
+ 1. Node Local Dashboard
+ 2. Grafana Dashboard
+
+While `Local Dashboard` gives flexibility in what and how to visualize, `Grafana Dashboard` is better at storing historic
+data but can only visualize time series. Therefore, both of these ways will be utilized, depending on which suits the best.
+
+`Local Dashboard` visualization:
+ - Pie chart of mana distribution within the network.
```
capossele, line 522: Why a Pie chart? wouldn't be better a plot on x y axis to also compare it to a zipf distribution?

docs/001-mana_proposal.md
lzpap: Right, let's change it like this:
```go
// GetWeightedMana returns the combination of Effective Base Mana 1 & 2, weighted by weight.
// mana = EBM1 * weight + EBM2 * ( 1- weight), where weight is in [0,1].
func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (float64, error) {
	if _, exist := bmv.vector[nodeID]; !exist {
		return 0.0, errors.New("node not present in base mana vector")
	}
	if weight < 0.0 || weight > 1.0 {
		return 0.0, errors.New("invalid weight parameter, outside of [0,1]")
	}
	bmv.Update(nodeID, time.Now())
	baseMana := bmv.vector[nodeID]
	return baseMana.EffectiveBaseMana1*weight + baseMana.EffectiveBaseMana2*(1-weight), nil
}

// GetMana returns the 50 - 50 split combination of Effective Base Mana 1 & 2.
func (bmv *BaseManaVector) GetMana(nodeID identity.ID) (float64, error) {
	return bmv.GetWeightedMana(nodeID, 0.5)
}
```

docs/001-mana_proposal.md
lzpap: Do you mean a histogram? We can do that too.

docs/001-mana_proposal.md
lzpap: let's call it `mana/percentile` then.

capossele (approved):

PR #722, opened 2020-09-08
Persist base mana vectors

acha-bill:
It saves all base mana vectors in storage on shutdown and loads from storage on startup.

### DIFF
--- a/packages/mana/persitablebase.go
+++ b/packages/mana/persitablebase.go
@@ -1,0 +1,183 @@
+package mana
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/iotaledger/hive.go/stringify"
+
+	"github.com/iotaledger/hive.go/identity"
+	"github.com/iotaledger/hive.go/marshalutil"
+	"github.com/iotaledger/hive.go/objectstorage"
+)
+
+type PersistableBaseMana struct {
+	objectstorage.StorableObjectFlags
+	ManaType           Type
+	BaseMana1          float64
+	EffectiveBaseMana1 float64
+	BaseMana2          float64
+	EffectiveBaseMana2 float64
+	LastUpdated        time.Time
+	NodeID             identity.ID
+
+	bytes []byte
+}
+
+// String returns a human readable version of the PersistableBaseMana.
+func (persistableBaseMana *PersistableBaseMana) String() string {
+	return stringify.Struct("PersistableBaseMana",
+		stringify.StructField("ManaType", fmt.Sprint(persistableBaseMana.ManaType)),
+		stringify.StructField("BaseMana1", fmt.Sprint(persistableBaseMana.BaseMana1)),
+		stringify.StructField("EffectiveBaseMana1", fmt.Sprint(persistableBaseMana.EffectiveBaseMana1)),
+		stringify.StructField("BaseMana2", fmt.Sprint(persistableBaseMana.BaseMana2)),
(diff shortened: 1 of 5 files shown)


packages/mana/persitablebase.go
```diff
@@ -0,0 +1,183 @@
...
+		stringify.StructField("BaseMana1", fmt.Sprint(persistableBaseMana.BaseMana1)),
+		stringify.StructField("EffectiveBaseMana1", fmt.Sprint(persistableBaseMana.EffectiveBaseMana1)),
+		stringify.StructField("BaseMana2", fmt.Sprint(persistableBaseMana.BaseMana2)),
+		stringify.StructField("EffectiveBaseMana2", fmt.Sprint(persistableBaseMana.EffectiveBaseMana2)),
+		stringify.StructField("LastUpdated", fmt.Sprint(persistableBaseMana.LastUpdated)),
+		stringify.StructField("NodeID", persistableBaseMana.NodeID.String()),
+	)
+}
+
+var _ objectstorage.StorableObject = &PersistableBaseMana{}
```
lzpap, line 42: what does this line do?

lzpap (commented):
LGTM, although I still think that using the object storage is a bit of an overkill for this. We don't need caching for example. Can you think of how to write tests mana persistence?

packages/mana/persitablebase.go
capossele: It checks, at compile time, that `PersistableBaseMana` satisfy the `objectstorage.StorableObject` interface

acha-bill:
Yes. We dont need caching. But Object storage is the only interface we have to the database in goshimmer. I think.

For testing, we might have integration test that checks this persistence on shutdown and startup.

Shutdown the node and save its base mana vector.
Start it up again. Then use a new API (e.g get mana vector) to get the vector and assert its not empty or is equal to what was available before shutdown.

plugins/mana/plugin.go
```diff
@@ -127,10 +145,32 @@ func configureEvents() {
 
 func run(_ *node.Plugin) {
 	if err := daemon.BackgroundWorker("Mana", func(shutdownSignal <-chan struct{}) {
-		// TODO: Read base mana vectors from storage
+		readStoredManaVectors()
 		<-shutdownSignal
-		// TODO: write base mana vectors to object storage
+		storeManaVectors()
```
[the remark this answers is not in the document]
acha-bill: - yeah. We should shutdown stores.
- yeah. We should prune old values

plugins/mana/plugin.go
```diff
@@ -127,10 +144,46 @@ func configureEvents() {
...
+func storeManaVectors() {
+	for vectorType, baseManaVector := range baseManaVectors {
+		persitables := baseManaVector.ToPersistables()
+		for _, p := range persitables {
+			storages[vectorType].Store(p).Release()
+		}
+	}
+}
+
+func pruneStroages() {
```
lzpap, line 179:
```suggestion
func pruneStorages() {
```

plugins/mana/plugin.go
```diff
@@ -127,10 +144,46 @@ func configureEvents() {
 
 func run(_ *node.Plugin) {
 	if err := daemon.BackgroundWorker("Mana", func(shutdownSignal <-chan struct{}) {
-		// TODO: Read base mana vectors from storage
+		readStoredManaVectors()
+		pruneStroages()
```
lzpap, line 148:
```suggestion
		pruneStorages()
```

lzpap (approved):

PR #730, opened 2020-09-09
Added mana plugin methods

acha-bill:
Added more methods for mana plugin

- GetHighestManaNodes
- GetManaMap
- GetAccessMana
- GetConsensusMana
- GetNeighborsMana
- GetAllManaVectors
- GetWeightedRandomNodes
- OverrideMana

### DIFF
--- a/packages/mana/basevector.go
+++ b/packages/mana/basevector.go
@@ -146,7 +146,7 @@ func (bmv *BaseManaVector) ForEach(callback func(ID identity.ID, bm *BaseMana) b
 }
 
 //GetManaMap return mana perception of the node.
-func (bmv *BaseManaVector) GetManaMap() Map {
+func (bmv *BaseManaVector) GetManaMap() NodeMap {
 	res := make(map[identity.ID]float64)
 	for ID := range bmv.vector {
 		mana, _ := bmv.GetMana(ID)
@@ -157,11 +157,11 @@ func (bmv *BaseManaVector) GetManaMap() Map {
 
 // GetHighestManaNodes return the n highest mana nodes in descending order.
 // It also updates the mana values for each node.
-func (bmv *BaseManaVector) GetHighestManaNodes(n int) []NodeIDMana {
-	var res []NodeIDMana
+func (bmv *BaseManaVector) GetHighestManaNodes(n uint) []Node {
+	var res []Node
 	for ID := range bmv.vector {
 		mana, _ := bmv.GetMana(ID)
-		res = append(res, NodeIDMana{
+		res = append(res, Node{
 			ID:   ID,
 			Mana: mana,
 		})
@@ -171,7 +171,7 @@ func (bmv *BaseManaVector) GetHighestManaNodes(n int) []NodeIDMana {
 		return res[i].Mana > res[j].Mana
 	})
 
-	if n <= len(res) {
+	if int(n) <= len(res) {
 		return res[:n]
 	}
 	return res[:]
@@ -180,14 +180,13 @@ func (bmv *BaseManaVector) GetHighestManaNodes(n int) []NodeIDMana {
 // SetMana sets the base mana for a node
 func (bmv *BaseManaVector) SetMana(nodeID identity.ID, bm *BaseMana) {
 	bmv.vector[nodeID] = bm
-	_ = bmv.Update(nodeID, time.Now())
 }
 
-// NodeIDMana represents a node and its mana value.
-type NodeIDMana struct {
+// Node represents a node and its mana value.
+type Node struct {
 	ID   identity.ID
 	Mana float64
 }
 
-// Map is a map of nodeID and mana value.
-type Map map[identity.ID]float64
+// NodeMap is a map of nodeID and mana value.
+type NodeMap map[identity.ID]float64
--- a/packages/mana/basevector.go
+++ b/packages/mana/basevector.go
@@ -2,6 +2,8 @@ package mana
 
 import (
 	"errors"
+	"fmt"
+	"sort"
 	"time"
 
 	"github.com/iotaledger/hive.go/identity"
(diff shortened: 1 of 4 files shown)


packages/mana/basevector.go
```diff
@@ -133,3 +135,59 @@ func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (
...
+	for ID := range bmv.vector {
+		mana, _ := bmv.GetMana(ID)
+		res[ID] = mana
+	}
+	return res
+}
+
+// GetHighestManaNodes return the n highest mana nodes in descending order.
+// It also updates the mana values for each node.
+func (bmv *BaseManaVector) GetHighestManaNodes(n int) []NodeIDMana {
```
lzpap, line 160: either make `n` a `uint` or throw an error if it is negative

packages/mana/basevector.go
```diff
@@ -133,3 +135,59 @@ func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (
...
+	if n <= len(res) {
+		return res[:n]
+	}
+	return res[:]
+}
+
+// SetMana sets the base mana for a node
+func (bmv *BaseManaVector) SetMana(nodeID identity.ID, bm *BaseMana) {
+	bmv.vector[nodeID] = bm
+	_ = bmv.Update(nodeID, time.Now())
```
lzpap, line 183: then mention in the method description that `LastUpdated` field of `bm` determines the exact time when values are changed. Maybe this doesn't even have to update, as any other method that accesses mana will update it.

packages/mana/basevector.go
```diff
@@ -133,3 +135,59 @@ func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (
...
+}
+
+// NodeIDMana represents a node and its mana value.
+type NodeIDMana struct {
```
lzpap, line 187: let's just call it `Node`. It has an ID and a Mana value associated with it.

packages/mana/weightedrand.go
```diff
@@ -0,0 +1,56 @@
...
+// Utilizes global rand as the source of randomness -- you will likely want to seed it.
+func (chs RandChooser) Pick(n int) []interface{} {
+	rands := rand.Perm(chs.max)
+	var res []interface{}
+	for _, r := range rands {
+		r = r + 1
+		i := sort.SearchInts(chs.totals, r)
+		res = append(res, chs.data[i].Item)
+	}
+	if n <= len(res) {
```
lzpap, line 52: `n` negative?

plugins/mana/plugin.go
```diff
@@ -189,3 +179,77 @@ func pruneStorages() {
...
+
+// GetHighestManaNodes returns the n highest type mana nodes in descending order.
+// It also updates the mana values for each node.
+func GetHighestManaNodes(manaType mana.Type, n int) []mana.NodeIDMana {
+	bmv := baseManaVectors[manaType]
+	return bmv.GetHighestManaNodes(n)
+}
+
+// GetManaMap return type mana perception of the node.
+func GetManaMap(manaType mana.Type) mana.Map {
```
lzpap, line 191: it looks good from the plugin

plugins/mana/plugin.go
```diff
@@ -189,3 +179,77 @@ func pruneStorages() {
...
+
+//GetWeightedRandomNodes returns a weighted random selection of n nodes.
+func GetWeightedRandomNodes(n int, manaType mana.Type) mana.Map {
+	rand.Seed(time.Now().UTC().UnixNano())
+	manaMap := GetManaMap(manaType)
+	var choices []mana.RandChoice
+	for nodeID, manaValue := range manaMap {
+		choices = append(choices, mana.RandChoice{
+			Item:   nodeID,
+			Weight: int(manaValue * 1000000), //scale float mana to int
```
lzpap, line 242: in the current implementation, transferring 1 IOTA results in getting 1 mana. Therefore, I don't think we need to multiply by that much.
I was thinking about just ignoring any node that has less, than let's say 0.001 mana, since we don't want our mana map to be bigger than necessary. (and due to the asymptotic decay of mana 2, it will never be exactly zero in theory..)
Also, if 1 IOTA is the smallest amount, 1 mana should represent quite an insignificant weight in the network.

lzpap (commented):
yaay, we are getting there! :cat: 
just needs some small adjustments, see below

packages/mana/basevector.go
```diff
@@ -133,3 +135,59 @@ func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (
...
+	ID   identity.ID
+	Mana float64
+}
+
+// Map is a map of nodeID and mana value.
+type Map map[identity.ID]float64
```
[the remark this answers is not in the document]
acha-bill: Na. Lets leave `types.go` for mana types.

packages/mana/basevector.go
acha-bill: You are right. This doesn't have to call update.

plugins/mana/plugin.go
acha-bill: ok. I'll use 1000 for now. We can change this as we see fit.

packages/mana/basevector.go
```diff
@@ -133,3 +135,59 @@ func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (
...
+func (bmv *BaseManaVector) ForEach(callback func(ID identity.ID, bm *BaseMana) bool) {
+	for nodeID, baseMana := range bmv.vector {
+		if !callback(nodeID, baseMana) {
+			return
+		}
+	}
+}
+
+//GetManaMap return mana perception of the node.
+func (bmv *BaseManaVector) GetManaMap() Map {
```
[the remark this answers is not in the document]
acha-bill: We should be consistent with using `map[identity.ID]float64` or `NodeMap`. 
I prefer `NodeMap`

packages/mana/basevector.go
acha-bill: Going with `uint`

packages/mana/basevector.go
acha-bill: yeah

packages/mana/weightedrand.go
acha-bill: Fixed with `uint`

PR #735, opened 2020-09-10
(mana) fix value integration test

acha-bill:
Since we move all faucets funds from the faucets 0 address to the next on startup, we have to use the new address to pull funds in `SendTransactionFromFaucet`


lzpap (approved):
lgtm

PR #742, opened 2020-09-18
Timestamps specification proposal

jonastheis:
[Rendered](https://github.com/iotaledger/goshimmer/blob/feat/timestamps-proposal/docs/timestamps.md)

### DIFF
--- a/docs/timestamps.md
+++ b/docs/timestamps.md
@@ -1,0 +1,215 @@
+# Timestamps
+
+## Motivation
+In order to enable snapshotting based on time constraints rather than special messages in the Tangle (e.g., milestones), nodes need to share the same perception of time. Specifically, they need to have consensus on the *age of messages*. Therefore, messages contain a field `timestamp` which represents the creation time of the message and is signed by the issuing node. 
+
+Having consensus on the creation time of messages enables not only partial ordering but also new applications that require certain guarantees regarding time. 
+
+In this document we propose a mechanism achieving consensus on message timestamps by combining a synchronous and an asynchronous approach. While online nodes can leverage FPC to vote on timestamps, nodes that join the network at a later time use an approach similar to On Tangle Voting to determine valid timestamps. 
+
+## Requirements
+1. Nodes participating in the network need to share the same perception of time.
+2. Consensus on timestamps.
+
+## Dependencies
++ Local markers: they are used to optimize the calculation required to check whether a given message future cone is approved by a given mana threshold.
++ FPC: it's used to perform voting on timestamps.
+
+## Parameters
+- `D` gratuitous network delay ~5 minutes. We assume all messages are delivered within this time.
+- `w` window ~30 minutes. Require w>2D
+- `Delta` max difference in consecutive timestamps. Require Delta>w+D
+- `tw` transaction timestamp window. Max difference between message and transaction timestamp.
+
+
+## Clock synchronization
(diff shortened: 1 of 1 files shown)


docs/timestamps.md
```diff
@@ -0,0 +1,215 @@
+# Timestamps
+
+## Motivation
+In order to enable snapshotting based on time constraints rather than special messages in the Tangle (e.g., milestones), nodes need to share the same perception of time. Specifically, they need to have consensus on the *age of messages*. Therefore, messages contain a field `timestamp` which represents the creation time of the message and is signed by the issuing node.
```
vekkiokonio, line 4: Probably better not to mention milestones (say checkpoints?)
Another application for timestamps is for Adaptive PoW

docs/timestamps.md
```diff
@@ -0,0 +1,215 @@
...
+    if message.timestamp <= parent2.timestamp {
+        return false
+    }
+    
+    return true
+}
+```
+
+### Message timestamp vs transaction timestamp
+Transactions contain a timestamp that is signed by the user when creating the transaction. It is thus different from the timestamp in the message which is created and signed by the node. We require
```
vekkiokonio, line 78: Does this imply that PoW computation should be smaller than `tw`?

vekkiokonio: What's the order of magnitude of `tw`?

vekkiokonio (commented):
Comments on Timestamps specs

vekkiokonio (commented):
Additional comment on Timestamps (Consensus section)

docs/timestamps.md
```diff
@@ -0,0 +1,215 @@
...
+Inputs to a transaction (unspent outputs) inherit their spent time from the transaction timestamp. Similarly, unspent outputs inherit their creation time from the transaction timestamp as well. For a transaction to be considered valid we require
+```
+inputs.timestamp < transaction.timestamp
+```
+In other words, all inputs to a transaction need to have a smaller timestamp than the transaction. In turn, all created unspent outputs will have a greater timestamp than all inputs.
+
+## Consensus
+The timestamp should define the time when the message was created and issued to the Tangle, and this must be enforced to some degree through voting. Specifically, nodes will vote on whether the timestamp was issued within `w` of current local time. This time window is large to account for the network delay. 
+Clearly, in order to have a correct perception of the timestamp quality, **we assume the node is in sync** (see section [Not in Sync](#Not_in_Sync) otherwise).
+Voting on timestamps should not occur for every messsage. Specifically, only for those that arrive around the border of the threshold +-`w`.
```
Thoralf-M, line 99: What protects against an attacker sending many messages that will arrive in the time around the threshold +-`w` threshold? Couldn't that lead to a problem, because then quite a lot of voting is needed?

jonastheis: That is correct. However, an attacker could also force a lot of voting by spamming double spends. So this does not really open a new attack vector. 

We're currently restructuring a lot and need to update all the specs with the new details. In #880 you can see a new data flow (still subject to changes at the moment). Generally, everything that happens *after* the scheduler is governed by the congestion control algorithm. As such, an attacker can not force unlimited voting (be it through conflicts or timestamps) but is limited by its fair share

docs/timestamps.md
```diff
@@ -0,0 +1,215 @@
...
+
+Lets consider now a new message with timestamp 11:30:10. Since |11:30:10-12:00:00| < 30 minutes we will set the opinion to `LIKE`. However, since ||11:30:10-12:00:00| - 30 minutes | is lower than 5 minutes, we will set the level of knowledge for this opinion to 1, meaning that this message timestamp will be object of voting. 
+
+In general, timestamps with level of knowledge 1 will be input of FPC, that will eventually trigger the `finalized` event, after which we can set a message as eligible (or discard, depending on the outcome). If instead, the timestamp we are considering, has already level of knowledge >= 2, we do not need to vote. Either it is eligible (marked as liked) or marked as disliked.
```
apenzk, line 120:
```suggestion
In general, timestamps with level-of-knowledge 1 will be input into FPC, that will eventually trigger the `finalized` event, after which we can set a message as eligible (or discard, depending on the outcome). If instead, the timestamp we are considering, has already level of knowledge >= 2, we do not need to vote, but we will reply to queries. Either it is eligible (marked as liked) or marked as disliked. If the timestamp has level-of-knowledge 3 we do not reply to FPC queries.
```

jonastheis:
This will be included in #825.

PR #743, opened 2020-09-21
Mana Metrics & Wallet Integration

lzpap:
# Description of change

This PR adds several mana related updates:
 - Metrics plugin enriched with mana metrics.
 - Prometheus plugin exports mana related metrics.
 - Grafana `local_dashboard` updated with mana charts and info.
 - `transaction.New()` accepts `pledge` nodeIDs. In case of 1 nodeID supplied, both `Access` and `Consensus` mana is pledged to one node. In case of 2 supplied `pledge` nodeIDs, the first defines `Access` and the second `Consensus` mana pledge. Otherwise, transaction pledges both manas to `emptyNodeID`.
- `transaction.New()` sets the timestamp of the transaction to the time of invocation.
- Faucet pledges mana to the requester node's ID.
- New webapi endpoint `value/allowedManaPledge`, that returns the list of nodeIDs for both `Access` and `Consensus` allowed pledge IDs. The list of allowed pledge IDs can be controlled via config file, by default both contain the node's own ID.
- Added `GetAllowedManaPledgeNodeIDs()` method to client API.
- Added `GetAllowedPledgeIDs` to client wallet library.
- Added `pledge-id` command to `cli-wallet` to query allowed pledge IDs.
- Refactored `cli-wallet` to query node about pledge IDs before signing a new transaction, and choose the first in both lists by default.

NOTE: transaction layout and methods will be refactored once value tangle merge is complete. Until then, above changes are required to test mana functionality.

Screenshots with 3 nodes having mana in the network:
Screenshot from 2020-09-21 18-15-03
Screenshot from 2020-09-21 18-15-30


client/value.go
```diff
@@ -80,3 +84,25 @@ func (api *GoShimmerAPI) SendTransactionByJSON(txn webapi_sendtxnbyjson.Request)
...
+	if err := api.do(http.MethodGet, routeAllowedPledgeNodeIDs, nil, res); err != nil {
+		return nil, err
+	}
+
+	if len(res.Access.Allowed) == 0 {
+		return nil, errors.New("empty list of allowed access mana pledge IDs returned")
+	}
+	if len(res.Consensus.Allowed) == 0 {
+		return nil, errors.New("empty list of allowed consensus mana pledge IDs returned")
+	}
```
acha-bill, line 100: Why is it an error if no pledgeNodeIDs are allowed?

dapps/valuetransfers/packages/transaction/transaction.go
```diff
@@ -61,12 +61,36 @@ type Transaction struct {
...
+		}
+	default:
+		tx = &Transaction{
+			inputs:     inputs,
+			outputs:    outputs,
+			signatures: NewSignatures(),
+			timestamp:  time.Now(),
+		}
 	}
+	return tx
```
acha-bill, line 93: Can we refactor this to use fewer lines? 
Maybe start with a default tx and then add more properties according to the switch.

packages/mana/types.go
```diff
@@ -9,3 +9,14 @@ const (
 	// ConsensusMana is mana associated with consensus weights in the network.
 	ConsensusMana
 )
+
+func (t Type) String() string {
```
acha-bill, line 13: Comments?

tools/cli-wallet/pledge.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/iotaledger/goshimmer/client/wallet"
+)
+
+func execAllowedPledgeNodeIDsCommand(command *flag.FlagSet, cliWallet *wallet.Wallet) {
+	err := command.Parse(os.Args[2:])
+	if err != nil {
+		printUsage(nil, err.Error())
+	}
+
+	fmt.Println()
```
acha-bill, line 17: Empty line?

tools/cli-wallet/pledge.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// query about pledgeNodeIDs
+	res, err := cliWallet.AllowedPledgeNodeIDs()
+	if err != nil {
+		panic(err)
+	}
+
+	for key, value := range res {
+		fmt.Printf("\nAllowed %s full nodeIDs:", key.String())
+		fmt.Println()
```
acha-bill, line 28: You could add `\n` in the `fmt.Printf` instead of adding an empty line with `fmt.Pritln()`

client/value.go
lzpap: You are right, webapi endpoint should return error in this case.

PR #752, opened 2020-09-23
Flatten Mana Webapi & Add mana/online Endpoint

lzpap:
- Mana webapi package structured flattened.
- Add `mana/access/online` and `mana/consensus/online` endpoints to query about mana of online nodes, sorted in descending order.
- Add `Has()` method to `BaseManaVector`.
- Add `GetOnlineNodes()` function to mana plugin. Relies on currently known verified peers taken from the autopeering discovery module.


plugins/webapi/mana/online.go
```diff
@@ -0,0 +1,44 @@
...
+
+func getOnlineConsensusHandler(c echo.Context) error {
+	return getOnlineHandler(c, manaPkg.ConsensusMana)
+}
+
+// getOnlineHandler handles the request.
+func getOnlineHandler(c echo.Context, manaType manaPkg.Type) error {
+	onlinePeersMana, err := manaPlugin.GetOnlineNodes(manaType)
+	if err != nil {
+		return c.JSON(http.StatusNotFound, GetOnlineResponse{Error: err.Error()})
```
acha-bill, line 23: Don't shoot me LOL. :)

The error that is returned here is from `bmv.GetMana` which returns error on 2 cases. 
If the nodeID is not found in the map or if an invalid weight param is used.

Therefore, I don't think returning `http.StatusNotFound` for either of these cases is appropriate.
A more general `http.StatusBadRequest` could be used.

acha-bill (commented):
:)

plugins/webapi/mana/online.go
lzpap: My thinking was that there is nothing wrong with the request, it is the requested resource that has problems in both cases (not in map or invalid weight parameter). It's not the client's fault, rather something in the server, hence this status code.

Btw because of how `GetOnlineNodes` is implemented, we won't see `node not found` errors (mana value will just be `0`), and the only way to get an `invalid weight parameter` is if that hard coded `0.5` is changed in the source code for something invalid.

acha-bill: I see your point.

acha-bill (approved):
🏖

PR #756, opened 2020-09-29
Refactor of the ledgerstate / Feat: Added Colors and ColoredBalances

hmoog:
# Description of change

This PR adds types related to Colors and their corresponding ColoredBalances to the ledgerstate.

## Type of change

## Change checklist

### DIFF
--- a/packages/ledgerstate/color.go
+++ b/packages/ledgerstate/color.go
@@ -1,0 +1,228 @@
+package ledgerstate
+
+import (
+	"bytes"
+	"sort"
+
+	"github.com/iotaledger/hive.go/datastructure/orderedmap"
+	"github.com/iotaledger/hive.go/marshalutil"
+	"github.com/iotaledger/hive.go/stringify"
+	"github.com/mr-tron/base58"
+	"golang.org/x/xerrors"
+)
+
+// region Color ////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// ColorLength represents the length of a Color (amount of bytes).
+const ColorLength = 32
+
+// Color represents a marker that is associated to a token balance and that can give tokens a certain "meaning".
+type Color [ColorLength]byte
+
+// ColorFromBytes unmarshals a Color from a sequence of bytes.
+func ColorFromBytes(bytes []byte) (color Color, consumedBytes int, err error) {
+	marshalUtil := marshalutil.New(bytes)
+	if color, err = ColorFromMarshalUtil(marshalUtil); err != nil {
+		err = xerrors.Errorf("failed to parse Color from MarshalUtil: %w", err)
+		return
+	}
+	consumedBytes = marshalUtil.ReadOffset()
+
+	return
+}
+
+// ColorFromBase58EncodedString creates a Color from a base58 encoded string.
+func ColorFromBase58EncodedString(base58String string) (color Color, err error) {
+	bytes, err := base58.Decode(base58String)
+	if err != nil {
+		err = xerrors.Errorf("error while decoding base58 encoded Color (%v): %w", err, ErrBase58DecodeFailed)
+		return
+	}
+
+	if color, _, err = ColorFromBytes(bytes); err != nil {
+		err = xerrors.Errorf("failed to parse Color from bytes: %w", err)
+		return
+	}
+
+	return
+}
+
+// ColorFromMarshalUtil unmarshals a Color using a MarshalUtil (for easier unmarshaling).
+func ColorFromMarshalUtil(marshalUtil *marshalutil.MarshalUtil) (color Color, err error) {
+	colorBytes, err := marshalUtil.ReadBytes(ColorLength)
+	if err != nil {
+		err = xerrors.Errorf("failed to parse Color (%v): %w", err, ErrParseBytesFailed)
+		return
+	}
+	copy(color[:], colorBytes)
+
+	return
+}
+
+// Bytes marshals the Color into a sequence of bytes.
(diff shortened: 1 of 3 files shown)


packages/ledgerstate/color.go
```diff
@@ -0,0 +1,228 @@
...
+
+// ColorLength represents the length of a Color (amount of bytes).
+const ColorLength = 32
+
+// Color represents a marker that is associated to a token balance and that can give tokens a certain "meaning".
+type Color [ColorLength]byte
+
+// ColorFromBytes unmarshals a Color from a sequence of bytes.
+func ColorFromBytes(bytes []byte) (color Color, consumedBytes int, err error) {
+	marshalUtil := marshalutil.New(bytes)
```
jonastheis, line 24: `bytes` collides with the import `"bytes"`. Not that important but could lead to some issues/confusion.

packages/ledgerstate/color.go
```diff
@@ -0,0 +1,228 @@
...
+		return "IOTA"
+	case ColorMint:
+		return "MINT"
+	default:
+		return c.Base58()
+	}
+}
+
+// ColorIOTA is the zero value of the Color and represents uncolored tokens.
+var ColorIOTA = Color{}
```
jonastheis, line 85: Why are these Colors at the bottom and not constants?
I'd move them to the top and if no particular reason for being `var` make them `const`.

packages/ledgerstate/color.go
```diff
@@ -0,0 +1,228 @@
...
+// NewColoredBalances returns a new deterministically ordered collection of ColoredBalances.
+func NewColoredBalances(balances map[Color]uint64) (coloredBalances *ColoredBalances) {
+	coloredBalances = &ColoredBalances{balances: orderedmap.New()}
+
+	// deterministically sort colors
+	sortedColors := make([]Color, 0, len(balances))
+	for color := range balances {
+		sortedColors = append(sortedColors, color)
+	}
+	sort.Slice(sortedColors, func(i, j int) bool { return bytes.Compare(sortedColors[i][:], sortedColors[j][:]) < 0 })
```
jonastheis, line 109: Maybe it'd be worth to fulfil the `sort.Interface` for `Color` so this could just be a `sort.Sort`.

packages/ledgerstate/color.go
```diff
@@ -0,0 +1,228 @@
...
+	coloredBalances = NewColoredBalances(nil)
+	for i := uint32(0); i < balancesCount; i++ {
+		color, colorErr := ColorFromMarshalUtil(marshalUtil)
+		if colorErr != nil {
+			err = xerrors.Errorf("failed to parse Color from MarshalUtil: %w", colorErr)
+			return
+		}
+
+		// check semantic correctness (ensure ordering)
+		if previousColor != nil && bytes.Compare(previousColor[:], color[:]) != -1 {
```
jonastheis, line 149: If `Color` has a method to compare then this could be simplified to be more readable.

packages/ledgerstate/color.go
```diff
@@ -0,0 +1,228 @@
...
+			err = xerrors.Errorf("parsed Colors are not in correct order: %w", ErrParseBytesFailed)
+			return
+		}
+
+		balance, balanceErr := marshalUtil.ReadUint64()
+		if balanceErr != nil {
+			err = xerrors.Errorf("failed to parse balance (%v): %w", balanceErr, ErrParseBytesFailed)
```
jonastheis, line 156: Make this error more precise.
```suggestion
			err = xerrors.Errorf("failed to parse balance of Color %s (%v): %w", color.String(), balanceErr, ErrParseBytesFailed)
```

lzpap (approved):
LGTM

packages/ledgerstate/color.go
hmoog: We can move them to the top but Colors cannot be constants because constants in Go can only be scalar types like string, integer and so on.

packages/ledgerstate/color.go
hmoog: Our errors already provide stack traces so I prefer to not be "too specific" about errors as this information is already provided by the stack trace and adding the "context" kind of becomes ambigious but I guess its up discussion.

packages/ledgerstate/color_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	clonedColoredBalances, consumedBytes, err := ColoredBalancesFromBytes(marshaledColoredBalances)
+	require.NoError(t, err)
+	assert.Equal(t, len(marshaledColoredBalances), consumedBytes)
+	assert.Equal(t, clonedColoredBalances.Size(), coloredBalances.Size())
+	assert.Equal(t, clonedColoredBalances.Bytes(), coloredBalances.Bytes())
+
+	color0Balance, color0Exists := clonedColoredBalances.Get(Color{0})
+	assert.True(t, color0Exists)
+	assert.Equal(t, color0Balance, uint64(200))
```
jkrvivian, line 47: The expected value should be in front.
```suggestion
	assert.Equal(t, uint64(200), color0Balance)
```

packages/ledgerstate/color.go
hmoog: This would require us to introduce sth like `type Colors []Color` so I think it doesn't really make sense

packages/ledgerstate/color.go
hmoog: agree

jkrvivian (approved):

PR #762, opened 2020-10-02
Refactor of the ledgerstate / Feat: Added Output models for UTXODAG

hmoog:
# Description of change

This PR adds the refactored models for Outputs and their metadata that are being using in the UTXODAG.

## Type of change

## Change checklist

### DIFF
--- a/packages/ledgerstate/output.go
+++ b/packages/ledgerstate/output.go
@@ -42,7 +42,8 @@ func (o OutputType) String() string {
 // OutputIDLength contains the amount of bytes that a marshaled version of the OutputID contains.
 const OutputIDLength = TransactionIDLength + marshalutil.UINT16_SIZE
 
-// OutputID is the data type that represents the identifier of an Output.
+// OutputID is the data type that represents the identifier of an Output (which consists of a TransactionID and the
+// index of the Output in the Transaction that created it).
 type OutputID [OutputIDLength]byte
 
 // NewOutputID is the constructor for the OutputID.
@@ -679,14 +680,6 @@ func (o *OutputMetadata) BranchID() BranchID {
 
 // SetBranchID sets the identifier of the Branch that the Output was booked in.
 func (o *OutputMetadata) SetBranchID(branchID BranchID) (modified bool) {
-	o.branchIDMutex.RLock()
-	if o.branchID == branchID {
-		o.branchIDMutex.RUnlock()
-
-		return
-	}
-
-	o.branchIDMutex.RUnlock()
 	o.branchIDMutex.Lock()
 	defer o.branchIDMutex.Unlock()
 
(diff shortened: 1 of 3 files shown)


packages/ledgerstate/output.go
```diff
@@ -0,0 +1,1031 @@
...
+	preferred               bool
+	preferredMutex          sync.RWMutex
+	liked                   bool
+	likedMutex              sync.RWMutex
+	finalized               bool
+	finalizedMutex          sync.RWMutex
+	confirmed               bool
+	confirmedMutex          sync.RWMutex
+	rejected                bool
+	rejectedMutex           sync.RWMutex
```
jonastheis, line 580: These are mutually exclusive and could be locked with a single mutex. In that way we avoid weird behaviour like an output being `confirmed` and `rejected` at the same time. 

What I'm saying is this: the model could support us by preventing to get into inconsistent states. I.e. there could be some logic in here that helps with developing and debugging.

packages/ledgerstate/output.go
```diff
@@ -0,0 +1,1031 @@
...
+// BranchID returns the identifier of the Branch that the Output was booked in.
+func (o *OutputMetadata) BranchID() BranchID {
+	o.branchIDMutex.RLock()
+	defer o.branchIDMutex.RUnlock()
+
+	return o.branchID
+}
+
+// SetBranchID sets the identifier of the Branch that the Output was booked in.
+func (o *OutputMetadata) SetBranchID(branchID BranchID) (modified bool) {
```
jonastheis, line 681: I'd simplify all `SetX` methods to only acquire the write lock. It becomes less verbose, easier to read and you safe the replicated `if`.

jonastheis (changes requested):

packages/ledgerstate/output.go
hmoog: I would like to keep the models as dumb as possible but would it maybe make sense to introduce an enum for this kind of "final state" (i.e. Pending, Confirmed, Rejected) where it can have exactly one of those values?

`Pending`: `confirmed = false && rejected == false` (in the old model)
`Confirmed`: `confirmed = true && rejected = false` (in the old model)
`Rejected`: `confirmed = false && rejected = true` (in the old model)

Then the invalid state of `confirmed = true && rejected = true` wouldn't even exist

packages/ledgerstate/output.go
hmoog: Done

packages/ledgerstate/output.go
hmoog: According to our discussion in the slack channel, we will need to revisit this once we introduce grades of finality and levels of knowledge anyway and leave it like this for now.

jonastheis (approved):

PR #769, opened 2020-10-05
Remove dgrijalva/jwt-go dependency

capossele:


jonastheis (approved):
When removing this we should probably provide a quick tutorial or at least a link as to how to protect the API (reverse proxy) in the wiki.

Wollac:
How about we just set the [`basic-auth` middleware ](https://echo.labstack.com/middleware/basic-auth) for the API. Could be easier than writing a tutorial :stuck_out_tongue:

PR #774, opened 2020-10-05
Write rand-seed output to a file

capossele:


tools/rand-seed/main.go
```diff
@@ -16,6 +17,18 @@ func main() {
 		return
 	}
 
-	fmt.Printf("base64:%s\n", base64.StdEncoding.EncodeToString(b))
-	fmt.Printf("base58:%s\n", base58.Encode(b))
+	// If the file doesn't exist, create it, or append to the file
+	f, err := os.OpenFile("random-seed.txt", os.O_WRONLY|os.O_CREATE|os.O_WRONLY, 0644)
```
jkrvivian, line 21:
```suggestion
	f, err := os.OpenFile("random-seed.txt", os.O_WRONLY|os.O_CREATE|os.O_APPEND , 0644)
```

jkrvivian (changes requested):

tools/rand-seed/main.go
capossele: Actually here the comment is misleading. We do not want to append but to override what was in it before.

tools/rand-seed/main.go
```diff
@@ -16,6 +17,18 @@ func main() {
...
-	fmt.Printf("base58:%s\n", base58.Encode(b))
+	// If the file doesn't exist, create it, or truncate the file
+	f, err := os.Create("random-seed.txt")
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	defer f.Close()
+
+	_ = f.Truncate(0)
```
jkrvivian, line 28: `Create` truncates the file if it exists, then this line can be removed. :beer:

jkrvivian (approved):

PR #778, opened 2020-10-05
Flatten the Faucet

lzpap:
Flatten the `Faucet` dApp.
 - Remove packages,
 - Rename faucet payload to faucet object.
 - Rename `dapp.go` to `plugin.go`,
 - Move faucet under `plugins`.


plugins/faucet/object.go
```diff
@@ -8,39 +8,89 @@ import (
...
 
-// Payload represents a request which contains an address for the faucet to send funds to.
-type Payload struct {
+// region Object implementation ///////////////////////////////////////////////////////////////////////////////////////
+
+// Type represents the identifier for the faucet Object type.
+var Type = tangle.PayloadType(2)
+
+// Object represents a faucet request which contains an address for the faucet to send funds to.
+type Object struct {
```
capossele, line 29:
```suggestion
type FaucetRequest struct {
```

capossele: We have discussed and decided to use domain specific names for the model in the package (e.g., transaction, facuetRequest). Please change this accordingly 👍

capossele (changes requested):

plugins/faucet/object.go
lzpap: I named it as `Request`, because linter was complaining about it starting with the same name as the package. From outside, it will be `faucet.Request`.

capossele (approved):

Issue #771, opened 2020-10-05
Additional Mana Metrics


lzpap:
Add these new metrics to grafana:
 - Average pledge amount of a node.
 - Average mana of neighbors.

PR #780, opened 2020-10-06
Additional mana  metrics

acha-bill:
- Average pledge amount of a node.
- Average mana of neighbors.

#771

### DIFF
--- a/plugins/metrics/mana.go
+++ b/plugins/metrics/mana.go
@@ -6,16 +6,26 @@ import (
 	"github.com/iotaledger/goshimmer/packages/mana"
 	"github.com/iotaledger/goshimmer/plugins/autopeering/local"
 	manaPlugin "github.com/iotaledger/goshimmer/plugins/mana"
+	"github.com/iotaledger/hive.go/identity"
 	"go.uber.org/atomic"
 )
 
+// NodePledgeMap is a map of node and a list of mana pledges.
+type NodePledgeMap map[identity.ID][]float64
+
 var (
-	accessMap           mana.NodeMap
-	accessPercentile    atomic.Float64
-	accessLock          sync.RWMutex
-	consensusMap        mana.NodeMap
-	consensusPercentile atomic.Float64
-	consensusLock       sync.RWMutex
+	accessMap                 mana.NodeMap
+	accessPercentile          atomic.Float64
+	accessLock                sync.RWMutex
+	consensusMap              mana.NodeMap
+	consensusPercentile       atomic.Float64
+	consensusLock             sync.RWMutex
+	accessPledgeLock          sync.RWMutex
+	accessPledge              = NodePledgeMap{}
+	consensusPledgeLock       sync.RWMutex
+	consensusPledge           = NodePledgeMap{}
+	averageNeighborsAccess    atomic.Float64
+	averageNeighborsConsensus atomic.Float64
 )
 
 // AccessPercentile returns the top percentile the node belongs to in terms of access mana holders.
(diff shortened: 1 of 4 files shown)


plugins/metrics/mana.go
```diff
@@ -64,6 +74,65 @@ func ConsensusManaMap() mana.NodeMap {
...
+			sum += v
+		}
+		result[k] = sum / float64(len(pledges))
+	}
+	return result
+}
+
+// addPledge populates the pledge logs.
+func addPledge(event *mana.PledgedEvent) {
+	amount := (event.AmountBM1 + event.AmountBM2) / 2
```
lzpap, line 119: This `/2` should actually be a function of the default weight we use for mana calculation. It is 0.5 atm, but what if we change it after testing?
Alternatively, we could say that we need this metric to visualize average base mana pledge amounts, then we decouple from mana calculation and weight. We end up with a graph that shows how much base mana was pledged to a node on average (BM1 and BM2 separately in the same graph). Because of mana calculation, it doesn't sound right to equate pledge amounts to calculated mana values.

lzpap (commented):
Looks good!
Couple remarks for the grafana dashboard:
 - Add detailed descriptions to the panels you introduce.
 - `Average Mana of Neighbors` panel can display both access and consensus, see how panels on the left do it (multiple data series in panel).
 -  Add the new dashboard file to `tools/docker-network/grafana/dashboards` as well.

plugins/metrics/mana.go
```diff
@@ -6,16 +6,26 @@ import (
 	"github.com/iotaledger/goshimmer/packages/mana"
 	"github.com/iotaledger/goshimmer/plugins/autopeering/local"
 	manaPlugin "github.com/iotaledger/goshimmer/plugins/mana"
+	"github.com/iotaledger/hive.go/identity"
 	"go.uber.org/atomic"
 )
 
+// NodePledgeMap is a map of node and a list of mana pledges.
+type NodePledgeMap map[identity.ID][]float64
```
lzpap, line 14: this struct is unused now, isn't it?

plugins/metrics/mana.go
```diff
@@ -64,6 +74,73 @@ func ConsensusManaMap() mana.NodeMap {
...
+	var sum float64
+	for _, pledge := range pledges {
+		sum += pledge
+	}
+	return sum / float64(len(pledges))
+}
+
+// addPledge populates the pledge logs for the node.
+func addPledge(event *mana.PledgedEvent) {
+	if event.NodeID != local.GetInstance().ID() {
```
lzpap, line 119: if we do it like that, we could log all pledge events, not just our own.

plugins/prometheus/mana.go
```diff
@@ -6,10 +6,16 @@ import (
...
+	accessManaMap             *prometheus.GaugeVec
+	accessPercentile          prometheus.Gauge
+	consensusManaMap          *prometheus.GaugeVec
+	consensusPercentile       prometheus.Gauge
+	averageNeighborsAccess    prometheus.Gauge
+	averageNeighborsConsensus prometheus.Gauge
+	averageAccessPledgeBM1    prometheus.Gauge
+	averageAccessPledgeBM2    prometheus.Gauge
+	averageConsensusPledgeBM1 prometheus.Gauge
+	averageConsensusPledgeBM2 prometheus.Gauge
```
lzpap, line 18: these would also have to be `*prometheus.GaugeVec`  (or just make two, and label data with base mana type and nodeID)

plugins/metrics/mana.go
```diff
@@ -6,16 +6,26 @@ import (
...
-	consensusPercentile atomic.Float64
-	consensusLock       sync.RWMutex
+	accessMap                 mana.NodeMap
+	accessPercentile          atomic.Float64
+	accessLock                sync.RWMutex
+	consensusMap              mana.NodeMap
+	consensusPercentile       atomic.Float64
+	consensusLock             sync.RWMutex
+	accessPledgeLock          sync.RWMutex
+	accessPledge              = map[uint][]float64{}
```
lzpap, line 24: currently we only log the pledges of our own node, but have the ability to do so for all nodes in the network.
```go
accessPledge = NodePledgeMap{}

type NodePledgeMap map[identity.id]PledgeLog

type PledgeLog struct {
    BM1Pledges []float64
    BM2Pledges []float64
}
```
and then we could define the methods on these types.
```go
func (p PledgeLog) AddBM1(val float54) {
    p.BM1Pledged = append(p.BM1Pledged, val)
}

func (p PledgeLog) GetBM1Average() float64 {
    for _, val := range p.BM1Pledges {...}
    return avg
}
```
.. and so on. Not only it makes the code more readable, but we measure average pledge amounts of all nodes.

plugins/prometheus/mana.go
```diff
@@ -43,10 +49,62 @@ func registerManaMetrics() {
...
+			"nodeID",
+		})
+	averageAccessPledgeBM2 = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mana_average_access_pledge_bm2",
+			Help: "Average base mana 2 of pledged access mana.",
+		},
+		[]string{
+			"nodeID",
+		})
```
lzpap, line 67: these two could be merged into one by introducing a new label "baseMana" which could either be "1" or "2".

plugins/prometheus/mana.go
```diff
@@ -43,10 +49,62 @@ func registerManaMetrics() {
...
+	averageConsensusPledgeBM2 = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mana_average_consensus_pledge_bm2",
+			Help: "Average base mana 2 of pledged consensus mana.",
+		},
+		[]string{
+			"nodeID",
+		})
```
lzpap, line 84: same for these two

plugins/prometheus/mana.go
```diff
@@ -60,4 +118,22 @@ func collectManaMetrics() {
 		consensusManaMap.WithLabelValues(nodeID.String()).Set(value)
 	}
 	consensusPercentile.Set(metrics.ConsensusPercentile())
+	averageNeighborsAccess.Set(metrics.AverageNeighborsAccess())
+	averageNeighborsConsensus.Set(metrics.AverageNeighborsConsensus())
+
+	accessBM1Pledges, accessBM2Pledges := metrics.AveragePledgeAccessBM()
+	for nodeID, value := range accessBM1Pledges {
+		averageAccessPledgeBM1.WithLabelValues(nodeID.String()).Set(value)
```
lzpap, line 126: and here we could tag with "baseMana" label

PR #782, opened 2020-10-07
Refactor of the ledgerstate / Feat: Added Input models

hmoog:
# Description of change

This PR introduces the models for Transaction Inputs.

## Type of change

## Change checklist

### DIFF
--- a/packages/ledgerstate/input.go
+++ b/packages/ledgerstate/input.go
@@ -154,7 +154,7 @@ func InputsFromMarshalUtil(marshalUtil *marshalutil.MarshalUtil) (inputs Inputs,
 	parsedInputs := make([]Input, inputsCount)
 	for i := uint16(0); i < inputsCount; i++ {
 		if parsedInputs[i], err = InputFromMarshalUtil(marshalUtil); err != nil {
-			err = xerrors.Errorf("failed to parse Input from MarshalUtil: %w", err, ErrParseBytesFailed)
+			err = xerrors.Errorf("failed to parse Input from MarshalUtil: %w", err)
 			return
 		}
 
@@ -173,9 +173,7 @@ func InputsFromMarshalUtil(marshalUtil *marshalutil.MarshalUtil) (inputs Inputs,
 // Clone creates a copy of the Inputs.
 func (i Inputs) Clone() (clonedInputs Inputs) {
 	clonedInputs = make(Inputs, len(i))
-	for i, input := range i {
-		clonedInputs[i] = input
-	}
+	copy(clonedInputs[:], i)
 
 	return
 }
--- a/packages/ledgerstate/input.go
+++ b/packages/ledgerstate/input.go
@@ -45,9 +45,9 @@ type Input interface {
 	// String returns a human readable version of the Input.
 	String() string
 
-	// Compare offers a comparator for Inputs which returns -1 if otherInput is bigger, 1 if it is smaller and 0 if they
+	// Compare offers a comparator for Inputs which returns -1 if other Input is bigger, 1 if it is smaller and 0 if they
 	// are the same.
-	Compare(otherInput Input) int
+	Compare(other Input) int
 }
 
 // InputFromBytes unmarshals an Input from a sequence of bytes.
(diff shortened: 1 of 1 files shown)


packages/ledgerstate/input.go
```diff
@@ -0,0 +1,270 @@
...
+
+// InputsFromMarshalUtil unmarshals a collection of Inputs using a MarshalUtil (for easier unmarshaling).
+func InputsFromMarshalUtil(marshalUtil *marshalutil.MarshalUtil) (inputs Inputs, err error) {
+	inputsCount, err := marshalUtil.ReadUint16()
+	if err != nil {
+		err = xerrors.Errorf("failed to parse inputs count (%v): %w", err, ErrParseBytesFailed)
+		return
+	}
+
+	var previousInput *Input
```
jonastheis, line 153: Is the pointer here really needed?

packages/ledgerstate/input.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/iotaledger/hive.go/stringify"
+	"github.com/iotaledger/hive.go/types"
+	"github.com/iotaledger/hive.go/typeutils"
+	"golang.org/x/xerrors"
+)
+
+// region InputType ////////////////////////////////////////////////////////////////////////////////////////////////////
+
+const (
+	// UTXOInputType is the type of an Input that references a UTXO Output.
```
jonastheis, line 19:
```suggestion
	// UTXOInputType is the type of an Input that references an UTXO Output.
```

packages/ledgerstate/input.go
hmoog: yes, because otherwise we are not able to determine the "unset" value. Of course you could introduce another bool to track that but that would not just be ugly but also slower :P

packages/ledgerstate/input.go
```diff
@@ -0,0 +1,270 @@
...
+	// Type returns the type of the Input.
+	Type() InputType
+
+	// Bytes returns a marshaled version of the Input.
+	Bytes() []byte
+
+	// String returns a human readable version of the Input.
+	String() string
+
+	// Compare offers a comparator for Inputs which returns -1 if otherInput is bigger, 1 if it is smaller and 0 if they
```
jkrvivian, line 48:
```suggestion
	// Compare offers a comparator for Inputs which returns -1 if other Input is bigger, 1 if it is smaller and 0 if they
```

packages/ledgerstate/input.go
```diff
@@ -0,0 +1,270 @@
...
+func (u *UTXOInput) ReferencedOutputID() OutputID {
+	return u.referencedOutputID
+}
+
+// Bytes returns a marshaled version of the Input.
+func (u *UTXOInput) Bytes() []byte {
+	return byteutils.ConcatBytes([]byte{byte(UTXOInputType)}, u.referencedOutputID.Bytes())
+}
+
+// Compare offers a comparator for Inputs which returns -1 if otherInput is bigger, 1 if it is smaller and 0 if they
```
jkrvivian, line 254:
```suggestion
// Compare offers a comparator for Inputs which returns -1 if other Input is bigger, 1 if it is smaller and 0 if they
```

packages/ledgerstate/input.go
jonastheis: I think it should work without the pointer in this case since an interface that does not hold a concrete value is `nil`. We do assign with `previousInput = &parsedInputs[i]` (effectively making `previousInput != nil` always true) but that happens after the initial check and then there should never be a `nil` concrete value be returned from `InputFromMarshalUtil`. See this example: https://play.golang.org/p/CzbZbUhLtXl

hmoog: Oh yeah you are right its an Interface! O_O

packages/ledgerstate/input.go
hmoog: otherInput is the name of the variable

packages/ledgerstate/input.go
hmoog: otherInput is the name of the variable

PR #785, opened 2020-10-07
Add webapi basic-auth

### LINKED ISSUE #783: Add basic-auth
## Description

Add [basic-auth](https://echo.labstack.com/middleware/basic-auth) as a replacement of the previous JWT

capossele:
This PR adds support to http basic-auth for:

- web-api
- client lib
- cli-wallet

Fixes #783

### DIFF
--- a/client/lib.go
+++ b/client/lib.go
@@ -30,18 +30,79 @@ const (
 	contentTypeJSON = "application/json"
 )
 
-// NewGoShimmerAPI returns a new *GoShimmerAPI with the given baseURL and httpClient.
-func NewGoShimmerAPI(baseURL string, httpClient ...http.Client) *GoShimmerAPI {
-	if len(httpClient) > 0 {
-		return &GoShimmerAPI{baseURL: baseURL, httpClient: httpClient[0]}
+// Options define the options of a GoShimmer API client.
+type Options struct {
+	// The initial committee of the DRNG.
+	BasicAuth *BasicAuth
+	// The initial randomness of the DRNG.
+	HTTPClient *http.Client
+}
+
+// Option is a function which sets the given option.
+type Option func(*Options)
+
+// BasicAuth defines the basic-auth struct.
+type BasicAuth struct {
+	enabled  bool
+	username string
+	password string
+}
+
+// NewBasicAuth returns a new BasicAuth.
+func NewBasicAuth(username, password string) *BasicAuth {
+	return &BasicAuth{
+		enabled:  true,
+		username: username,
+		password: password,
 	}
-	return &GoShimmerAPI{baseURL: baseURL}
+}
+
+// Enabled returns the enabled state of a given BasicAuth.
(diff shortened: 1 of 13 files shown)


client/lib.go
```diff
@@ -30,18 +30,82 @@ const (
...
+// NewGoShimmerAPI returns a new *GoShimmerAPI with the given baseURL and options.
+func NewGoShimmerAPI(baseURL string, setters ...Option) *GoShimmerAPI {
+	args := &Options{}
+
+	for _, setter := range setters {
+		setter(args)
+	}
+	if args.HTTPClient == nil {
+		args.HTTPClient = &http.Client{}
+	}
```
jkrvivian, line 96: Don't need this check if we have `newOptions`.
```
func newOptions() *Options {
    return &Options{
         BasicAuth: &BasicAuth{},
         HTTPClient: &http.Client{},
    }
}
```

capossele: @jkrvivian good suggestions! I've refactored the way we handle options, so that now feels more intuitive (I hope)!

jkrvivian: That's even better! :smile:

jkrvivian (approved):

plugins/webapi/plugin.go
```diff
@@ -47,6 +47,17 @@ func Server() *echo.Echo {
...
+			server.Use(middleware.BasicAuth(func(username, password string, c echo.Context) (bool, error) {
+				if username == config.Node().GetString(CfgBasicAuthUsername) &&
+					password == config.Node().GetString(CfgBasicAuthPassword) {
+					return true, nil
+				}
+				return false, nil
+			}))
+		}
 	})
 	return server
```
Wollac, line 62: The error server does not have an `HTTPErrorHandler` set, this should be done to log errors similar to `plugin/dashboard`

Wollac (changes requested):
I also have some remarks about stuff not actually touched in the PR:

Unauthorized access is not logged correctly:
In https://github.com/iotaledger/goshimmer/blob/fd03b22ff7b4ddde48a4907be55455c45bbfb704/plugins/dashboard/routes.go#L77 
this should be replaced with 
```log.Warnf("Request failed: %s", err)```
to log errors in our usual style.

In https://github.com/iotaledger/goshimmer/blob/fd03b22ff7b4ddde48a4907be55455c45bbfb704/plugins/webapi/plugin.go#L87
it could make sense to also log whether basic_auth is on or not.

Same for https://github.com/iotaledger/goshimmer/blob/fd03b22ff7b4ddde48a4907be55455c45bbfb704/plugins/dashboard/plugin.go#L115

jkrvivian (approved):

jkrvivian (approved):

frohal:
A small hint on how to create the password-hash to put in the JSON  could possibly have been added as well ?
@Wollac

PR #795, opened 2020-10-14
Mana Research Dashboard (Grafana)

### LINKED ISSUE #770: Mana Dashboards for Research
Create additional grafana dashboards visualizing mana metrics with taking only EBM1 or EBM2 into account, so that research can analyze the difference between the two. The main  grafana dashboard should still visualize the chosen way of mana calculation (50-50 split).

lzpap:
Fixes #770 

- Refactor `baseManaVector` methods to accept mana calculation weights instead of using hardcoded weight of `0.5`.
- Introduce constants for weights: `OnlyMana1`, `OnlyMana2` and `Mixed` (50-50).
- Collect mana research metrics for all three "modes" of mana calculation. (metrics pkg)
- Export mana research metrics for all three "modes" of mana calculation. (prometheus)
- Compare mana research metrics on `Mana Research Dashboard` in Grafana.
- Mana research metrics collection (and export) is disabled by default, can be enabled via config:

```json
  "metrics": {
    "manaResearch": true,
  },
```
- Mana webapi endpoints return `Mixed` mode values.

image

### DIFF
--- a/packages/mana/basevector.go
+++ b/packages/mana/basevector.go
@@ -173,7 +173,7 @@ func (bmv *BaseManaVector) ForEach(callback func(ID identity.ID, bm *BaseMana) b
 	}
 }
 
-//GetManaMap returns mana perception of the node.
+// GetManaMap returns mana perception of the node.
 func (bmv *BaseManaVector) GetManaMap(weight float64) (NodeMap, error) {
 	bmv.Lock()
 	defer bmv.Unlock()
--- a/packages/mana/basevector.go
+++ b/packages/mana/basevector.go
@@ -153,19 +153,12 @@ func (bmv *BaseManaVector) UpdateAll(t time.Time) error {
 	return nil
 }
 
-// GetWeightedMana returns the combination of Effective Base Mana 1 & 2, weighted by weight.
+// GetMana returns combination of Effective Base Mana 1 & 2, weighted by `weight` parameter.
 // mana = EBM1 * weight + EBM2 * ( 1- weight), where weight is in [0,1].
-func (bmv *BaseManaVector) GetWeightedMana(nodeID identity.ID, weight float64) (float64, error) {
+func (bmv *BaseManaVector) GetMana(nodeID identity.ID, weight float64) (float64, error) {
 	bmv.Lock()
 	defer bmv.Unlock()
 	return bmv.getWeightedMana(nodeID, weight)
-}
-
-// GetMana returns the 50 - 50 split combination of Effective Base Mana 1 & 2.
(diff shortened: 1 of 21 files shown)


packages/mana/basevector.go
```diff
@@ -180,44 +173,54 @@ func (bmv *BaseManaVector) ForEach(callback func(ID identity.ID, bm *BaseMana) b
 	}
 }
 
-//GetManaMap return mana perception of the node.
-func (bmv *BaseManaVector) GetManaMap() NodeMap {
+//GetManaMap returns mana perception of the node.
```
acha-bill, line 176:
```suggestion
// GetManaMap returns mana perception of the node.
```

plugins/metrics/manaresearch.go
```diff
@@ -0,0 +1,148 @@
...
+var (
+	accessMapBM1           mana.NodeMap
+	accessPercentileBM1    atomic.Float64
+	accessLockBM1          sync.RWMutex
+	consensusMapBM1        mana.NodeMap
+	consensusPercentileBM1 atomic.Float64
+	consensusLockBm1       sync.RWMutex
+)
+
+// AccessPercentileBM1 returns the top percentile the node belongs to in terms of access mana holders.
```
acha-bill, line 21: Maybe specify in the comments that it's for access mana "using BM1".
I don't feel strongly about this though. (if you disagree)

plugins/webapi/info/plugin.go
```diff
@@ -5,6 +5,7 @@ import (
 	"sort"
 	goSync "sync"
 
+	manaPkg "github.com/iotaledger/goshimmer/packages/mana"
```
acha-bill, line 8: It is preferable to refer to mana package as `mana` by default.
If we need the plugin, then we can use the less common `manaPlugin`.

tools/docker-network/docker-compose.yml
```diff
@@ -47,6 +47,7 @@ services:
       --valueLayer.snapshot.file=/tmp/assets/7R1itJx5hVuo9w9hjg5cwKFmek4HMSoBDgJZN8hKGxih.bin
       --syncbeacon.broadcastInterval=5
       --syncbeacon.startSynced=true
+      --metrics.manaResearch=true
```
acha-bill, line 50: Shouldn't this be `false` by default?

acha-bill (commented):
Nice :)

tools/docker-network/docker-compose.yml
lzpap: It is `false` by default, and only enabled for the faucet node (peer_master) in the docker network, because this is the node grafana (and prometheus) queries for data.

Issue #800, opened 2020-10-16
Mana Integration Tests


lzpap:
Integration tests for mana plugin

 - [x] mana calculation
 - [x] webapi
 - [x] mana state persistence
 - [x] allowed pledgeIDs
 - etc.

acha-bill:
Mana calculations have unit tests.

Issue #749, opened 2020-09-23
Mana State Persistence


lzpap:
Base Mana Vectors are stored in database upon shutdown of a node, and then loaded upon restart, to persist the mana state.
However, this mechanism doesn't always work as expected in its current form, because the base vector storage doesn't always shut down correctly. As a result, after restart, all mana entries are initialized to zero. Then, when a new transaction confirms and revokes mana, base mana values become negative and node crashes. This is exactly what is happening in `FaucetPersistence` test: https://github.com/iotaledger/goshimmer/pull/743/checks?check_run_id=1149068089

TODO: debug and fix mana state persistence

lzpap:
close, because error is actually caused by #806 . The issue described above is addresses via #801 .

PR #803, opened 2020-10-16
feat: mana test persistence

acha-bill:
Integration test for mana persistence.

Part of #800

### DIFF
--- a/tools/integration-tests/tester/framework/parameters.go
+++ b/tools/integration-tests/tester/framework/parameters.go
@@ -43,7 +43,7 @@ var (
 	// ParaSyncBeaconOnEveryNode defines whether all nodes should be sync beacons.
 	ParaSyncBeaconOnEveryNode = false
 	// ParaManaOnEveryNode defines whether all nodes should have mana enabled.
-	ParaManaOnEveryNode = false
+	ParaManaOnEveryNode = true
 )
 
 var (
--- a/tools/integration-tests/tester/framework/parameters.go
+++ b/tools/integration-tests/tester/framework/parameters.go
@@ -42,6 +42,8 @@ var (
 	ParaPoWFaucetDifficulty = 2
 	// ParaSyncBeaconOnEveryNode defines whether all nodes should be sync beacons.
 	ParaSyncBeaconOnEveryNode = false
+	// ParaManaOnEveryNode defines whether all nodes should have mana enabled.
+	ParaManaOnEveryNode = false
 )
 
 var (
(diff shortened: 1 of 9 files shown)


tools/integration-tests/tester/tests/mana/mana_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/iotaledger/goshimmer/tools/integration-tests/tester/tests"
+	"github.com/stretchr/testify/require"
+)
+
+func TestManaPersistence(t *testing.T) {
+	n, err := f.CreateNetwork("mana_TestPersistence", 1, 0, true)
```
lzpap, line 12: let's create at least 5 nodes. If you take a look at the faucet tests, there is usually 1 out of 5 that fails the faucet persistence test for the same reason this test should fail

acha-bill: yeah but that shouldn't matter here. This tests that `a` node persists its mana and it starts up with the stored values.
Fixing #749 is another thing.

lzpap: okay

tools/integration-tests/tester/framework/parameters.go
```diff
@@ -42,6 +42,8 @@ var (
 	ParaPoWFaucetDifficulty = 2
 	// ParaSyncBeaconOnEveryNode defines whether all nodes should be sync beacons.
 	ParaSyncBeaconOnEveryNode = false
+	// ParaManaOnEveryNode defines whether all nodes should have mana enabled.
+	ParaManaOnEveryNode = false
```
lzpap, line 46:
```suggestion
	ParaManaOnEveryNode = true
```
mana is an essential component which is enabled by default on a node, why wouldn't we enable it then on all nodes in tests?

tools/integration-tests/tester/tests/mana/mana_test.go
```diff
@@ -0,0 +1,47 @@
...
+		err = peer.Start()
+		require.NoError(t, err)
+	}
+
+	// wait for peers to start
+	time.Sleep(5 * time.Second)
+
+	info, err = peers[0].Info()
+	require.NoError(t, err)
+	manaAfter := info.Mana
```
lzpap, line 44: with default mana parameters, we know that during this test, `manaBefore` < `manaAfter` should hold true. maybe add an extra check for this?

lzpap (approved):

PR #805, opened 2020-10-20
feat: Add commit tag to version

### LINKED ISSUE #494: Add commit tag to version (similar to Hornet)
**Is your feature request related to a problem? Please describe.**
When building goshimmer from source it is not apparent what commit is running.
 
**Describe the solution you'd like**
Provide a way to append current commit tag to the version via the -X build parameter. This is done in Hornet via the [build script](https://github.com/gohornet/hornet/blob/master/scripts/build_hornet.sh)

**Additional context**
Currently banner.AppVersion is defined as a constant and this not able to be overridden in the build process. Changing this to a var causes the portcheck plugin to fail

jkrvivian:
# Description of change
- Enable users to build with specific version and commit hash.
- Add `SimplifiedAppVersion` which contains `vXX.XX.XX` only, for use by other plugins, e.g. `analysis`, `remotelog` to send the consistent packets.

Fix #494 

## Type of change
- Enhancement 

Build with
`go build -ldflags="-s -w -X github.com/iotaledger/goshimmer/plugins/banner.AppVersion=<tag>-<commit_hash>`

and run the executed file

## Change checklist

### DIFF
--- a/plugins/remotelog/plugin.go
+++ b/plugins/remotelog/plugin.go
@@ -105,7 +105,7 @@ func run(plugin *node.Plugin) {
 
 func sendLogMsg(level logger.Level, name string, msg string) {
 	m := logMessage{
-		banner.AppVersion,
+		banner.SimplifiedAppVersion,
 		myGitHead,
 		myGitBranch,
 		myID,
--- a/plugins/remotelog/plugin.go
+++ b/plugins/remotelog/plugin.go
@@ -105,7 +105,7 @@ func run(plugin *node.Plugin) {
 
 func sendLogMsg(level logger.Level, name string, msg string) {
 	m := logMessage{
-		banner.SimplifiedAppVersion,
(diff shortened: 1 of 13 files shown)


plugins/remotelog/plugin.go
```diff
@@ -105,7 +105,7 @@ func run(plugin *node.Plugin) {
 
 func sendLogMsg(level logger.Level, name string, msg string) {
 	m := logMessage{
-		banner.AppVersion,
+		banner.SimplifiedAppVersion,
```
jonastheis, line 108: I think here we should actually send the `AppVersion`. Then we can specifically see which version is used.

jonastheis (commented):
Would be good to add a build script similar to https://github.com/gohornet/hornet/blob/master/scripts/build_hornet.sh

jonastheis (approved):

PR #807, opened 2020-10-22
test: mana allowed pledgeIDS

acha-bill:
Part of #800

### DIFF
--- a/tools/integration-tests/tester/framework/network.go
+++ b/tools/integration-tests/tester/framework/network.go
@@ -124,6 +124,9 @@ func (n *Network) CreatePeer(c GoShimmerConfig) (*Peer, error) {
 	if config.SyncBeaconBroadcastInterval == 0 {
 		config.SyncBeaconBroadcastInterval = 5
 	}
+	if config.SyncBeacon {
+		config.Seed = syncBeaconSeed
+	}
 
 	// create wallet
 	var nodeSeed *walletseed.Seed
--- a/tools/integration-tests/tester/framework/network.go
+++ b/tools/integration-tests/tester/framework/network.go
@@ -124,9 +124,6 @@ func (n *Network) CreatePeer(c GoShimmerConfig) (*Peer, error) {
 	if config.SyncBeaconBroadcastInterval == 0 {
 		config.SyncBeaconBroadcastInterval = 5
 	}
-	if config.SyncBeacon {
-		config.Seed = syncBeaconSeed
-	}
 
 	// create wallet
 	var nodeSeed *walletseed.Seed
--- a/tools/integration-tests/tester/tests/mana/mana_test.go
+++ b/tools/integration-tests/tester/tests/mana/mana_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestManaPersistence(t *testing.T) {
-	n, err := f.CreateNetwork("mana_TestPersistence", 1, 0, true)
+	n, err := f.CreateNetwork("mana_TestPersistence", 1, 0, framework.CreateNetworkConfig{Faucet: true, Mana: true})
 	require.NoError(t, err)
 	defer tests.ShutdownNetwork(t, n)
 
@@ -51,7 +51,7 @@ func TestManaPersistence(t *testing.T) {
 
 func TestAPI(t *testing.T) {
 	numPeers := 2
-	n, err := f.CreateNetwork("mana_TestAPI", 0, 0)
+	n, err := f.CreateNetwork("mana_TestAPI", 0, 0, framework.CreateNetworkConfig{})
 	require.NoError(t, err)
 	defer tests.ShutdownNetwork(t, n)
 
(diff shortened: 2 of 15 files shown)


tools/integration-tests/tester/tests/testutil.go
```diff
@@ -205,21 +207,15 @@ func SendTransactionOnRandomPeer(t *testing.T, peers []*framework.Peer, addrBala
 
 // SendIotaTransaction sends sentValue amount of IOTA tokens and remainders from and to a given peer and returns the fail flag and the transaction ID.
 // Every peer sends and receives the transaction on the address of index 0 by default.
-// Optionally, the address index for from and to can be specified.
-func SendIotaTransaction(t *testing.T, from *framework.Peer, to *framework.Peer, addrBalance map[string]map[balance.Color]int64, sentValue int64, fromAndToAddressIndex ...uint64) (fail bool, txId string) {
+// Optionally, the nodes to pledge access and consensus mana can be specified.
+func SendIotaTransaction(t *testing.T, from *framework.Peer, to *framework.Peer, addrBalance map[string]map[balance.Color]int64, sentValue int64, fromAddressIndex uint64, toAddressIndex uint64, manaPledgeIDs ...identity.ID) (fail bool, txId string) {
```
lzpap, line 211: I would rather have `accessPledgeID` and `consensusPledgeID` as individual arguments, like you did for `from/toAddressIndex`. Alternatively, we could introduce a `TransactionConfig` or `TransactionParameters` struct that this function accepts as argument. When you instantiate such a config struct, fields you leave out will be the zero value for their type, which we can detect in the function and apply default values to the actual transaction. For example, leaving out `from/toAddressIndex` will init them to 0, which is actually the default here. Or when you leave out `pledgeIDs`, their `emptyID` field denotes to use the `ID` of the `from` node, and so on. What do you think?

tools/integration-tests/tester/tests/mana/mana_test.go
```diff
@@ -45,3 +48,62 @@ func TestManaPersistence(t *testing.T) {
...
+	unspentOutputs, err := faucet.GetUnspentOutputs([]string{faucetAddrStr})
+	require.NoErrorf(t, err, "could not get unspent outputs on %s", faucet.String())
+	addrBalance[faucetAddrStr][balance.ColorIOTA] = unspentOutputs.UnspentOutputs[0].OutputIDs[0].Balances[0].Value
+
+	// pledge mana to allowed pledge
+	fail, _ := tests.SendIotaTransaction(t, faucet, allowedPeer, addrBalance, 100, 1, 0, allowedPeer.Identity.ID(), allowedPeer.Identity.ID())
+	require.False(t, fail)
+
+	// pledge mana to disallowed pledge
+	fail, _ = tests.SendIotaTransaction(t, faucet, disallowedPeer, addrBalance, 100, 1, 0, disallowedPeer.Identity.ID(), disallowedPeer.Identity.ID())
```
lzpap, line 107: why are we not spending from the second address of the faucet? Didn't the previous transaction above move the funds there?

lzpap (commented):
Looks okay, couple questions.

tools/integration-tests/tester/framework/network.go
```diff
@@ -124,6 +124,9 @@ func (n *Network) CreatePeer(c GoShimmerConfig) (*Peer, error) {
 	if config.SyncBeaconBroadcastInterval == 0 {
 		config.SyncBeaconBroadcastInterval = 5
 	}
+	if config.SyncBeacon {
+		config.Seed = syncBeaconSeed
+	}
```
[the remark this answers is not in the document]
acha-bill: When creating peers via `framework.CreateNetwork`, the syncbeacon (1st peer, just like the faucet) has a predefined seed: `framework.syncBeaconSeed`.
This was lacking if you created the peers via `network.createPeer`.

tools/integration-tests/tester/tests/mana/mana_test.go
acha-bill: Yes. We should

tools/integration-tests/tester/tests/testutil.go
acha-bill: Good idea

tools/integration-tests/tester/framework/network.go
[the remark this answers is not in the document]
acha-bill: This is actually bad. Coz in the syncbeacon tests where there are several beacons, this will set all of the them to use the same seed.
So I'm getting rid of this.

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -262,13 +261,6 @@ func SendIotaTransaction(t *testing.T, from *framework.Peer, to *framework.Peer,
 
 	// sign transaction
 	var txn *transaction.Transaction
-	emptyID := identity.ID{}
-	if txConfig.AccessManaPledgeID == emptyID {
-		txConfig.AccessManaPledgeID = local.GetInstance().ID()
```
lzpap: instead of `local`, this should be the ID of the from peer, same below

acha-bill: It's already being set to the ID of the peer in the test.
That's why this is deleted.

lzpap: ah, okay then

lzpap (approved):

PR #815, opened 2020-10-29
Use clock in network delay

capossele:
This PR adds the use of the `SyncedTime()` (if enabled) in the network delay app. It also adds two new fields in the `networkDelay` struct: 

- Clock: whether the clock plugin is enabled
- Sync: whether the node is in sync

### DIFF
--- a/plugins/networkdelay/plugin.go
+++ b/plugins/networkdelay/plugin.go
@@ -4,11 +4,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/iotaledger/goshimmer/packages/clock"
 	"github.com/iotaledger/goshimmer/packages/tangle"
 	"github.com/iotaledger/goshimmer/plugins/autopeering/local"
 	"github.com/iotaledger/goshimmer/plugins/config"
 	"github.com/iotaledger/goshimmer/plugins/messagelayer"
 	"github.com/iotaledger/goshimmer/plugins/remotelog"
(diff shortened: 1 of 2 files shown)


plugins/networkdelay/plugin.go
```diff
@@ -39,6 +41,9 @@ var (
 	myID            string
 	myPublicKey     ed25519.PublicKey
 	originPublicKey ed25519.PublicKey
+
+	// clockEnabled defines if the clock plugin is enabled.
+	clockEnabled bool
```
jonastheis, line 46: I don't think we need this. We should always use `clock.SyncedTime()` and if the clock is not actually synced (i.e. the plugin is disabled) we just return `time.Now().Add(offset)` where `offset=0`, effectively resulting in `time.Now()`. Makes the code much simpler :)

jonastheis (approved):

jonastheis (changes requested):
Oopsi pressed the wrong button!

plugins/networkdelay/plugin.go
capossele: But I would still keep that variable so that we can use it in the networkDelay struct to tell whether the node is using the clock or not. wdyt?

jonastheis: Yep that makes sense!

jonastheis (approved):

PR #817, opened 2020-11-02
pending mana predictor

acha-bill:
# Description of change

Predicts how much mana will be pledged by spending an output.
#817 

- Added API to query pending mana for a given `outputID`
- Added command in cli wallet to get pending mana for unspent outputs `pending-mana`
```
IOTA Pollen CLI-Wallet 0.1

Pending Mana for all unspent outputs
-----------------------------------
Address: SSZSK5kDZSgvirg9fxyJ46ELBwKLuRabCXLcTXTWZaRk
        OutputID: 8ZLJ3vT7z3vENqFYZfGRiqZMfRCFhxqUw9mjCPUBJZvxiWNSbj6WAQmvHGqFdTZC2P3pt1cjrMWP3g6EoFk4gAr1 - Pending Mana: 1327.405267
```
- Added pending mana to explorer output details page
Screenshot 2020-11-02 at 17 17 53

## Type of change

- unit tests

## Change checklist

### DIFF
--- a/client/wallet/output.go
+++ b/client/wallet/output.go
@@ -29,5 +29,6 @@ type InclusionState struct {
 
 // OutputMetadata is metadata about the output.
 type OutputMetadata struct {
+	// Timestamp is the timestamp of the tx that created the output.
 	Timestamp time.Time
 }
--- a/client/wallet/output.go
+++ b/client/wallet/output.go
@@ -1,6 +1,8 @@
 package wallet
 
 import (
+	"time"
+
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/address"
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/balance"
 	"github.com/iotaledger/goshimmer/dapps/valuetransfers/packages/transaction"
@@ -8,10 +10,12 @@ import (
 
 // Output is a wallet specific representation of an output in the IOTA network.
 type Output struct {
+	ID             transaction.OutputID
 	Address        address.Address
 	TransactionID  transaction.ID
 	Balances       map[balance.Color]uint64
 	InclusionState InclusionState
+	Metadata       OutputMetadata
 }
 
 // InclusionState is a container for the different flags of an output that define if it was accepted in the network.
@@ -21,4 +25,9 @@ type InclusionState struct {
 	Rejected    bool
 	Conflicting bool
 	Spent       bool
+}
+
+// OutputMetadata is metadata about the output.
+type OutputMetadata struct {
+	Timestamp time.Time
 }
--- a/packages/mana/base.go
+++ b/packages/mana/base.go
@@ -37,6 +37,11 @@ func (bm *BaseMana) updateBM2(n time.Duration) {
 	bm.BaseMana2 = bm.BaseMana2 * math.Pow(math.E, -decay*n.Seconds())
 }
 
+// GetBM2 computes and returns the EBM2 based on the value and duration specified.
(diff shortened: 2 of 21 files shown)


packages/mana/base.go
```diff
@@ -34,17 +34,22 @@ func (bm *BaseMana) updateEBM1(n time.Duration) {
 }
 
 func (bm *BaseMana) updateBM2(n time.Duration) {
-	bm.BaseMana2 = bm.BaseMana2 * math.Pow(math.E, -decay*n.Seconds())
+	bm.BaseMana2 = bm.BaseMana2 * math.Pow(math.E, -Decay*n.Seconds())
+}
+
+// GetBM2 computes and returns the EBM2 based on the value and duration specified.
+func GetBM2(value float64, n time.Duration) float64 {
+	return value * math.Pow(math.E, -Decay*n.Seconds())
```
lzpap, line 42: this doesn't make sense: `GetBM2` that returns `EBM2`?
We want to calculate the `pending mana`, that is, the amount of `BM2` that would be pledged at time `t` by spending the output. It has nothing to do with `EBM2`, and is calculated differently.
`value * math.Pow(math.E, -Decay*n.Seconds())` gives you a decreasing function, but we need the opposite. The correct equation is on line 96.
Rename this function to `GetPendingMana` or smthing more descriptive, with the calculation from line 96. Or, since now you have `Decay` exposed, you could place this calculation anywhere. Functions that are not used in `base.go` (and don't need data from there) should be in the plugin imho.

tools/cli-wallet/pendingmana.go
```diff
@@ -0,0 +1,43 @@
...
+			}
+			pendingMana := getBM2(total, time.Since(output.Metadata.Timestamp), status.ManaDecay)
+			fmt.Printf("\tOutputID: %s - Pending Mana: %f\n", output.ID.String(), pendingMana)
+		}
+	}
+	fmt.Println()
+}
+
+func getBM2(value float64, n time.Duration, decay float64) float64 {
+	return value * math.Pow(math.E, -decay*n.Seconds())
```
lzpap, line 42: this is not the equation you are looking for...

client/wallet/output.go
```diff
@@ -22,3 +26,8 @@ type InclusionState struct {
 	Conflicting bool
 	Spent       bool
 }
+
+// OutputMetadata is metadata about the output.
+type OutputMetadata struct {
+	Timestamp time.Time
```
lzpap, line 32: maybe a comment that it is the timestamp of the tx that created this output

plugins/dashboard/frontend/src/app/components/ExplorerAddressResult.tsx
```diff
@@ -117,7 +117,8 @@ export class ExplorerAddressQueryResult extends React.Component<Props, any> {
                             <div>{status}</div>
                             <div>{consumed}</div>
                             <div>{conflicting}</div>
-                            <div>{'Balance:'} {balances}</div>   
+                            <div>{'Balance:'} {balances}</div>
+                            <div>Pending Mana: {output.pending_mana}</div>
```
lzpap, line 121: there is a `displayManaUnit` function in `ManaGauge`, I think we could put it into some utils or helpers dir and reuse it here too.

lzpap (commented):
Please add back all `packr` related files and change the pending mana equation.

lzpap (approved):
lgtm!
write a comment on `Decay` and let's merge this.

PR #819, opened 2020-11-04
Refactor message structure

lzpap:
Refactoring message structure to account for `strong`/`weak` parents.

Changes:
 - [x] Rename `marhalutil` size variables to UpperCamelCase, due to `hive.go` change.
 - [x] `MessageIDLength` changes to 32 bytes.
 - [x] Introduce `MessageVersion` to distinguish between message layout versions.
 - [x] Messages have `strong` and `weak` parents instead of fixed 2 parents. A message must have at least one strong parent, and all together max 8 parents (strong and weak together). Strong parents express strong approval, while weak parents mean that the payload of the weak parent is liked, but the message itself does not.
- [x] Change marshaling and umarshaling of messages to/from binary. Serialized version of a message has a `ParentsCount` byte, a `ParentsType` bitmask, and parent `MessageID`s are sorted into lexicographic order, ASC.
- [x] `MessageID` is the `blake2b-256` hash of all message bytes, removed `ContentID`.
- [x] Exhaustive unit test for `MessageID` and `Message` structs.
- [x] `TipSelector` interface changes, `Tips()` awaits a number of tips to return instead of 2 fixed.
- [x] `MessageTipSelector.Tips()` refactored, fetches `count` number of `RandomUniqueEntries` from available tip list.
- [x] `MessageFactory` adjusted, still uses hardcoded 2 strong parents.
- [x] Max payload size changed to 65157 bytes. 
```
MaxPayloadSize = MaxMessageSize - [version(1) + parentsCount(1) + parentsType(1) + maxParents(8) * 32 +
issuerPK(32) + issuanceTime(8) + seqNum(8) + nonce(8) + signature(64) ]=
MaxMessageSize - 379 bytes = 65157
```
 - [x] Unit tests adjusted throughout the codebase. 
 - [x] Local dashboard message feed handling for parents.
 - [x] Local dashboard explorer handling for parents.
 - [x] Local dashboard visualizer handling for parents.
 - [x] WebApi handling for parents.


capossele (approved):

packages/tangle/message.go
```diff
@@ -119,19 +159,28 @@ type Message struct {
...
+	idMutex    sync.RWMutex
+	bytes      []byte
+	bytesMutex sync.RWMutex
 }
 
 // NewMessage creates a new message with the details provided by the issuer.
-func NewMessage(parent1ID MessageID, parent2ID MessageID, issuingTime time.Time, issuerPublicKey ed25519.PublicKey, sequenceNumber uint64, payload payload.Payload, nonce uint64, signature ed25519.Signature) (result *Message) {
+func NewMessage(strongParents []MessageID, weakParents []MessageID, issuingTime time.Time, issuerPublicKey ed25519.PublicKey, sequenceNumber uint64, payload payload.Payload, nonce uint64, signature ed25519.Signature) (result *Message) {
+	// syntactical validation
+	parentsCount := len(strongParents) + len(weakParents)
```
acha-bill, line 171: It would be clearer if parents were sorted and filtered before counting imho.

packages/tangle/message.go
```diff
@@ -1,56 +1,81 @@
...
+	// StrongParent identifies a strong parent in the bitmask.
+	StrongParent uint8 = 1
+
+	// WeakParent identifies a weak parent in the bitmask.
+	WeakParent uint8 = 0
 
-// ContentID identifies the content of a message without its parent1/parent2 ids.
-type ContentID = MessageID
+	// MinParentsCount defines the minimum number of parents a message must have.
+	MinParentsCount = 1
```
acha-bill, line 41: Should this not be `2`?

acha-bill (approved):
lgtm. 
Message tests is 🔥

packages/tangle/message.go
lzpap: you are right

packages/tangle/message.go
lzpap: we decided to make it `1` because you can't get `2` unique parents if there is only one tip, for example during bootstrapping the tangle. In current implementation, you just have `parent1` = `parent2` in these cases, which is equivalent to have 1 strong parent.

Issue #773, opened 2020-10-05
Enrich Pollen Analyzer with Mana


lzpap:
- Autopeering node graph, where (size or color) of a node corresponds to its mana value.
- Display mana leaderboard + stats from local dashboard so people can take a look without running a node.  As an input, a high mana node's perception should be used (entry node? faucet node?).

PR #823, opened 2020-11-10
feat: add mana to analyzer

acha-bill:
# Description of change

Adds mana to analysis dashboard. 
`analysis.dashboard.manaAddress` defines the address of the mana dashboard to stream mana info from.

Screenshot 2020-11-10 at 02 19 55

On the visualizer, the size of the node corresponds to its mana percentile across the network.
Screenshot 2020-11-11 at 04 16 05

Part of #773 

## Type of change

- Tests pass

## Change checklist


plugins/analysis/client/plugin.go
```diff
@@ -63,6 +69,20 @@ func run(_ *node.Plugin) {
 		valuetransfers.Voter().Events().RoundExecuted.Attach(onRoundExecutedClosure)
 		defer valuetransfers.Voter().Events().RoundExecuted.Detach(onRoundExecutedClosure)
 
+		mana.Events().Pledged.Attach(events.NewClosure(func(ev *mana.PledgedEvent) {
+			if len(pledgeEvents) >= config.Node().GetInt(CfgMaxManaEventBufferSize) {
```
lzpap, line 73: let's have a `maxManaEventBufferSize` variable instead, which is initialized at the beginning like `conn`. No need to invoke the config plugin every time we receive an event, because this will not change.

plugins/analysis/client/plugin.go
```diff
@@ -73,9 +93,36 @@ func run(_ *node.Plugin) {
...
 }
+
+// GetPledgeEvents gets the cached mana pledge events and clears the cache afterwards.
+func GetPledgeEvents() []mana.PledgedEvent {
+	var evs []mana.PledgedEvent
+	for _, ev := range pledgeEvents {
+		evs = append(evs, *ev)
+	}
+	// clear buffer
+	pledgeEvents = []*mana.PledgedEvent{}
```
lzpap, line 115: protect the buffers with a mutex, since events can trigger asynchronously.

plugins/analysis/client/plugin.go
```diff
@@ -73,9 +93,36 @@ func run(_ *node.Plugin) {
 			case <-ticker.C:
 				sendHeartbeat(conn, createHeartbeat())
 				sendMetricHeartbeat(conn, createMetricHeartbeat())
+				manaHb, err := createManaHeartbeat()
```
lzpap, line 96: I think we need more fine grained control over what a client is sending. In this form, all clients will flood the server with mana hb packets every 5s. Since we only need one reliable source in the analyzer,  sending mana hb packets should be disabled by default, and has to be explicitly enabled in config. On top of course we can discard mana hb packets from other nodes. My original thinking was that we don't even need the mana hb packets, why don't we just execute api calls to a node we configure from the anaylzer server? That seem to be less hassle, and analysis client can stay as it is. Why add more code and logic if we already have the api calls for this? (except for the events, true) 
What if we connect the analyzer dashboard frontend to the WS msg stream of the local dashboard of a node we choose? We would have all the mana data without using either the API or the heartbeats.

acha-bill: >What if we connect the analyzer dashboard frontend to the WS msg stream of the local dashboard of a node we choose?

That's a great idea.

plugins/analysis/dashboard/frontend/src/app/components/Mana/Mana.tsx
```diff
@@ -0,0 +1,135 @@
...
+@observer
+export default class Mana extends React.Component<Props, any> {
+    render() {
+        let manaStore = this.props.manaStore
+        const {searchNode, searchTxID} = manaStore
+        return (
+            <Container>
+                <Row className={"mb-3 mt-3"}>
+                    <Col>
+                        <ManaGauge
```
lzpap, line 26: I don't think we should show this ManaGauge here, it is for a node, not for the whole network.

plugins/analysis/dashboard/frontend/src/app/models/mana/IManaMessage.ts
```diff
@@ -0,0 +1,7 @@
+export interface IManaMessage {
+    nodeID: string;
+    access: number;
+    consensus: number;
+    // in s?
```
lzpap, line 5: yep, this is unix time in seconds

plugins/analysis/dashboard/frontend/src/app/models/ws/wsMsgType.ts
```diff
@@ -4,5 +4,13 @@ export enum WSMsgType {
...
-}
+    disconnectNodes = 5,
+    Mana= 8,
+    ManaMapOverall,
+    ManaMapOnline,
+    ManaAllowedPledge,
+    ManaPledge,
+    ManaRevoke,
+    MsgManaDashboardAddress,
+    MsgReqManaDashboardAddress
```
lzpap, line 15: `MsgManaDashboardAddress` and `MsgReqManaDashboardAddress` are legit here, as they are msg types for communicating with the analysis dashboard. The other, mana related types are for communicating with a local dashboard, they don't belong here. Put them in WSMsgTypeDashboard or smthing. Let's not mix the analysis ws message enums with the dashboard ws message enums.

plugins/analysis/dashboard/frontend/src/app/stores/AutopeeringStore.tsx
```diff
@@ -364,6 +374,7 @@ export class AutopeeringStore {
             () => this.updateWebSocketConnected(true),
             () => this.updateWebSocketConnected(false),
```
lzpap, line 375: add reconnect feature on lost ws conn

plugins/analysis/dashboard/frontend/src/app/stores/ManaStore.tsx
```diff
@@ -0,0 +1,521 @@
...
+
+    @action
+    public updateDashboardWebsocketConnect(connected: boolean): void {
+        this.dashboardWebsocketConnected = connected
+    }
+
+    public connect(): void {
+        connectDashboardWebSocket(this.manaDashboardAddress,
+            () => this.updateDashboardWebsocketConnect(true),
+            () => this.updateDashboardWebsocketConnect(false),
```
lzpap, line 172: add reconnect feature on lost ws conn

plugins/analysis/dashboard/ws.go
```diff
@@ -100,20 +111,57 @@
...
+			if err := ws.WriteJSON(msg); err != nil {
+				break
+			}
+			if err := ws.SetWriteDeadline(time.Now().Add(webSocketWriteTimeout)); err != nil {
+				break
+			}
+		case <-shutdownSignal:
+			return nil
+		default:
+			_, msg, err := ws.ReadMessage()
```
lzpap, line 141: is this blocking if there are no incoming msgs?

lzpap:
@acha-bill  There is some probelm with sizing the nodes, probably node selection doesn't like it either.
image

plugins/analysis/dashboard/ws.go
acha-bill: no

plugins/analysis/dashboard/frontend/src/app/components/Mana/Mana.tsx
acha-bill: Lets use this gauge to show the total mana in the network

lzpap: Good idea!

lzpap:
Hey @acha-bill, nice job, we are getting there!
It might worth looking into addressing the following things:
- `packrd/packed-packr.go` is still there, do we need it for anything?
- `ManaGauge` should display mana of one node, not overall.
- weird error when closing a ws client:
```
entry_node      | 2020-11-18T09:59:57Z  ERROR   Analysis-Dashboard      dashboard/ws.go:98      recovered from websocket handle func: repeated read on failed websocket connection
entry_node      | github.com/iotaledger/goshimmer/plugins/analysis/dashboard.websocketRoute.func1
entry_node      |       /tmp/goshimmer/plugins/analysis/dashboard/ws.go:98
entry_node      | runtime.gopanic
entry_node      |       /usr/local/go/src/runtime/panic.go:969
entry_node      | github.com/gorilla/websocket.(*Conn).NextReader
entry_node      |       /go/pkg/mod/github.com/gorilla/websocket@v1.4.2/conn.go:1001
entry_node      | github.com/gorilla/websocket.(*Conn).ReadMessage
entry_node      |       /go/pkg/mod/github.com/gorilla/websocket@v1.4.2/conn.go:1064
entry_node      | github.com/iotaledger/goshimmer/plugins/analysis/dashboard.websocketRoute
entry_node      |       /tmp/goshimmer/plugins/analysis/dashboard/ws.go:141
entry_node      | github.com/labstack/echo.(*Echo).Add.func1
entry_node      |       /go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/echo.go:490
entry_node      | github.com/labstack/echo/middleware.RecoverWithConfig.func1.1
entry_node      |       /go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/middleware/recover.go:78
entry_node      | github.com/labstack/echo.(*Echo).ServeHTTP
entry_node      |       /go/pkg/mod/github.com/labstack/echo@v3.3.10+incompatible/echo.go:593
entry_node      | net/http.serverHandler.ServeHTTP
entry_node      |       /usr/local/go/src/net/http/server.go:2807
entry_node      | net/http.(*conn).serve
entry_node      |       /usr/local/go/src/net/http/server.go:1895
```
- some comments are missing, reviewdog complains about ineffective break statements in `plugins/analysis/dashboard/ws.go`.
- size of nodes in autopeering graph only updated in mana messages. It would be better from UX perspective to keep the size even if I select a detailed node view and then return back to complete view. At the moment, after such an action, size of the nodes remain default until the next mana msg comes. Results in "visual jumping" behavior.

acha-bill:
>packrd/packed-packr.go is still there, do we need it for anything?
We don't need it?

>ManaGauge should display mana of one node, not overall.
Does not make sense to show on the analyzer the mana of a specific node.

>weird error when closing a ws client:
fixed

>some comments are missing, reviewdog complains about ineffective break statements in plugins/analysis/dashboard/ws.go.
Fixed

>size of nodes in autopeering graph only updated in m...
fixed

lzpap:
> > packrd/packed-packr.go is still there, do we need it for anything?
> 
> We don't need it?
> 
> > ManaGauge should display mana of one node, not overall.
> 
> Does not make sense to show on the analyzer the mana of a specific node.
> 
> > weird error when closing a ws client:
> 
> fixed
> 
> > some comments are missing, reviewdog complains about ineffective break statements in
> > plugins/analysis/dashboard/ws.go.
> 
> Fixed
> 
> > size of nodes in autopeering graph only updated in m...
> 
> fixed

1. We don't need `packrd/packed-packr.go`. What does it pack?
2. Sorry, my bad, I meant that ManaGauge should display total mana but currently it displays  the mana of a node by looking at the code.

PR #840, opened 2020-11-18
Adds FPC statements

capossele:


plugins/consensus/opiniongiver.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		if _, ok := opinionGiversMap[p.ID().String()]; !ok {
+			opinionGiversMap[p.ID().String()] = &OpinionGiver{
+				id: p.ID().String(),
+			}
+		}
+		opinionGiversMap[p.ID().String()].pog = &PeerOpinionGiver{p: p}
+	}
+
+	for _, v := range opinionGiversMap {
```
jonastheis, line 81: I assume you copy here for thread safety. Please add a comment. Also for each `for` loop above a short comment would be nice.

plugins/consensus/opiniongiver.go
```diff
@@ -0,0 +1,168 @@
...
+
+// PeerOpinionGiver implements the OpinionGiver interface based on a peer.
+type PeerOpinionGiver struct {
+	p *peer.Peer
+}
+
+// Query queries another node for its opinion.
+func (pog *PeerOpinionGiver) Query(ctx context.Context, conflictIDs []string, timestampIDs []string) (vote.Opinions, error) {
+	fpcServicePort := pog.p.Services().Get(service.FPCKey).Port()
+	fpcAddr := net.JoinHostPort(pog.p.IP().String(), strconv.Itoa(fpcServicePort))
```
jonastheis, line 100: This feels like it should be an utility function somewhere.

plugins/consensus/plugin.go
```diff
@@ -0,0 +1,218 @@
...
+	valuetransfers.FCOB().Events.Vote.Attach(events.NewClosure(func(id string, initOpn vote.Opinion) {
+		if err := Voter().Vote(id, vote.ConflictType, initOpn); err != nil {
+			log.Warnf("FPC vote: %s", err)
+		}
+	}))
+	valuetransfers.FCOB().Events.Error.Attach(events.NewClosure(func(err error) {
+		log.Errorf("FCOB error: %s", err)
+	}))
+
+	configureFPC()
```
jonastheis, line 97: Shouldn't this happen before attaching to any events?

plugins/consensus/plugin.go
```diff
@@ -0,0 +1,218 @@
...
+		if err := lPeer.UpdateService(service.FPCKey, "tcp", port); err != nil {
+			log.Fatalf("could not update services: %v", err)
+		}
+	}
+
+	Voter().Events().RoundExecuted.Attach(events.NewClosure(func(roundStats *vote.RoundStats) {
+		makeStatement(roundStats)
+		peersQueried := len(roundStats.QueriedOpinions)
+		voteContextsCount := len(roundStats.ActiveVoteContexts)
+		log.Infof("executed round with rand %0.4f for %d vote contexts on %d peers, took %v", roundStats.RandUsed, voteContextsCount, peersQueried, roundStats.Duration)
```
jonastheis, line 145: Too verbose?

plugins/consensus/plugin.go
```diff
@@ -0,0 +1,218 @@
...
+	return plugin
+}
+
+func configure(_ *node.Plugin) {
+	log = logger.NewLogger(ConsensusPluginName)
+	roundIntervalSeconds = config.Node().Int64(CfgFPCRoundInterval)
+	waitForStatement = config.Node().Int(CfgWaitForStatement)
+	listen = config.Node().Bool(CfgFPCListen)
+
+	valuetransfers.FCOB().Events.Vote.Attach(events.NewClosure(func(id string, initOpn vote.Opinion) {
```
jonastheis, line 88: Comments in code are useful :)

plugins/consensus/statement.go
```diff
@@ -0,0 +1,91 @@
...
+		issuerID := identity.NewID(msg.IssuerPublicKey()).String()
+
+		issuerRegistry := Registry().NodeView(issuerID)
+
+		for _, conflict := range statementPayload.Conflicts {
+			issuerRegistry.AddConflict(conflict)
+		}
+
+		for _, timestamp := range statementPayload.Timestamps {
+			issuerRegistry.AddTimestamp(timestamp)
```
jonastheis, line 88: These need to lock/unlock over and over again. Maybe it is more performant if we hand in the full lists and do it in bulk.

jonastheis (changes requested):
I think you can add some more comments in the code and structure it a bit better. But overall nice work! 

One question: currently we're not cleaning the registry up. How big does it get if a node runs for multiple days/weeks? Is that something we need to think about?

packages/vote/statement/opinion.go
```diff
@@ -0,0 +1,125 @@
...
+func (o Opinion) Bytes() (bytes []byte) {
+	bytes = make([]byte, OpinionLength)
+
+	// initialize helper
+	marshalUtil := marshalutil.New(bytes)
+
+	marshalUtil.WriteByte(byte(o.Value))
+	marshalUtil.WriteUint8(o.Round)
+
+	return bytes
```
jonastheis, line 37: I guess I missed this in the last review. Would be great to change it to the proper usage like for the others

jonastheis (approved):
Good stuff

PR #845, opened 2020-11-30
Fix ws panic dashboard

capossele:


plugins/dashboard/ws.go
```diff
@@ -90,11 +90,11 @@ func removeWsClient(clientID uint64) {
 	wsClientsMu.RLock()
 	wsClient := wsClients[clientID]
 	close(wsClient.exit)
+	close(wsClient.channel)
```
capossele, line 93: I'm not sure that by having the `wsClient.channel` closed here would solve the potential panic. That is because the loop we are reading from the channel is also protected by a read lock, and here we are closing the same channel within another read lock.

Wollac (approved):
lgtm

PR #850, opened 2020-12-04
Add statement analysis

capossele:


jonastheis (approved):
When merging we need to remember to restart the remote logger

PR #851, opened 2020-12-04
Add configurable statement

capossele:


tools/docker-network/docker-compose.yml
```diff
@@ -47,6 +47,7 @@ services:
       --valueLayer.snapshot.file=/tmp/assets/7R1itJx5hVuo9w9hjg5cwKFmek4HMSoBDgJZN8hKGxih.bin
       --syncbeacon.broadcastInterval=5
       --syncbeacon.startSynced=true
+      --statement.writeStatement=true
```
jonastheis, line 50: Don't want to replica's to write statements?

jonastheis (approved):

tools/docker-network/docker-compose.yml
capossele: Yes, that was the idea. Should we let everybody do it?

jonastheis: I guess it's fine if it's only the master. Was just wondering whether it was on purpose.

PR #852, opened 2020-12-04
Add registry cleaner

capossele:


packages/vote/statement/registry.go
```diff
@@ -102,11 +148,16 @@ func (v *View) AddTimestamp(t Timestamp) {
 	defer v.tMutex.Unlock()
 
 	if _, ok := v.Timestamps[t.ID]; !ok {
-		v.Timestamps[t.ID] = Opinions{t.Opinion}
+		v.Timestamps[t.ID] = Entry{
+			Opinions:  Opinions{t.Opinion},
+			Timestamp: time.Now(),
```
jonastheis, line 153: This is local to the node so it's fine to use `time.Now()`. But isn't it a bit confusing to decide when to use `time.Now() vs clock.SyncedTime()`? I think it could lead to errors if it is suddenly used somewhere else that is not local to a node anymore. Out of consistency and simplicity we should simply use `clock.SyncedTime()` everywhere. In the worst case (plugin disabled) it is just a wrapper for `time.Now()` but it saves us from thinking about which perspective of time to use.

capossele: The problem of always using the `clock.SyncedTime()` is that uses a lock and could decrease performance. Also, since it adds an offset, when using it for timers and co, might cause side effects.
But yeah, here it could be used

jonastheis (approved):

PR #857, opened 2020-12-04
Use clock-based time

capossele:


jonastheis (approved):
There are still quite some matches for `time.Now()`. But I don't think they are critical. Out of consistency we should probably switch it everywhere I guess?

image

PR #894, opened 2020-12-18
Caching for integration tests

jonastheis:


jonastheis:
For now not a priority. Maybe I will pick this up at a later point in time again.

Issue #884, opened 2020-12-14
Calculate consensus mana vector in past


lzpap:
## Description

To calculate active consensus mana set (`ACMS`) of an epoch (#854), we need to be able to determine the consensus mana values of nodes in the past, at a certain point in time, that is the end of the epoch. As described in the epoch specification proposal, we need to keep a log of the pledge/revoke events for consensus mana in order to recreate the consensus mana vector at any given time.

## DoD
A tool, that given a time in the past, returns the consensus mana vector at that time.

## Tips
 - Listen to `PledgeEvents` and `RevokeEvents` for consensus mana.
 - Record all events, order them chronologically.
 - Given a request for consensus mana at `t`, consume logged events  until `t` and recreate consensus mana vector at `t`.
 - To optimize above calculation, introduce "checkpoint" events every `x` event where you also store that consensus mana vector together with the event. Then you don't need to start from beginning when consuming the events.

How far back in time to keep the logged mana events?
 - Since this will be used for `ACMS` calculation of epochs, keeping the logs for only couple epochs (6-12 hours) should be fine. Theoretically, as soon as the `ACMS` for an epoch was calculated, all event logs  until the end of the epoch could be deleted.

## Nice to have:
 - Store event log in database, so events are persisted on restart.
 - API to retreive all logged events, with transaction details and local transaction metadata.
 - API to retreive logged events for a particular node.
 - Tool for visualizing consensus mana evolution over time based on these events.
 - Store all events in a csv file/mongodb/whatever for research purposes. We could analyze for example the transaction confirmation time by looking at tx timestamp and when the event was fired.

PR #896, opened 2020-12-20
consensus mana vector in the past

acha-bill:
Given a time `t` in the past, rebuild a consensus mana vector at `t`.

This logs all pledge/revoke events as they happen and saves them in storage. Then uses the logs to rebuild the base mana vector up to the requested time.

#884


packages/mana/persistableevent.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Update updates the event in storage.
+func (p *PersistableEvent) Update(objectstorage.StorableObject) {
+	panic("should not be updated")
+}
+
+// ObjectStorageKey returns the key of the persistable mana.
+func (p *PersistableEvent) ObjectStorageKey() []byte {
+	return []byte(p.TransactionID.String() + p.Time.String() + p.NodeID.String())
```
lzpap, line 50: ```p.TransactionID.String() + p.Time.String() + p.NodeID.String()``` is not a unique key, as a `Pledge` and a `Revoke` event that were caused by the same tx to happen, at the same time, for the same node would resolve to the same storage key. This might happen when I just move my funds to a different address, but from the same node.

plugins/mana/plugin.go
```diff
@@ -371,6 +395,91 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+
+	var eventLogs []mana.Event
+	var readErr error
+	consensusEventsLogStorage.ForEach(func(key []byte, cachedObject objectstorage.CachedObject) bool {
+		cachedPe := &mana.CachedPersistableEvent{CachedObject: cachedObject}
+		pe := cachedPe.Unwrap()
+		defer cachedPe.Release()
+		if pe.Time.After(t) {
+			return false
+		}
```
lzpap, line 431: this assumes that ```consensusEventsLogStorage.ForEach()`` iterates through the logged events in chronological order. Does it really?

plugins/mana/plugin.go
```diff
@@ -371,6 +395,91 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+		return nil, err
+	}
+
+	// store the most recent bmv
+	if err = consensusBaseManaPastVectorStorage.Prune(); err != nil {
+		return nil, err
+	}
+	for _, p := range cbmvPast.ToPersistables() {
+		consensusBaseManaPastVectorStorage.Store(p).Release()
+	}
```
lzpap, line 463: let's make this optional with a parameter `storePastReference` or smthing. Imagine that with this tool now we have the power to check if nodes had the exact same consensus mana perception for example 1 minute ago. But the main usecase for this being the calculation of ACMS after the end of the epochs.
With the following scenario:
 1. `GetPastConsensusManaVector()` gets called for `t`. (for anything else than epoch calc.)
 2. `consensusBaseManaPastVectorStorage` contains the bmv at `t`.
 3. `GetPastConsensusManaVector()` gets called for `t-10`. (for epoch calculation)
 4. Since `t-10` is before `t`, we skip building the initial past vector from `consensusBaseManaPastVectorStorage` entries, and have to consume ALL logged events.
 
So we should only store the latest bmv "checkpoint" when we are fairly sure that we will not need stuff from before (epoch calcualted). Ofc you could still query older timestamps, but that would be a fairly expensive operation compared to consuming only events since the checkpoint.

Another solution would be to have multiple checkpoints stored, let's say every X mins or X hours, then the amount of events you need to process would be determined by TPS.

plugins/mana/plugin.go
```diff
@@ -371,6 +395,91 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+
+	if !cachedMetadata.Exists() {
+		consensusBaseManaPastVectorMetadataStorage.Store(metadata)
+	} else {
+		m := cachedMetadata.Unwrap()
+		m.Update(metadata)
+	}
+
+	// TODO: delete logs? what if we want to go further in the past
```
lzpap, line 478: Logs should be deleted after some time from the object storage, but let's archive them into something standalone, like a csv file or mongoDB or whatever. So even if the node crashes, db is dirty and deleted, we have them somewhere for analysis.

packages/mana/persistableevent.go
acha-bill: you are right.

lzpap:
- add inputID to `RevokeEvent` to be able to uniquely identify each revoke event. One transaction might cause multiple revoke events for the same nodeID by spending multiple outputs.
- `PledgeEvent` is already unique, becasue a transaction can only pledge access/consensus mana to one node.

plugins/mana/plugin.go
acha-bill: As discussed, we store only the latest bmv every X events.

packages/mana/consensusbasevector_test.go
```diff
@@ -653,3 +654,91 @@ func TestConsensusBaseManaVector_ToAndFromPersistable(t *testing.T) {
 	}
 	assert.Equal(t, bmv.(*ConsensusBaseManaVector).vector, restoredBmv.(*ConsensusBaseManaVector).vector)
 }
+
+func TestConsensusBaseManaVector_BuildPastBaseVector(t *testing.T) {
```
lzpap, line 658: Let's extend this test with new cases:
 - called with empty event log set,
 - called with time that is older than any of the logged events,
 - called with a time that causes some events not to be consumed in event log,
 - accidentally called with events that are already included in the vector, revoke should panic due to negative mana values.

Also, see below my comments on the constructed `txInfo`s.

packages/mana/consensusbasevector_test.go
```diff
@@ -653,3 +654,91 @@ func TestConsensusBaseManaVector_ToAndFromPersistable(t *testing.T) {
...
+	bmv, err := NewBaseManaVector(ConsensusMana)
+	assert.NoError(t, err)
+	var eventsLog []Event
+	emptyID := identity.ID{}
+
+	tx1Info := &TxInfo{
```
lzpap, line 664: As I understand, the idea here is to have 3 transactions.
 - Tx1 spends 10 iotas from genesis output.
 - Tx2 spends 5 iotas from Tx1 outputs.
 - Tx3 spends 5 iotas from Tx1 outputs, 5 iotas from Tx2 output.

Assuming above scenario, `txInfo` definitions should be modified:
 - Tx1 can stay the same.
 - Tx2 timestamp should be `txTime+delta`, Tx2 `InputInfo` timestamp should be `txTime`, as it corresponds to the output created by Tx1.
 - Tx3 timestamp should be `txTime+2*delta`, `InputInfo` timestamps modified to correspond to Tx1 and Tx2 timestamps.

packages/mana/consensusbasevector.go
```diff
@@ -36,6 +36,66 @@ func (c *ConsensusBaseManaVector) Has(nodeID identity.ID) bool {
 	return exists
 }
 
+// BuildPastBaseVector builds a consensus base mana vector from past events upto time `t`
+func (c *ConsensusBaseManaVector) BuildPastBaseVector(eventsLog []Event, t time.Time) error {
+	emptyID := identity.ID{}
+	if len(c.vector) == 0 {
+		c.vector = make(map[identity.ID]*ConsensusBaseMana)
```
lzpap, line 43: a freshly created vector (with `NewBaseManaVector()`) has length 0, but the map is initialized. Check for `nil` instead.

plugins/mana/plugin.go
```diff
@@ -371,8 +408,253 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+		return nil, nil, err
+	}
+
+	return cbmvPast, eventLogs, nil
+}
+
+func checkConsensusEventStorage() {
+	size := consensusEventsLogStorage.GetSize()
+	if size < maxConsensusEventsInStorage {
+		return
```
lzpap, line 571: When we reach this number, the function body runs and creates a `ConsensusBaseManaVectorPast` at `now`, and prunes all the events up to this point. So If I wanted to ask for the vector at `now-t`, I would fail, as there are no more events in the log. Of course I could still go to the csv file.
But my problem with this approach is that it doesn't leave me with a buffer of time in the past which would let me get the vector after the pruning has happened.
How about having a sliding window instead?

mana_event_storage

Instead of consuming all events up until `now` when building the `PastConsensusBaseManaVector`, consume only `SlidingInterval` number of events.
- How to determine then what `t` to feed into `GetPastConsensusManaVector(t)`? (or how to determine `t_newCBMV` on the figure): just like you do the chronological event sort within GetPastConsensusManaVector, you could get the `SlidingInterval` number oldest events from the event storage. `t_newCBMV` then will be the timestamp of the youngest among these.
- Once done, delete the consumed events only from the event storage.
- Wait for the number of events to reach `MaxEventInStorage` , and repeat.

Then we would be able to still query the `ConsensusBaseManaVector` in `now-t_newCBMV`.

P.S: when selecting the edge at the sliding interval, be careful to not separate events belonging to the same transaction (they will have the same timestamp anyway).

plugins/mana/plugin.go
```diff
@@ -371,8 +408,253 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+	defer cachedMetadata.Release()
+
+	if !cachedMetadata.Exists() {
+		consensusBaseManaPastVectorMetadataStorage.Store(metadata).Release()
+	} else {
+		m := cachedMetadata.Unwrap()
+		m.Update(metadata)
+	}
+
+	err = writeEventsToCSV(logs)
```
lzpap, line 605: Yes, we could do it like this. Or an easier (and real time) approach is to have a separate background worker listening to `pledge/revoke` events, gathering them for couple seconds, and writing them to the file if there are things to write. Then we don't need too have the logic intertwined with the event storage. We can even have it as a separate extension plugin.

plugins/webapi/mana/logsHandler.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/iotaledger/goshimmer/packages/mana"
+	manaPlugin "github.com/iotaledger/goshimmer/plugins/mana"
+	"github.com/iotaledger/hive.go/identity"
+	"github.com/labstack/echo"
+	"github.com/mr-tron/base58"
+)
+
+// getEventLogsHandler handles the request.
+func getEventLogsHandler(c echo.Context) error {
```
lzpap, line 14: Would be nice to:
 - Include the time of the current oldest logged event in the response, to know that that is the "oldest" event I could expect.
 - Add a time range in the request.
 - Possibility to get all logs, not just selected nodes.
 - Return the events sorted.

plugins/webapi/mana/pastconsensusbasemanavector.go
```diff
@@ -0,0 +1,42 @@
...
+	})
+}
+
+// PastConsensusManaVectorRequest is the request.
+type PastConsensusManaVectorRequest struct {
+	Timestamp int64 `json:"timestamp"`
+}
+
+// PastConsensusManaVectorResponse is the response.
+type PastConsensusManaVectorResponse struct {
```
lzpap, line 39: include the timestamp in the response as well!

packages/mana/consensusbasevectormetadata_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/iotaledger/hive.go/marshalutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConsensusBasePastManaVectorMetadata_Bytes(t *testing.T) {
+	c := &ConsensusBasePastManaVectorMetadata{}
```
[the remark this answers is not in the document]
acha-bill: Not very relevant to the test though. The zero value of time is enough.

packages/mana/persistableevent.go
```diff
@@ -0,0 +1,171 @@
...
+
+// ObjectStorageKey returns the key of the persistable mana.
+func (p *PersistableEvent) ObjectStorageKey() []byte {
+	return p.Bytes()
+}
+
+// ObjectStorageValue returns the bytes of the event.
+func (p *PersistableEvent) ObjectStorageValue() []byte {
+	return p.Bytes()
+}
```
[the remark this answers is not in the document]
acha-bill: IMO we don't need this over-optimization on storage. Coz we will be pruning events anyways. The faster we can index the db the better.

plugins/mana/plugin.go
```diff
@@ -74,6 +83,9 @@ func configure(*node.Plugin) {
 		storages[mana.ResearchAccess] = osFactory.New(storageprefix.ManaAccessResearch, mana.FromObjectStorage)
 		storages[mana.ResearchConsensus] = osFactory.New(storageprefix.ManaConsensusResearch, mana.FromObjectStorage)
 	}
+	consensusEventsLogStorage = osFactory.New(storageprefix.ManaEventsStorage, mana.FromEventObjectStorage)
```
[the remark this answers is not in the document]
acha-bill: I don't follow.

plugins/mana/plugin.go
acha-bill: good point

plugins/mana/plugin.go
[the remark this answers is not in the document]
acha-bill: We have a package and storage prefix already. Do you mean a prefix on the persistable itself?

plugins/webapi/mana/logsHandler.go
acha-bill: >Include the time of the current oldest logged event in the response, to know that that is the "oldest" event I could expect.

would mean that we travel the whole storage every time even if we wanted to get logs only for 1 node.
I don't think its worth it.

plugins/manaeventlogger/plugin.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func configureEvents() {
+	mana.Events().Pledged.Attach(onPledgeEventClosure)
+	mana.Events().Revoked.Attach(onRevokeEventClosure)
+}
+
+func logPledge(ev *mana.PledgedEvent) {
+	eventsBuffer = append(eventsBuffer, ev)
+	checkBuffer()
```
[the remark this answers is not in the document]
lzpap: And also write the content of the buffer to the file when shutting down the worker.

plugins/webapi/mana/pastconsensusbasemanavector.go
```diff
@@ -0,0 +1,44 @@
...
+	var req PastConsensusManaVectorRequest
+	if err := c.Bind(&req); err != nil {
+		return c.JSON(http.StatusBadRequest, PendingResponse{Error: err.Error()})
+	}
+	timestamp := time.Unix(req.Timestamp, 0)
+	consensus, _, err := manaPlugin.GetPastConsensusManaVector(timestamp)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, PastConsensusManaVectorResponse{Error: err.Error()})
+	}
+	manaMap, err := consensus.GetManaMap()
```
lzpap, line 23: `GetManaMap()` calls `getMana()` under the hood, that was designed for real time tracking so it updates the `BaseMana` for the node wrt to current time. This is obviously not what we want. Add an `update` flag to `getMana()` arguments to be able to disable updating when needed.

plugins/mana/plugin.go
```diff
@@ -371,8 +413,299 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+	return logs, err
+}
+
+// GetPastConsensusManaVectorMetadata gets the past consensus mana vector metadata.
+func GetPastConsensusManaVectorMetadata() mana.ConsensusBasePastManaVectorMetadata {
+	cachedObj := consensusBaseManaPastVectorMetadataStorage.Get([]byte(mana.ConsensusBaseManaPastVectorMetadataStorageKey))
+	cachedMetadata := &mana.CachedConsensusBasePastManaVectorMetadata{CachedObject: cachedObj}
+	defer cachedMetadata.Release()
+	metadata := cachedMetadata.Unwrap()
+	return *metadata
```
lzpap, line 485: when we don't have a saved metadata, dereferencing the nil pointer causes a panic. Check for `nil` in the return of `Unwrap()`.

plugins/mana/plugin.go
```diff
@@ -29,19 +29,27 @@ import (
...
+	baseManaVectors                            map[mana.Type]mana.BaseManaVector
+	osFactory                                  *objectstorage.Factory
+	storages                                   map[mana.Type]*objectstorage.ObjectStorage
+	allowedPledgeNodes                         map[mana.Type]AllowedPledge
+	consensusBaseManaPastVectorStorage         *objectstorage.ObjectStorage
+	consensusBaseManaPastVectorMetadataStorage *objectstorage.ObjectStorage
+	consensusEventsLogStorage                  *objectstorage.ObjectStorage
+	onTransactionConfirmedClosure              *events.Closure
+	onPledgeEventClosure                       *events.Closure
+	onRevokeEventClosure                       *events.Closure
```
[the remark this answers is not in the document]
acha-bill: 👀

plugins/manaeventlogger/plugin.go
```diff
@@ -0,0 +1,126 @@
...
+		return
+	}
+	evs := make([]mana.Event, len(eventsBuffer))
+	copy(evs, eventsBuffer)
+	go func() {
+		if err := writeEventsToCSV(evs); err != nil {
+			log.Infof("error writing events to csv: %w", err)
+		}
+	}()
+	eventsBuffer = nil
```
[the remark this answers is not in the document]
acha-bill: nil slice and empty slice are practically the same. But nil slice is preferrable.
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

plugins/manaeventlogger/plugin.go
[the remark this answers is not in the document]
acha-bill: Lets leave this responsibility only with the ticker. Ticker clears buffer if it's full and on shutdown.

packages/mana/consensusbasevector.go
```diff
@@ -187,7 +187,7 @@ func (c *ConsensusBaseManaVector) GetManaMap() (NodeMap, error) {
 	defer c.Unlock()
 	res := make(map[identity.ID]float64)
 	for ID := range c.vector {
-		mana, err := c.getMana(ID)
+		mana, err := c.getMana(ID, false)
```
lzpap, line 190: `GetManaMap()` function in mana plugin also calls `ConsensusBaseManaVector.GetManaMap()`. Since higher level functions in the mana plugin all return current values, we should add ` update ...bool` to `ConsensusBaseManaVector.GetManaMap()` and propagate `update` to this `getMana` call, instead of passing `false`.
Therefore, the higher level `manaPlugin.GetManaMap()` can get the mana map with updated values, and `getPastConsensusManaVectorHandler` can get the non updated values.

lzpap (commented):
Looks really good now @acha-bill! When you add that small call argument, I can approve and we can merge ;)

plugins/webapi/mana/logsHandler.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, nodeID := range req.NodeIDs {
+		_nodeID, err := mana.IDFromStr(nodeID)
+		if err != nil {
+			return c.JSON(http.StatusBadRequest, GetEventLogsResponse{Error: err.Error()})
+		}
+		nodeIDs = append(nodeIDs, _nodeID)
+	}
+	startTime := time.Unix(req.StartTime, 0)
+	endTime := time.Unix(req.EndTime, 0)
+	var epoch time.Time
```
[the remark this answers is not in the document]
acha-bill: I don't think this small detail matters. I simply want the oldest time. `0001` or `1970` will do.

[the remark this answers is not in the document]
lzpap: It is not a small detail from API and usability point of view. Would you expect your email client to give you your emails only up to `1970` if you want to fetch all your mails?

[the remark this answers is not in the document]
lzpap: > So if you don't specify a start/end time, current implementation gives you back events until 1970-01-01 00:00:00 +0000 UTC.

Because this is what happens now with mana events...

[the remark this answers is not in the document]
acha-bill: I don't think it really matters though. Practically no event goes past 2020. :)

plugins/mana/plugin.go
```diff
@@ -371,8 +414,255 @@ func GetPendingMana(value float64, n time.Duration) float64 {
...
+	err = cbmvPast.UpdateAll(t)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return cbmvPast, eventLogs, nil
+}
+
+func checkConsensusEventStorage() {
+	size := consensusEventsLogStorage.GetSize()
```
lzpap, line 560: @acha-bill another bug/feature I found while testing is that `objectStorage.GetSize()` returns the size of the cache, not the size of the elements in cache + persistence. The cache content is persisted after couple of seconds, so  `size` will never be bigger, than `maxConsensusEventsInStorage`. The actual pruning of the events and constructing the past vector never happens.

acha-bill: Good catch. I'll load the size of the storage on start and increment it as events are being stored.
Sometime in the future we should have an `objectstorage.GetSize(skipCache ...bool)`

lzpap: submit a ticket in `hive` ;)

plugins/mana/parameters.go
```diff
@@ -31,4 +37,5 @@ func init() {
 	flag.Bool(CfgAllowedAccessFilterEnabled, false, "if filtering on access mana pledge nodes is enabled")
 	flag.Bool(CfgAllowedConsensusFilterEnabled, false, "if filtering on consensus mana pledge nodes is enabled")
 	flag.Bool(CfgManaEnableResearchVectors, false, "enable mana research vectors")
+	flag.Duration(CfgPruneConsensusEventLogsInterval, 5*time.Minute, "interval to check and prune consensus event storage")
```
lzpap, line 40: @acha-bill don't we stick to the counter based approach?

plugins/mana/plugin.go
```diff
@@ -47,6 +47,8 @@ var (
 	consensusBaseManaPastVectorStorage         *objectstorage.ObjectStorage
 	consensusBaseManaPastVectorMetadataStorage *objectstorage.ObjectStorage
 	consensusEventsLogStorage                  *objectstorage.ObjectStorage
+	consensusEventsLogsStorageSize             int64
+	consensusEventsLogsStorageSizeMu           sync.Mutex
```
lzpap, line 51: an atomic counter does the job, no need for mutex. Consider the `go.uber.org/atomic` package or the built in `sync/atomic` https://gobyexample.com/atomic-counters

plugins/mana/plugin.go
```diff
@@ -104,16 +107,22 @@ func configureEvents() {
 	mana.Events().Revoked.Attach(onRevokeEventClosure)
 }
 
+func incConsensusEventsLogsStorageSize() {
```
lzpap, line 110: this would just become `counter.Inc()`

plugins/mana/plugin.go
lzpap: also, probably we don't need `int64`, an `uint32` might be more than enough with couple hundred thousand max storage size.

plugins/mana/parameters.go
acha-bill: what counter approach?

plugins/mana/plugin.go
acha-bill: Totally forgot about atomic.

plugins/mana/parameters.go
lzpap: Yes. Pruning x% of the events from storage when we reach max size.

acha-bill: This is the interval to check the size and prune if necessary.

lzpap: @acha-bill if you only have the ticker there would be more events added in `tickerInterval` at a certain TPS than events that get removed. So log storage just grows. With current values, assuming best case scenario for tx input count, this would be 18 TPS. I'm not saying we should tweak the parameters, becasue this problem will arise at some TPS for sure becasue of the design. I'm saying to add the size check to `logPledge` and `logRevoke`, and trigger the pruning from there if the buffer gets full before the tick comes. Kind of the same flow as I suggested for the `csv` logger.

acha-bill: There's no point of ticking if we check the size every single time we save an event. If the buffer is full, the logger will always know first.

It only makes sense to go with 1 approach. 
The choice of approach simply depends on the TPS.

I thought about this and went with the timer coz it's more configurable.

lzpap: Alright, let's go with the timer approach.
Notes:
 - Decrement `consensusEventsLogsStorageSize` when doing `DeleteEntriesFromStore`.
 - We could add an adaptive window approach to prevent the growing of the storage. Downside being is that when there is high tps, your window to reconstruct the past vector will decrease. (unless you set greater storage allowance). Let's discuss this in a call.

acha-bill: >We could add an adaptive window approach to prevent the growing of the storage.

I think it's fine as is tbh. We can make more optimizations are we see fit later on.

plugins/mana/plugin.go
acha-bill: >also, probably we don't need int64, an uint32 might be more than enough with couple hundred thousand max storage size.

This is one of those things that just come down to preference. I don't see any real benefit of `uint` here.  If we were doing stuff like with bit shifts or overflow maths, it'll be useful. Furthermore, it impaires readability with all the castings here and there. 
A `length, size, capacity, index` should be an `int`.  Whether you want `int32` or `int64` depends on memory or intent.

plugins/mana/plugin.go
```diff
@@ -84,87 +105,141 @@ func configure(*node.Plugin) {
...
+	return atomic.LoadUint32((*uint32)(&c))
+}
+
+func (c countUint32) dec(v ...uint32) uint32 {
+	delta := uint32(1)
+	if len(v) > 0 {
+		delta = v[0]
+	}
+	return atomic.AddUint32((*uint32)(&c), ^delta-1)
+}
```
lzpap, line 131: if you used the `uber` package, you wouldn't need to reinvent the wheel

lzpap (approved):
Wohoo! 🎉🎉🎉

PR #913, opened 2021-01-11
Feat aPoW

capossele:


packages/tangle/filter.go
```diff
@@ -104,55 +105,78 @@ type PowFilter struct {
...
 	mu             sync.Mutex
-	acceptCallback func([]byte, *peer.Peer)
-	rejectCallback func([]byte, error, *peer.Peer)
+	acceptCallback func(*Message, *peer.Peer)
+	rejectCallback func(*Message, error, *peer.Peer)
 }
 
 // NewPowFilter creates a new PoW bytes filter.
 func NewPowFilter(worker *pow.Worker, difficulty int) *PowFilter {
+	pow.BaseDifficulty = difficulty
```
jkrvivian, line 118: Remove this line, since `pow.BaseDifficulty` is already set in `plugins/pow/pow.go`.

packages/tangle/filter_test.go
```diff
@@ -51,14 +78,102 @@ func TestPowFilter_Filter(t *testing.T) {
...
+	msg.nonce = nonce
+	msg.bytes = nil
+
+	// accpet since apow difficulty should be >= diffficulty
+	m.On("Accept", msg, testPeer)
+	filter.Filter(msg, testPeer)
+
+	// loop to force the number of zeros to be == testDifficulty
+	zeros := 0
+	for zeros != testDifficulty {
```
jkrvivian, line 112: `Mine` should output the valid nonce with the given `testDifficulty`, why do we need this loop?

capossele: Because mine could also output a nonce with a greater number of zeroes and we need to test the case of rejecting a nonce when the provided difficulty is not increased accordingly to the aPoW.

Thoralf-M:
Is it possible to abort the PoW if it takes to long and restart with new parents, so when the PoW is finally done, the referenced parents will be always very new, even if the node has a very slow cpu?

capossele:
> Is it possible to abort the PoW if it takes to long and restart with new parents, so when the PoW is finally done, the referenced parents will be always very new, even if the node has a very slow cpu?

That is a good suggestion. We should do the same also for the issuance timestamp because currently that is set before starting the aPoW

packages/pow/apow.go
```diff
@@ -0,0 +1,81 @@
+package pow
+
```
ubaistp, line 2: Hi

PR #924, opened 2021-01-19
Adds msg approval analysis api

### LINKED ISSUE #921: Number of tips per node and local knowledge of each node in Prometheus
Hi everyone,

I think it would be really interesting to have the following data in Prometheus.
- The number of tips belonging to each node (every 5s).
- The knowledge of each node regarding the number of tips of the neighboring nodes (every 5s).
- Also the time/average time for each node to see its own tips to be validated (every 5s).

I hope my suggestions are useful. 

Alexandre

capossele:
Fix #921

### DIFF
--- a/packages/tangle/analysis.go
+++ b/packages/tangle/analysis.go
@@ -112,9 +112,8 @@ func (t *Tangle) FirstApprovalAnalysis(nodeID string, filePath string) error {
 			if err := w.Error(); err != nil {
 				return err
 			}
-			return nil
 		}
-		return err
+		return nil
 	})
 }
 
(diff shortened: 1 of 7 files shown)


packages/tangle/analysis.go
```diff
@@ -0,0 +1,231 @@
...
+				processedMessages[approverID] = types.Void
+			}
+		})
+	}
+
+	return nil
+}
+
+func (t *Tangle) firstApprovers(msgID MessageID) ([]MsgInfo, error) {
+	approversInfo := []MsgInfo{}
```
lzpap, line 170: my linter complains about this
```suggestion
	approversInfo := make([]MsgInfo, 0)
```

lzpap (approved):
Looks good!
It would be insteresting to see at what tangle size it breaks, or becomes too hot to handle 🔥

PR #930, opened 2021-01-24
Add core logic for timestamp voting

capossele:
- Make `opinion` an independent package
- Add `Timestamp` and `Level of Knowledge` types


packages/tangle/messagemetadata.go
```diff
@@ -174,6 +184,35 @@ func (m *MessageMetadata) SetBooked(booked bool) (modified bool) {
...
+	m.timestampOpinionMutex.RLock()
+	defer m.timestampOpinionMutex.RUnlock()
+	return m.timestampOpinion
+}
+
+// SetTimestampOpinion sets the timestampOpinion flag.
+// It returns true if the timestampOpinion flag is modified. False otherwise.
+func (m *MessageMetadata) SetTimestampOpinion(timestampOpinion TimestampOpinion) (modified bool) {
+	m.timestampOpinionMutex.RLock()
+	if m.timestampOpinion != timestampOpinion {
```
jkrvivian, line 198: I think `Equal` should be used here.

packages/tangle/timestamp.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"time"
+
+	"github.com/iotaledger/goshimmer/packages/vote/opinion"
+	"github.com/iotaledger/hive.go/cerrors"
+	"github.com/iotaledger/hive.go/marshalutil"
+	"golang.org/x/xerrors"
+)
+
+// LevelOfKnowledge defines the Leve lOf Knowledge type.
```
jkrvivian, line 12:
```suggestion
// LevelOfKnowledge defines the Level Of Knowledge type.
```

PR #936, opened 2021-01-28
Feat/cleanup mana vector

### LINKED ISSUE #907: Clean up zero values from mana vectors
## Description

Both `Access` and `Consensus` mana vectors use exponential moving averages to arrive at the effective ("usable") mana value of a node. As a consequence, when base mana values become zero, the related effective mana values tend to zero with an exponentially decreasing function, but their value will never become exactly `0`.

 - Find a suitable value close to zero `minimumEffectiveMana` at which the effective value is considered practically zero, and the entry can be removed from the base mana vector.
 - Implement a background worker that periodically cleans up the vectors if their effective mana value drops below `minimumEffectiveMana`.

acha-bill:
Clean up empty mana nodes from the vectors.

Fixes #907


plugins/mana/plugin.go
```diff
@@ -678,6 +685,29 @@ func pruneConsensusEventLogsStorage() {
...
+				if baseMana.EffectiveValue() < minEffectiveMana && baseMana.BaseValue() < minBaseMana1 {
+					toRemove[vecType] = append(toRemove[vecType], id)
+				}
+			}
+			return true
+		})
+	}
+	for _, vecType := range types {
+		baseManaVectors[vecType].Remove(toRemove[vecType])
+	}
```
lzpap, line 708: @acha-bill let's make the removal atomic, becasue now it can happen that between determining which nodes to remove and actually removing them a pledge happens

acha-bill: done

lzpap (approved):

PR #938, opened 2021-01-29
feat: Introduce invalid to MessageMetadata

### LINKED ISSUE #931: Cleaner of messageStore
Should clean upon message invalid events. Be sure to keep enough info so that the recentlySeenBytes filter can still work.

TODO:
* Discuss how we're going to store the invalid message info after deletion:
    * **map**: similar to recentlySeenBytes, the old ones && the less frequently seen ones will be removed when the map is full. but need to process the invalid msg again if it's not in the list anymore.
    * **object storage**: the invalid msg will never be processed again.
* Discuss the data field of an invalid message to be stored, besides MessageID
* `Invalid` in MessageMetadata can be removed after resolving the previous bullet points.

jkrvivian:
# Description of change

Mark the message as invalid when:
* at least one of its parent is invalid
* at least one parent doesn't pass the parents age check

## Type of change
- Enhancement

## Change checklist

### DIFF
--- a/packages/tangle/messagemetadata.go
+++ b/packages/tangle/messagemetadata.go
@@ -220,7 +220,7 @@ func (m *MessageMetadata) SetEligible(eligible bool) (modified bool) {
 	m.eligibleMutex.Lock()
 	defer m.eligibleMutex.Unlock()
 
-	if m.eligble == eligible {
+	if m.eligible == eligible {
 		return false
 	}
 
--- a/packages/tangle/messagemetadata.go
+++ b/packages/tangle/messagemetadata.go
@@ -22,12 +22,14 @@ type MessageMetadata struct {
 	timestampOpinion   TimestampOpinion
 	booked             bool
 	eligible           bool
+	invalid            bool
 
 	solidMutex              sync.RWMutex
 	solidificationTimeMutex sync.RWMutex
 	timestampOpinionMutex   sync.RWMutex
 	bookedMutex             sync.RWMutex
 	eligibleMutex           sync.RWMutex
+	invalidMutex            sync.RWMutex
 }
 
 // NewMessageMetadata creates a new MessageMetadata from the specified messageID.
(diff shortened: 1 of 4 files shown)


capossele (changes requested):
I'm afraid that is not enough. Suppose you have a chain of messages building on top an invalid one. If you do not have a way to `remember` what message has been marked as invalid, then you'll try to solidify over and over again the deleted message. BTW, messages building on top an invalid one should also be considered as invalid.

packages/tangle/solidifier.go
```diff
@@ -197,6 +200,39 @@ func (t *Tangle) isAgeOfParentValid(childTime time.Time, parentID MessageID) boo
...
+
+	valid := true
+	msg.ForEachParent(func(parent Parent) {
+		valid = valid && t.isMessageValid(parent.ID)
+	})
+
+	return valid
+}
+
+// checks whether parent is valid.
```
capossele, line 217:
```suggestion
// checks whether the given message is valid.
```

capossele (approved):

packages/tangle/messagemetadata.go
```diff
@@ -236,6 +242,38 @@ func (m *MessageMetadata) SetEligible(eligible bool) (modified bool) {
...
+			modified = true
+		}
+		m.invalidMutex.Unlock()
+
+	} else {
+		m.invalidMutex.RUnlock()
+	}
+
+	return
+}
```
jonastheis, line 275: Please get rid of this ugly double locking business and add guard style. The code gets much harder to read and maintain even though it does such a simple thing. I agree that it might make sense to do this for heavy operations (e.g. computing a messageID) but setting a bool is definitely not one of them.

Something like this
```
lock()
defer unlock()

if m.invalid == invalid {
return false
}

m.invalid = invalid
m.SetModified()
return true

```

jkrvivian: Agree, I'll also refactor other bool setters.

packages/tangle/messagemetadata.go
```diff
@@ -216,23 +217,43 @@ func (m *MessageMetadata) SetTimestampOpinion(timestampOpinion TimestampOpinion)
...
-	} else {
-		m.eligibleMutex.RUnlock()
+	return
+}
+
+// IsInvalid returns true if the message represented by this metadata is invalid. False otherwise.
+func (m *MessageMetadata) IsInvalid() (result bool) {
+	m.invalidMutex.RLock()
+	result = m.invalid
+	m.invalidMutex.RUnlock()
```
acha-bill, line 238: `defer` is prefered.

acha-bill (approved):
Lgtm

PR #940, opened 2021-01-31
Markers integration

### LINKED ISSUE #939: Markers integration
## TODOs
- [x] integrate `StructureDetails` into `MessageMetadata`: We need to be careful to not release `MessageMetadata` while accompanying `StructureDetails` is still in use!
- [x] create `markers` into `messagelayer` plugin

## Markers in `messagelayer` plugin
- initialize `markers.Manager`
- define rule to create markers (e.g. for simplicity every 10 messages)

### Creating new Markers
- attach to `MessageBookedEvent`
- call `manager.InheritStructureDetails` for every message (possibly creating new marker, depending on closure function)
- if new marker is created
  - propagate/walk FM to past cone using `UpdateStructureDetails`

#### Creating new MSIs
This should be implicitly handled with the alias for the branchID in the `markers.Manager`. We need to gather a message parents' branches and pass it to the manager via alias.

### Testing
- simulate events according to specific scenario and check

jonastheis:
Closes #939

### DIFF
--- a/dapps/valuetransfers/packages/payload/payload.go
+++ b/dapps/valuetransfers/packages/payload/payload.go
@@ -184,7 +184,7 @@ func (p *Payload) String() string {
 // region Payload implementation ///////////////////////////////////////////////////////////////////////////////////////
 
 // Type represents the identifier which addresses the value Payload type.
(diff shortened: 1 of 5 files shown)


jonastheis:
This already has been manually integrated into the Booker component of the tangle with #946.

PR #941, opened 2021-02-01
feat: Avoid gossiping requested messages

### LINKED ISSUE #928: Avoid gossiping requested messages

jkrvivian:
# Description of change

* Add `requestedMsgs` to record requested messages
* Insert messages to `requestedMsgs` in `MissingMessageReceived` event
* Remove messages in `MessageSolid` and `MessageInvalid` event
* Add a check in `MessageSolid` event so that we won't gossip the requested message

Close #928 

## Type of change

- Enhancement

## Change checklist

### DIFF
--- a/plugins/gossip/plugin.go
+++ b/plugins/gossip/plugin.go
@@ -31,7 +31,7 @@ var (
 	ageThreshold            time.Duration
 	tipsBroadcasterInterval time.Duration
 
-	requestedMsgs map[tangle.MessageID]types.Empty
+	requestedMsgs     map[tangle.MessageID]types.Empty
 	requestedMsgMutex sync.Mutex
 )
 
@@ -143,7 +143,7 @@ func configureMessageLayer() {
 		}
 
 		msg := cachedMsgEvent.Message.Unwrap()
-		
+
 		// do not gossip requested messages
 		requestedMsgMutex.Lock()
 		if _, exist := requestedMsgs[msg.ID()]; exist {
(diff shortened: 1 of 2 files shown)


plugins/gossip/plugin.go
```diff
@@ -138,11 +143,40 @@ func configureMessageLayer() {
 		}
 
 		msg := cachedMsgEvent.Message.Unwrap()
+
+		// do not gossip requested messages
+		requestedMsgMutex.Lock()
+		if _, exist := requestedMsgs[msg.ID()]; exist {
+			delete(requestedMsgs, msg.ID())
```
capossele, line 150: Why do we delete here? Isn't going to then forget we've requested that msg?

plugins/gossip/plugin.go
```diff
@@ -138,11 +143,40 @@ func configureMessageLayer() {
...
+			return
```
capossele, line 151: You forgot to unlock the lock here. That is the problem of having to deal with the lock at this level. It would be less error prone if the requestedMsgs would be handling the lock internally.

capossele (changes requested):

plugins/gossip/plugin.go
jkrvivian: Yes, it is. Since the msg is solid and already stored, the msg shouldn't trigger `MessageSolid` event and be gossiped anymore, no?

capossele: You are right!

jonastheis (approved):

capossele (approved):

PR #944, opened 2021-02-01
New data flow

capossele:


packages/tangle/messagemetadata.go
```diff
@@ -126,6 +150,113 @@ func (m *MessageMetadata) SolidificationTime() time.Time {
...
+		if !m.timestampOpinion.Equal(timestampOpinion) {
+			m.timestampOpinion = timestampOpinion
+			m.SetModified()
+			modified = true
+		}
+		m.timestampOpinionMutex.Unlock()
+
+	} else {
+		m.timestampOpinionMutex.RUnlock()
+	}
```
jonastheis, line 212: Double locking 🤢

packages/tangle/messagefactory.go
```diff
@@ -90,6 +90,19 @@ func (f *MessageFactory) IssuePayload(p payload.Payload) (*Message, error) {
...
+	if t != nil {
+		for _, parent := range strongParents {
+			t[0].Message(parent).Consume(func(msg *Message) {
+				if msg.ID() != EmptyMessageID && !msg.IssuingTime().Before(issuingTime) {
+					time.Sleep(msg.IssuingTime().Sub(issuingTime) + 1*time.Nanosecond)
+					issuingTime = clock.SyncedTime()
+				}
+			})
+		}
+	}
```
jonastheis, line 104: This seems more like a quick hack. Not sure if necessary now but there could be a flag if it should wait for the time or otherwise return an error.

packages/tangle/messagestore.go
```diff
@@ -0,0 +1,254 @@
...
+			MessageMetadata: cachedMsgMetadata})
+	}
+
+	// messages are stored, trigger MessageStored event to move on next check
+	m.Events.MessageStored.Trigger(&CachedMessageEvent{
+		Message:         cachedMessage,
+		MessageMetadata: cachedMsgMetadata})
+
+	cachedMessage.Release()
+	cachedMsgMetadata.Release()
```
jonastheis, line 100: Just for my understanding: Releasing it here potentially means that the message might be in cache for the full data flow?

packages/tangle/messagemetadata.go
lzpap: how exactly?

jonastheis: I mean this rlock and lock business. It should just be something like 
```
func (m *MessageMetadata) SetBooked(booked bool) (modified bool) {
	m.bookedMutex.Lock()
	defer m.bookedMutex.Unlock()

	if m.booked == booked {
		return false
	}

	m.booked = booked
	m.SetModified()
	modified = true

	return
}```

Reads much easier :)

lzpap: true, but then this has to acquire the write lock even if the value is not modified. I can live with that :D

tactical-drone: Between line 200 and 202 another thread could have acquired the read lock also entered the condition and raced to win the write lock?

jonastheis: It could, that's why there's another if check when the write lock is acquired in line 203 to avoid a race condition. Another reason why it's so ugly :D

tactical-drone: I see, there is another check. No bug.

lzpap (approved):

PR #947, opened 2021-02-05
refactor: attachments

acha-bill:
Refactor: Attachments 
Mostly copy paste.😂😂

### DIFF
--- a/packages/tangle/attachment.go
+++ b/packages/tangle/attachment.go
@@ -1,0 +1,161 @@
+package tangle
+
+import (
+	"fmt"
+
+	"github.com/iotaledger/goshimmer/packages/ledgerstate"
+	"github.com/iotaledger/hive.go/byteutils"
+	"github.com/iotaledger/hive.go/marshalutil"
+	"github.com/iotaledger/hive.go/objectstorage"
+	"github.com/iotaledger/hive.go/stringify"
+)
+
+// Attachment stores the information which transaction was attached by which message. We need this to be able to perform
+// reverse lookups from transactions to their corresponding messages that attach them.
+type Attachment struct {
+	objectstorage.StorableObjectFlags
+
+	transactionID ledgerstate.TransactionID
+	messageID     MessageID
+}
+
+// NewAttachment creates an attachment object with the given information.
+func NewAttachment(transactionID ledgerstate.TransactionID, messageID MessageID) *Attachment {
+	return &Attachment{
+		transactionID: transactionID,
+		messageID:     messageID,
+	}
+}
+
+// AttachmentFromBytes unmarshals an Attachment from a sequence of bytes - it either creates a new object or fills the
+// optionally provided one with the parsed information.
+func AttachmentFromBytes(bytes []byte) (result *Attachment, consumedBytes int, err error) {
+	marshalUtil := marshalutil.New(bytes)
+	result, err = ParseAttachment(marshalUtil)
+	consumedBytes = marshalUtil.ReadOffset()
+
+	return
+}
+
+// ParseAttachment is a wrapper for simplified unmarshaling of Attachments from a byte stream using the marshalUtil package.
+func ParseAttachment(marshalUtil *marshalutil.MarshalUtil) (result *Attachment, err error) {
+	result = &Attachment{}
+	if result.transactionID, err = ledgerstate.TransactionIDFromMarshalUtil(marshalUtil); err != nil {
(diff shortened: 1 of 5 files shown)


packages/tangle/attachment.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return
+}
+
+// AttachmentFromObjectStorage gets called when we restore an Attachment from the storage - it parses the key bytes and
+// returns the new object.
+func AttachmentFromObjectStorage(key []byte, data []byte) (result objectstorage.StorableObject, err error) {
+	result, _, err = AttachmentFromBytes(byteutils.ConcatBytes(key, data))
+	if err != nil {
+		err = fmt.Errorf("failed to parse attachment from object storage: %w", err)
```
jonastheis, line 60: Please use `xerrors` to wrap errors or sentinel errors if applicable

packages/tangle/attachment.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return
+}
+
+// TransactionID returns the transactionID of this Attachment.
+func (attachment *Attachment) TransactionID() ledgerstate.TransactionID {
```
jonastheis, line 67:
```suggestion
func (a *Attachment) TransactionID() ledgerstate.TransactionID {
```

Only use the first letter as a receiver name everywhere.

packages/tangle/attachment.go
```diff
@@ -0,0 +1,161 @@
...
+		return
+	}
+
+	return
+}
+
+// AttachmentFromObjectStorage gets called when we restore an Attachment from the storage - it parses the key bytes and
+// returns the new object.
+func AttachmentFromObjectStorage(key []byte, data []byte) (result objectstorage.StorableObject, err error) {
+	result, _, err = AttachmentFromBytes(byteutils.ConcatBytes(key, data))
```
jonastheis, line 58: Since `data` is always empty the concat shouldn't be necessary.

packages/tangle/messagestore.go
```diff
@@ -155,6 +161,25 @@ func (m *MessageStore) MissingMessages() (ids []MessageID) {
...
+	attachment, stored := m.attachmentStorage.StoreIfAbsent(NewAttachment(transactionID, messageID))
+	if !stored {
+		return
+	}
+	cachedAttachment = &CachedAttachment{CachedObject: attachment}
+	return
+}
+
+// Attachments retrieves the attachment of a transaction in attachmentStorage.
+func (m *MessageStore) Attachments(transactionID ledgerstate.TransactionID) (cachedAttachments CachedAttachments) {
```
jonastheis, line 175: Since the attachments are immutable anyway we could always simply hand out a list of `Attachments` (type would need to be introduced) instead of cached objects. Makes it a bit easier to handle.

packages/tangle/messagebooker.go
```diff
@@ -183,7 +185,12 @@ func (m *MessageBooker) referencedTransactionIDs(transaction *ledgerstate.Transa
...
 }
 
+// Attachments retrieves the attachments of a transaction.
 func (m *MessageBooker) Attachments(transactionID ledgerstate.TransactionID) (attachments MessageIDs) {
+	cachedAttachments := m.messageStore.Attachments(transactionID)
+	cachedAttachments.Consume(func(attachment *Attachment) {
+		attachments = append(attachments, attachment.MessageID())
+	})
 	return
 }
```
jonastheis, line 195: I think this could be part of `MessageStore`. It could be a convenience function `AttachmentsMessageIDs` or something :)

jonastheis (changes requested):

packages/tangle/messagestore.go
acha-bill: I think It should follow the general pattern of the` messagestore` API. Methods return `cachedObjects`.

packages/tangle/messagebooker.go
acha-bill: uhmm, I prefer `messageStore` to be lean. But it doesn't matter if we want this convenience function tbh.

packages/tangle/messagestore.go
jonastheis: Fair enough. We talked about simplifying it yesterday in a call with Hans and Angelo. Maybe this is something we can do later as a part of a bigger refactor.

jonastheis (approved):

PR #949, opened 2021-02-05
Transaction filter

acha-bill:


packages/tangle/filter.go
```diff
@@ -234,3 +244,67 @@ func (f *RecentlySeenBytesFilter) getRejectCallback() (result func(bytes []byte,
...
+			return
+		}
+	}
+	f.getAcceptCallback()(msg, peer)
+}
+
+func isMessageAndTransactionTimestampsValid(transaction *ledgerstate.Transaction, message *Message) bool {
+	transactionTimestamp := transaction.Essence().Timestamp()
+	messageTimestamp := message.IssuingTime()
+	return transactionTimestamp.Before(messageTimestamp) && transactionTimestamp.Sub(messageTimestamp) <= MaxReattachmentTimeMin
```
capossele, line 281: 2 things here:
- the first part of the check  `transactionTimestamp.Before(messageTimestamp)` will return false if the times are equal. Unfortunately, we recently realised that on Windows the precision of time.Now() is in the order of millisecond, thus it could happen, especially during testing, that those times are the same.
- the second part of the check has a problem: imagine that transactionTimestamp is 0, and messageTimestamp is 100 => than you have 0-100 = -100 <= MaxReattachmentTime

capossele (changes requested):

packages/tangle/filter.go
capossele: I guess that having a unit test for these kind of functions always help

packages/tangle/filter_test.go
```diff
@@ -19,11 +25,56 @@ var (
...
+}
+
+func Test_isMessageAndTransactionTimestampsValid(t *testing.T) {
+	msg := &Message{}
+	t.Run("older tx timestamp within limit", func(t *testing.T) {
+		tx := newTransaction(time.Now())
+		msg.issuingTime = tx.Essence().Timestamp().Add(1 * time.Second)
+		assert.True(t, isMessageAndTransactionTimestampsValid(tx, msg))
+	})
+	t.Run("older timestamp but older  han max", func(t *testing.T) {
```
capossele, line 56:
```suggestion
	t.Run("older timestamp but older than max", func(t *testing.T) {
```

capossele (approved):
✨ Sparkling!

PR #950, opened 2021-02-07
(dummy) Scheduler implementation

capossele:
This PR adds the (dummy) scheduler implementation. However, the scheduler must still be wired into the Tangle flow.


packages/tangle/scheduler.go
```diff
@@ -0,0 +1,299 @@
...
+					s.timeQueue.Add(message)
+					break
+				}
+				s.trySchedule(message)
+
+			// try to schedule messsage that have been waiting and are now current.
+			case message := <-s.timeQueue.C:
+				s.trySchedule(message)
+
+			// schedule messages that were waiting for of their parents to be booked.
```
jkrvivian, line 101:
```suggestion
			// schedule messages that were waiting for their parents to be booked.
```

packages/tangle/scheduler.go
```diff
@@ -0,0 +1,298 @@
...
+
+// Scheduler implements the scheduler.
+type Scheduler struct {
+	Events *SchedulerEvents
+
+	tangle *Tangle
+
+	inbox            chan *MessageToSchedule
+	timeQueue        *TimeMessageQueue
+	parentsMap       SchedulerParentPriorityMap
```
lzpap, line 37: should this be protected with a mutex?
we check on incoming message of parents have been booked, but the multithreaded outbox worker could modifiy the map in the meantime, no?

packages/tangle/scheduler.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	return parents
+}
+
+// region TimeMessageQueue /////////////////////////////////////////////////////////////////////////////////////////////
+
+// TimeMessageQueue is a time-based ordered queue.
+type TimeMessageQueue struct {
+	list    timeIssuanceSortedList
```
lzpap, line 175: what about this list? should this be protected as well?

packages/tangle/tangle.go
```diff
@@ -44,10 +45,12 @@ func New(options ...Option) (tangle *Tangle) {
 	tangle.Parser = NewParser()
 	tangle.Storage = NewStorage(tangle)
 	tangle.Solidifier = NewSolidifier(tangle)
+	tangle.Scheduler = NewScheduler(tangle)
+	tangle.LedgerState = NewLedgerState(tangle)
+	tangle.Booker = NewBooker(tangle)
```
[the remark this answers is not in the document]
capossele: Yeah, I think the LesgerState could also be not initialised, while instead the Booker initialisation is necessary as the scheduler attach to one of its events. However, is the setup phase that will do the proper wiring

packages/tangle/scheduler.go
capossele: No that map is managed entirely by the scheduler

packages/tangle/scheduler.go
capossele: Also managed only by the scheduler

capossele: I mean, it is internally used by the TimeMessageQueue, which is managed by the scheduler and its access is restricted via a select

lzpap (approved):

PR #955, opened 2021-02-11
Add the OpinionFormer component

capossele:


packages/tangle/fcob.go
```diff
@@ -0,0 +1,294 @@
...
+					return
+				}
+				opinion.SetLevelOfKnowledge(Two)
+				// trigger OpinionPayloadFormed
+				messageIDs := f.tangle.Storage.AttachmentMessageIDs(transactionID)
+				for _, messageID := range messageIDs {
+					f.Events.PayloadOpinionFormed.Trigger(&OpinionFormedEvent{messageID, opinion.liked})
+				}
+			})
+		}, timestamp.Add(LocallyFinalizedThreshold))
```
jonastheis, line 193: I don't know how this executor thing works but if the first function gets executed after 5 seconds then you execute the other one after a total of 15 seconds

capossele: no because you set wrt to the original timestamp. So the first one would be executed at timestamp + LikedThreshold and the second at timestamp + LocallyFinalizedThreshold

jonastheis (approved):

PR #956, opened 2021-02-11
Update hive with the improved autopeering

capossele:


jonastheis (approved):
Wow impressive changes!

PR #959, opened 2021-02-11
Setup flow up to scheduler

capossele:


packages/tangle/tangle.go
```diff
@@ -162,4 +169,11 @@ func Identity(identity *identity.LocalIdentity) Option {
 	}
 }
 
+// WithoutOpinionFormer an Option for the Tangle that allows to diable the OpinionFormer component.
```
lzpap, line 172: typo disable

packages/tangle/tangle.go
```diff
@@ -71,7 +72,10 @@ func (t *Tangle) Setup() {
 	t.Requester.Setup()
 	t.TipManager.Setup()
 	t.Scheduler.Setup()
-	t.OpinionFormer.Setup()
+	if !t.Options.WithoutOpinionFormer {
+		t.OpinionFormer.Setup()
+		return
+	}
```
lzpap, line 78: I would leave comment here: "Booker and LederState setup is left out until the old value tangle is in use."

lzpap (approved):

PR #963, opened 2021-02-12
Fix wrong handler in the MessageInvalidEvent

capossele:
Thanks to Xee Vee

### DIFF
--- a/plugins/gossip/plugin.go
+++ b/plugins/gossip/plugin.go
@@ -155,5 +155,5 @@ func configureMessageLayer() {
 	messagelayer.Tangle().Storage.Events.MissingMessageStored.Attach(events.NewClosure(requestedMsgs.append))
 
 	// delete the message from requestedMsgs if it's invalid, otherwise it will always be in the list and never get removed in some cases.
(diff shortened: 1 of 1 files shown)


plugins/gossip/plugin.go
```diff
@@ -155,5 +155,5 @@ func configureMessageLayer() {
 	messagelayer.Tangle().Storage.Events.MissingMessageStored.Attach(events.NewClosure(requestedMsgs.append))
 
 	// delete the message from requestedMsgs if it's invalid, otherwise it will always be in the list and never get removed in some cases.
-	messagelayer.Tangle().Events.MessageInvalid.Attach(events.NewClosure(requestedMsgs.delete))
+	messagelayer.Tangle().Events.MessageInvalid.Attach(events.NewClosure(func(messageID tangle.MessageID) { requestedMsgs.delete(messageID) }))
```
lzpap, line 158: what about the lock on `requestedMsgs`?

capossele: Isn't the delete method already taking care of it?

lzpap (approved):

PR #969, opened 2021-02-16
Add new flow unit tests

capossele:
- Fix several issues
- Add Scenario 1, 2, 3 tests for the Booker
- Add Scenario 2 for the opinionFormer
- Refactor Test Tangle flow


packages/tangle/booker_scenario_2_test.go
```diff
@@ -0,0 +1,348 @@
+package tangle
```
jkrvivian, line 1: The PastMarkers and FutureMarkers of message 5 in scenario 2 figure are incorrect.

The PastMarkers should be 2,3, and FutureMarkers should be 2,4.

packages/tangle/booker.go
```diff
@@ -119,30 +163,72 @@ func (m *Booker) allTransactionsApprovedByMessage(transactionIDs ledgerstate.Tra
...
+			FutureMarkers: markers.NewMarkers(),
+		},
+		timestampOpinion: TimestampOpinion{
+			Value: opinion.Like,
+			LoK:   Three,
+		},
+		booked:   true,
+		eligible: true,
+	}
+	b.tangle.Storage.messageMetadataStorage.Store(genesisMetadata).Release()
```
hmoog, line 229: This will panic if the node restarts since "Store" overwrites the previous entry which is disabled for the models in goshimmer.

Maybe better use:

```suggestion
	b.tangle.Storage.MessageMetadata(EmptyMessageID, func() *MessageMetadata {
		return genesisMetadata
	}).Release()
```

hmoog (changes requested):

packages/tangle/booker.go
```diff
@@ -119,30 +163,74 @@ func (m *Booker) allTransactionsApprovedByMessage(transactionIDs ledgerstate.Tra
...
+
+		}) {
+			panic(fmt.Errorf("failed to load MessageMetadata with %s", parentMessageID))
+		}
+	})
+
 	return
 }
 
+func (b *Booker) createGenesis() {
```
jonastheis, line 211: Is the booker the right place for this?

jonastheis (approved):

PR #971, opened 2021-02-17
The Value Tangle is dead, long live the Tangle!

capossele:


client/wallet/unspentoutputmanager.go
```diff
@@ -11,15 +11,15 @@ import (
 type UnspentOutputManager struct {
 	addressManager *AddressManager
 	connector      Connector
-	unspentOutputs map[walletaddr.Address]map[transaction.ID]*Output
+	unspentOutputs map[string]map[ledgerstate.TransactionID]*Output
```
lzpap, line 14: why is the `walletaddr.Address` replaced with a string?

acha-bill: Coz `walletaddr.Address` wraps a pointer that does not work well as map keys.

```go
seed := walletseed.NewSeed(bytes)
m := make(map[walletaddr.Address]int)
m[seed.Address(0)]++
m[seed.Address(0)]++
assert.Equal(t, 2, m[seed.Address(0)]) // FAIL
```

jonastheis (approved):

lzpap (approved):

PR #1010, opened 2021-03-01
Documentation webapi

apenzk:
# Description of change

Added documentation for WebAPI

## Type of change

- Documentation 

## Change checklist


docs/tooling/webAPI_clientLib.md
```diff
@@ -1 +1,69 @@
...
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, Response{Error: err.Error()})
+	}
+	return c.JSON(http.StatusOK, Response{ID: msg.ID().String()})
+}
+```
+As an example the JSON body   
+```
+{
+	"data":"HelloWor"
```
capossele, line 64:
```suggestion
	"data":"HelloIOTA"
```

docs/tooling/webAPI_clientLib.md
```diff
@@ -1 +1,69 @@
...
+}
+```
+can be sent to `http://127.0.0.1:8080/data`, which will issue a data message containing "HelloWor" (note that in this  example the data input is size limited.)
```
capossele, line 67:
```suggestion
can be sent to `http://127.0.0.1:8080/data`, which will issue a data message containing "HelloIOTA" (note that in this example the data input is size limited.)
```

docs/tooling/webAPI_clientLib.md
```diff
@@ -1 +1,69 @@
...
+```
+We can then use the Context to send a JSON response to the node: 
+```
+JSON(statuscode int, i interface{}) error
+```
+An implementation example is shown later for the POST method.
+
+## GET and POST 
+
+Two methods are currently employed. First, with `GET` we register a new GET route for a handler function. The handler is accessed via the address `path`. The handler for a GET method can set the node to perform certain actions.
```
capossele, line 28:
```suggestion
Two methods are currently used. First, with `GET` we register a new GET route for a handler function. The handler is accessed via the address `path`. The handler for a GET method can set the node to perform certain actions.
```

capossele (approved):

docs/tooling/webAPI_clientLib.md
apenzk: This does not work. The string cannot currently be chosen arbitrarily: 
{
  "error": "code=400, message=illegal base64 data at input byte 8"
}

capossele: Yeah I forgot that data is supposed to be base64..  Nevermind then

PR #1021, opened 2021-03-02
set cache time to 0

capossele:


jonastheis (commented):
Setting the cache time to 0 might not be enough, as we saw when we tested it. Maybe we need to add small delays?

Also you're not resetting the cache time, does that mean that it will apply for any test executed after the flow test as well? Is that something we want?

capossele:
> Setting the cache time to 0 might not be enough, as we saw when we tested it. Maybe we need to add small delays?
> 
> Also you're not resetting the cache time, does that mean that it will apply for any test executed after the flow test as well? Is that something we want?

We process 20000 messages here, adding a delay might be problematic. For the cache time, if other tests use it with 0, it's only a good thing!

PR #1025, opened 2021-03-03
update integration tests with mana changes

acha-bill:
Noteworthy change: 
Made scheduler shutdown before opinion former.

I was seeing an issue where the scheduler will hang on shutdown coz the opinion former has already shut down and left the scheduler with *opinionless* messages.

### DIFF
--- a/tools/integration-tests/tester/tests/testutil.go
+++ b/tools/integration-tests/tester/tests/testutil.go
@@ -280,7 +280,7 @@ func SendIotaTransaction(t *testing.T, from *framework.Peer, to *framework.Peer,
 // SendColoredTransaction sends IOTA and colored tokens from and to a given peer and returns the ok flag and transaction ID.
 // 1. Get the first unspent outputs of `from`
 // 2. Send 50 IOTA and 50 ColorMint to `to`
-func SendColoredTransaction(t *testing.T, from *framework.Peer, to *framework.Peer, addrBalance map[string]map[ledgerstate.Color]int64, txConfig TransactionConfig) (ok bool, txId string) {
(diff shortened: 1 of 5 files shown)


tools/integration-tests/tester/tests/testutil.go
```diff
@@ -258,13 +260,16 @@ func SendIotaTransaction(t *testing.T, from *framework.Peer, to *framework.Peer,
...
 	txEssence := ledgerstate.NewTransactionEssence(0, time.Now(), txConfig.AccessManaPledgeID, txConfig.ConsensusManaPledgeID, ledgerstate.NewInputs(input), ledgerstate.NewOutputs(outputs...))
-	sig := ledgerstate.NewED25519Signature(from.KeyPair(0).PublicKey, from.KeyPair(0).PrivateKey.Sign(txEssence.Bytes()))
+	sig := ledgerstate.NewED25519Signature(from.KeyPair(txConfig.FromAddressIndex).PublicKey, from.KeyPair(txConfig.FromAddressIndex).PrivateKey.Sign(txEssence.Bytes()))
 	unlockBlock := ledgerstate.NewSignatureUnlockBlock(sig)
 	txn := ledgerstate.NewTransaction(txEssence, ledgerstate.UnlockBlocks{unlockBlock})
 
 	// send transaction
 	txId, err = from.SendTransaction(txn.Bytes())
-	require.NoErrorf(t, err, "could not send transaction on %s", from.String())
+	if err != nil {
```
lzpap, line 269: why change the structure from `require`? why return true if tx could not be sent?

acha-bill: Coz I don't want the test run to terminate if it fails to send a transaction. There's a check for the failure late on.

lzpap: where is this check and how do you know there was a problem if you return no problem?

lzpap (approved):

PR #1028, opened 2021-03-03
same timestamp for mana maps

acha-bill:
Methods returning mana maps or lists should use the same timestamp to `getMana` for each node.
- GetManaMap
- GetHighestManaNodes

### DIFF
--- a/packages/mana/accessbasevector.go
+++ b/packages/mana/accessbasevector.go
@@ -92,19 +92,17 @@ func (a *AccessBaseManaVector) GetMana(nodeID identity.ID, t ...time.Time) (floa
 }
 
 // GetManaMap returns mana perception of the node..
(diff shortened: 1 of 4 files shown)


packages/mana/accessbasevector.go
```diff
@@ -95,10 +95,16 @@ func (a *AccessBaseManaVector) GetMana(nodeID identity.ID, t ...time.Time) (floa
...
+	if len(timestamp) > 0 {
+		if timestamp[0].Before(t) {
+			t = timestamp[0]
+		}
+	}
 	res = make(map[identity.ID]float64)
 	for ID := range a.vector {
 		var mana float64
-		mana, t, err = a.getMana(ID, timestamp...)
+		mana, t, err = a.getMana(ID, updateAt)
```
lzpap, line 107: probably we don't need `t` here, or just check that the returned time is the same as `t`. Why update the returned `t` for every getMana call?

lzpap (approved):

PR #1030, opened 2021-03-04
feat: Visualizer tweaks

jkrvivian:
# Description of change

* Preserve the information of clicked msg on the visualizer
* Update the color of messages when it's confirmed (listen to `MessageOpinionFormed` event)
   * tx confirmed -> yellow
   * non-tx confirmed -> purple
   * pending / tx not confirmed -> gray 

image

## Type of change

Enhancement

## Change checklist

### DIFF
--- a/plugins/dashboard/frontend/src/app/components/Visualizer.tsx
+++ b/plugins/dashboard/frontend/src/app/components/Visualizer.tsx
@@ -60,7 +60,7 @@ export class Visualizer extends React.Component<Props, any> {
 
     render() {
         let {
-            vertices, solid_count, selected,
+            vertices, confirmed_count, selected,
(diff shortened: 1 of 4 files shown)


capossele (commented):
One comment I have for this PR is:
now that we only visualise a new message after it becomes a TIP, we will not have a visualisation during sync. Is it something we are ok with?

jkrvivian:
> One comment I have for this PR is:
> now that we only visualise a new message after it becomes a TIP, we will not have a visualisation during sync. Is it something we are ok with?

Messages are visualized once it's stored (MessageStored event). It's visualized during syncing, but remains gray (pending) until it gets confirmed/finalized.

capossele (approved):

PR #1036, opened 2021-03-05
Mana Sync Behavior

### LINKED ISSUE #906: Mana Behavior While Syncing
## Description

Functions in the mana plugin that read and return the mana vector state (`GetHighestManaNodes()`,  `GetManaMap()`, etc.) are also updating the `LastUpdated` fields of the vector entries to access time.
While syncing, the mana vectors are in intermediary state, their value corresponds to the amount of past transactions that have been already processed. `LastUpdated` fields of mana vector entries therefore should not be updated to current access time until the sync is complete by aforementioned mana plugin functions.

## Suggestion
Forbid accessing (reading) mana vectors until the node is in sync.
 - Return an error "Node not in sync" from mana plugin functions that might update `LastUpdated` fields.
 - This effectively means, that the `metrics` plugin is only able to extract mana related information once the sync is complete.
 - Introduce a config parameter `EnableManaDebugging` that lets us switch off this behavior, while making sure, that the returned mana values corresponds to the latest `pledge/revoke` event rather than current access time. If until the sync is done, only pledges or revokes can modify the `LastUpdated` fields, it is easy to find the entry for the last event by sorting the `LastUpdated` fields. Then update all other entries up to this point.

lzpap:
Fixes #906
# Description
 Define mana plugin behavior while syncing:

 - Unless `mana.debuggingEnabled` is explicitly enabled in config, the mana plugin replies to all queries with zero value response and `ErrQueryNotAllowed` if the node is not in sync according to the syncbeacon.
 - This does not effect updating the mana vectors, only signals to the caller, that mana values don't make sense until the node is in synced state.
 - Display warning message on the mana page of the dashboard if node is not synced.

### DIFF
--- a/plugins/mana/plugin.go
+++ b/plugins/mana/plugin.go
@@ -18,6 +18,7 @@ import (
 	"github.com/iotaledger/goshimmer/plugins/database"
 	"github.com/iotaledger/goshimmer/plugins/gossip"
 	"github.com/iotaledger/goshimmer/plugins/messagelayer"
+	"github.com/iotaledger/goshimmer/plugins/syncbeaconfollower"
 	"github.com/iotaledger/hive.go/daemon"
 	"github.com/iotaledger/hive.go/datastructure/set"
 	"github.com/iotaledger/hive.go/events"
@@ -52,6 +53,7 @@ var (
 	onTransactionConfirmedClosure              *events.Closure
 	onPledgeEventClosure                       *events.Closure
 	onRevokeEventClosure                       *events.Closure
+	debuggingEnabled                           bool
 )
 
 // Plugin gets the plugin instance.
@@ -97,6 +99,8 @@ func configure(*node.Plugin) {
 	if err != nil {
 		log.Panic(err.Error())
 	}
+
+	debuggingEnabled = config.Node().Bool(CfgDebuggingEnabled)
 
 	configureEvents()
 }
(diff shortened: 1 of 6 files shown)


plugins/mana/plugin.go
```diff
@@ -281,27 +285,42 @@ func shutdownStorages() {
 // It also updates the mana values for each node.
 // If n is zero, it returns all nodes.
 func GetHighestManaNodes(manaType mana.Type, n uint) ([]mana.Node, time.Time, error) {
+	if !QueryAllowed() {
+		return []mana.Node{}, time.Now(), ErrQueryNotAllowed
```
capossele, line 289: Should we use `clock.SyncedTime()` rather than time.Now()? 
As a side note, should we do the same for whenever we use time within Mana?

acha-bill: I don't think so. The time here is the node's own time.

lzpap: we could

capossele: > I don't think so. The time here is the node's own time.

but the node's own time should be `clock.SyncedTime()`

capossele: anyhow, I guess that with this PR is ok since the content is empty. Then, we re-evaluate the time usage in the rest of mana

capossele (approved):

PR #1059, opened 2021-03-12
fix FCOB

capossele:
Implement FCOB*


hmoog (approved):
I didn't review it but Angelo said it works :D

Issue #890, opened 2020-12-15
Enable caching for integration tests


jonastheis:
If we enable [caching](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows) via any of [these options](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei) we should be able to speed up at least the pulling of dependencies and building by a large margin. 

This already takes a little bit longer than 3min.

![image](https://user-images.githubusercontent.com/4181434/102218229-92fde100-3ed5-11eb-888a-3adf7ab18c69.png)

georgysavva:
It seems that the article that you mentioned is a bit old. 
Here are the official docket GH actions to achieve image layers caching via Github caching:
https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md
And complete third-party article as a bonus: 
https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching
I will try to pull this off.

georgysavva:
@capossele Why did we close this issue?

capossele:
My bad, I thought we were done with this. I'll reopen it

georgysavva:
After running several benchmarks it turned out that docker layer caching doesn't give any performance benefit due to the fact that we can not cache the Go build. So we should get back this when they fix that in moby/buildkit#1512 
@jonastheis

PR #1074, opened 2021-03-16
Make docker build faster

georgysavva:
Docker build now mounts a volume for Go build-cache during the image build. Another small optimization is: mount the codebase instead of COPYing it.
As a result, the docker re-build time is 4.5s instead of the 50s before on my laptop

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-# syntax = docker/dockerfile@sha256:6ce32b7af840de740a21ba297f7d1c31e0f1eb0d6b03c0efc03ae0cca9c8fd2a
-# Use Docker fronted docker/dockerfile:1.2.1 to enable BuildKit features.
+# syntax = docker/dockerfile:1.2.1
 
 ############################
 # Build
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,13 +20,12 @@ ENV GO111MODULE=on
 RUN go mod download
 RUN go mod verify
 
-COPY . .
-
 # 1. Mount everything from the current directory to the PWD(Present Working Directory) inside the container
 # 2. Mount the build cache volume
 # 3. Build the binary
 # 4. Verify that goshimmer binary is statically linked
-RUN --mount=type=cache,target=/root/.cache/go-build \
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
     CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
     -ldflags='-w -s -extldflags "-static"' \
     -o /go/bin/goshimmer; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-# syntax = docker/dockerfile:1.2
+# syntax = docker/dockerfile@sha256:6ce32b7af840de740a21ba297f7d1c31e0f1eb0d6b03c0efc03ae0cca9c8fd2a
+# Use Docker fronted docker/dockerfile:1.2.1 to enable BuildKit features.
+
 ############################
 # Build
 ############################
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,13 +19,19 @@ ENV GO111MODULE=on
 RUN go mod download
 RUN go mod verify
 
-# Copy everything from the current directory to the PWD(Present Working Directory) inside the container
-COPY . .
+# 1. Mount everything from the current directory to the PWD(Present Working Directory) inside the container
+# 2. Mount the build cache volume
+# 3. Build the binary
+# 4. Verify that goshimmer binary is statically linked
+RUN --mount=target=. \
 
-# Build the binary
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
-      -ldflags='-w -s -extldflags "-static"' -a \
-       -o /go/bin/goshimmer
+    --mount=type=cache,target=/root/.cache/go-build \
+
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+    -ldflags='-w -s -extldflags "-static"' \
+    -o /go/bin/goshimmer; \
+
+    ./check_static.sh
 
 ############################
 # Image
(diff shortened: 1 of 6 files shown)


georgysavva:
I still need to figure out why it doesn't work in the Github Actions.

Dockerfile
```diff
@@ -19,13 +20,16 @@ ENV GO111MODULE=on
...
-      -ldflags='-w -s -extldflags "-static"' -a \
-       -o /go/bin/goshimmer
+# 1. Mount everything from the current directory to the PWD(Present Working Directory) inside the container
+# 2. Mount the build cache volume
+# 3. Build the binary
+# 4. Verify that goshimmer binary is statically linked
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+    -ldflags='-w -s -extldflags "-static"' \
```
jonastheis, line 30: I see you removed the `-a` (force rebuilding of packages that are already up-to-date.) flag. Was there any specific reason we had this in first place? @Wollac knows maybe?

Dockerfile
```diff
@@ -19,13 +20,16 @@ ENV GO111MODULE=on
...
-# Build the binary
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
-      -ldflags='-w -s -extldflags "-static"' -a \
-       -o /go/bin/goshimmer
+# 1. Mount everything from the current directory to the PWD(Present Working Directory) inside the container
+# 2. Mount the build cache volume
+# 3. Build the binary
+# 4. Verify that goshimmer binary is statically linked
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
```
jonastheis, line 28: This should significantly speed up local builds and thus make development for integration tests much less painful. But we won't be able to leverage the cache on GH Actions since this is not persisted anywhere and we'd need to share it across builds as suggested as a "hack" [here](https://stackoverflow.com/a/60485339/3346202) See this issue https://github.com/moby/buildkit/issues/1512. 
Another option would be using one of the techniques described in #890. 

But in general I think this is already great and we should maybe first implement static peering to speed up overall execution of the integration tests before investing too much time into CI optimizations/caching

jonastheis (commented):
Maybe we should also enable BuildKit for the images we push to DockerHub on develop and for a release?

Dockerfile
georgysavva: The only reason that I see is to always rebuild everything including the `std` to be more sure that the resulting binary is statically linked. But build-cache doesn't work in that case, so I disabled it and added an additional check after the build that the resulting binary is indeed statically linked so we could see an error during the build if it's not true for some mysterious reason.

georgysavva:
> Maybe we should also enable BuildKit for the images we push to DockerHub on develop and for a release?

Yes, I was going to do that shortly, I just wanted to test it indeed fixes the github actions

Dockerfile
georgysavva: Yes, I agree with you. Reusing `RUN  --mount=type=cache` in GH actions seems to be hard right now. let's postpone it.
The docker layer caching is pretty doable: here are the official docker GH actions for that: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md
and a third-party article https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching

georgysavva:
> Maybe we should also enable BuildKit for the images we push to DockerHub on develop and for a release?

BTW the official docker GH actions use BuildKit by default, so we don't need to change anything for develop and release.
 https://github.com/docker/build-push-action/blob/92e71463491f2d026a477188b8ad3a0fdd9d672c/.github/workflows/main.yml#L27

Dockerfile
Wollac: How is the mounting interacting with the docker build cache?
Before with a `COPY . .` it was assured that if none of the files (outside of .dockerignore) changed, all the successive build steps were skipped.
Is the `mount .` doing the same and is it also skipping files dockerignore? I think it could make sense to still keep the `COPY` for the sources.

Dockerfile
Wollac: I usually add the `-a` just to be on the safe side. However, the Go compiler should be able to correctly detect necessary rebuilds for everything.

Dockerfile
```diff
@@ -19,13 +21,16 @@ ENV GO111MODULE=on
...
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+    -ldflags='-w -s -extldflags "-static"' \
+    -o /go/bin/goshimmer; \
+    ./check_static.sh
```
Wollac, line 33: Is the static check really needed? Shouldn't otherwise the build with `-static` fail anyways?

Dockerfile
georgysavva: It provides the same features as `COPY . .`: if none of the files changed: everything is cached in docker
Yes, it also respects the .dockerignore file. 
This tips if from the official Docker article for Go applications https://www.docker.com/blog/containerize-your-go-developer-environment-part-3/:

> Performing a COPY will create an extra layer in the container image which slows things down and uses extra disk space. This can be avoided by using RUN --mount and bind mounting from the build context, from a stage, or an image. Adopting this pattern, the resulting Dockerfile is as follows.
The default mount type is a read only bind mount from the context that you pass with the docker build command. This means that you can replace the COPY . . with a RUN --mount=target=. wherever you need the files from your context to run a command but do not need them to persist in the final image.

Dockerfile
georgysavva: We don't need it with 99.99% probability. But since we don't rebuild everything from scratch as with `-a`. I added that check to also be on the safe side :)

jonastheis (approved):

georgysavva:
For some reason, GH actions fail when I set docker frontend image via SHA (the preferred way) this commit: https://github.com/iotaledger/goshimmer/pull/1074/commits/36f8a67befe32325e60d41720faf3624bf8491e7
So I used tag instead, it should be fine for now. I opened an issue for that error in docker BuildKit: https://github.com/moby/buildkit/issues/2027

capossele (approved):

PR #1088, opened 2021-03-18
FPC improvements - fixed threshold and bias

daria305:
# Description of change

Add FPC imporvements
 - fixed threshold for last l2 rounds
 - bias towards your own opinion - not tested yet
 
## Type of change

unit test for fixed threshold added

## Change checklist

- [x ] My code follows the contribution guidelines for this project
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes


packages/vote/opinion/opinion.go
```diff
@@ -94,3 +94,14 @@ func ConvertOpinionsToInts32(opinions []Opinion) []int32 {
...
+
+// ConvertOpinionToFloat64 converts the given Opinion to an float64.
+func ConvertOpinionToFloat64(x Opinion) float64 {
+	switch {
+	case x == Like:
+		return 1
+	case x == Dislike:
+		return 2
+	}
+	return 4
```
apenzk, line 106: why 4?

packages/vote/fpc/parameters.go
```diff
@@ -12,10 +12,14 @@ type Parameters struct {
...
 	// The upper bound liked percentage threshold used after the first round.
 	SubsequentRoundsUpperBoundThreshold float64
+	// The fixed liked percentage threshold used in last 'l2' rounds.
+	EndingRoundsFixedThreshold float64
 	// The amount of opinions to query on each round for a given vote context. Also called 'k'.
 	QuerySampleSize int
 	// The amount of rounds a vote context's opinion needs to stay the same to be considered final. Also called 'l'.
 	FinalizationThreshold int
+	// The amount of last rounds for the fixed threshold.
+	FixedEndingThreshold int
```
apenzk, line 22: can we call it TotalRoundsFinalization and TotalRoundsFixedThreshold instead of thresholds, TotalRoundsCoolingOffPeriod instead of CoolingOffPeriod?

packages/vote/vote_context.go
```diff
@@ -74,3 +78,29 @@ func (vc *Context) IsNew() bool {
 func (vc *Context) HadFirstRound() bool {
 	return vc.Rounds == 1
 }
+
+// HadFixedRound tells whether the vote context is in the last l2 rounds of fixed threshold
+func (vc *Context) HadFixedRound(coolingOffPeriod int, finalizationThreshold int, fixedEndingThreshold int) bool {
+	consecutiveRounds := finalizationThreshold - fixedEndingThreshold // l - l2
+	// check whether we have enough opinions to say whether this vote context is finalized.
```
apenzk, line 85: check whether we have enough opinions to say whether the random threshold for this vote context can be fixed.

packages/vote/fpc/fpc.go
```diff
@@ -321,3 +315,32 @@ func (f *FPC) voteContextIDs() (conflictIDs []string, timestampIDs []string) {
...
+func (f *FPC) setThreshold(voteCtx *vote.Context) (float64, float64) {
+	lowerThreshold := f.paras.SubsequentRoundsLowerBoundThreshold
+	upperThreshold := f.paras.SubsequentRoundsUpperBoundThreshold
+
+	if voteCtx.HadFirstRound() {
+		lowerThreshold = f.paras.FirstRoundLowerBoundThreshold
+		upperThreshold = f.paras.FirstRoundUpperBoundThreshold
+	}
+
+	if voteCtx.HadFixedRound(f.paras.CoolingOffPeriod, f.paras.FinalizationThreshold, f.paras.FixedEndingThreshold) {
```
apenzk, line 329: can these paras be passed with f alone or is there an advantage of sending them individually? 
or does this make the setup for the test too complicated?

packages/vote/vote_context.go
```diff
@@ -74,3 +78,29 @@ func (vc *Context) IsNew() bool {
...
+	if len(vc.Opinions[1:]) < coolingOffPeriod + consecutiveRounds {
+		return false
+	}
+	// check weather we have enough opinions and parameters are valid
+	if len(vc.Opinions) < consecutiveRounds || consecutiveRounds < 0 {
```
apenzk, line 90: what is the reason that above we check vc.Opinions[1:] but here vc.Opinions ?

packages/vote/vote_context.go
```diff
@@ -74,3 +78,29 @@ func (vc *Context) IsNew() bool {
 func (vc *Context) HadFirstRound() bool {
 	return vc.Rounds == 1
 }
+
+// HadFixedRound tells whether the vote context is in the last l2 rounds of fixed threshold
+func (vc *Context) HadFixedRound(coolingOffPeriod int, finalizationThreshold int, fixedEndingThreshold int) bool {
+	consecutiveRounds := finalizationThreshold - fixedEndingThreshold // l - l2
```
apenzk, line 84: a better term would be TotalRoundsRandomThreshold

packages/vote/vote_context.go
apenzk: why the case consecutiveRounds < 0 ? Since this is a parameter check should we check it when we initialise the parameters?

packages/vote/fpc/fpc_test.go
```diff
@@ -35,6 +35,27 @@ func TestVoteContext_IsFinalized(t *testing.T) {
 	}
 }
 
+func TestVoteContext_HadFixedRound(t *testing.T) {
+	type testInput struct {
+		voteCtx               vote.Context
+		coolOffPeriod         int
+		finalizationThreshold int
```
apenzk, line 42: would it work if the create a moc context and set the parameters as we need them in the context?

apenzk (changes requested):

packages/vote/vote_context.go
daria305: thx, corrected

packages/vote/vote_context.go
daria305: About parameters wasn't sure where to put the check as no other parameters have been checked.  They are provided in DefaultParameters() but what when we put wrong values in tests?
For vc.Opinions[1:] I took it from isFinalized(). but it can be misleading. Maybe I'll simplify it also there to be consistent

packages/vote/fpc/fpc.go
daria305: a little easier setup for tests

packages/vote/vote_context.go
daria305: corrected

packages/vote/fpc/parameters.go
daria305: looks good, I renamed them

packages/vote/opinion/opinion.go
daria305: this value was used in ConvertOpinionToInt32, not sure why maybe better to use -1? 
And add check: opinion>0 in biasTowardsOwnOpinion()
Anyway, I see I also need to change Like to 1 and Dislike to 0 (according to FPC paper)

packages/vote/fpc/fpc_test.go
daria305: Now it is similar to how we provide it normally in the code. Those parameters are in FPC instanse paras. Do you mean changing the context struct definition?

packages/vote/opinion/opinion.go
apenzk: it would be more consistent. but check with Angelo why there is this current numbering, maybe there is a reason for this

packages/vote/vote_context.go
apenzk: so from our discussion i think it makes sense to ignore failures in parameters input. rather we add a comment to the parameter default description, where we say which range makes sense. This should be in a different PR though, once we merged our current PRs

packages/vote/fpc/fpc_test.go
apenzk: i was considering filling the context outside of the test with many of the fields empty. but as discussed your solution works good so lets keep it your way

apenzk (approved):

PR #1089, opened 2021-03-18
bug fix in Get method of ColoredBalances

lunfardo314:
# Description of change

Fixed wrong type assertion which caused panic when color wasn't present in colored balances

## Type of change

Bug fix

unit test in ledgerstate/utxotests


packages/ledgerstate/color_test.go
```diff
@@ -1,6 +1,7 @@
 package ledgerstate
 
 import (
+	"golang.org/x/crypto/blake2b"
```
lzpap, line 4: move this to the bottom half, then do `goimports`

capossele (approved):

PR #1091, opened 2021-03-18
Feat/mana fpc/split statement

apenzk:
# Description of change

If there are many conflicts, FPC statements may become larger than the maximum allowed message size. This is resolved in this PR. Also the variable for a proportion of Liked Opinions is renamed ProportionLiked.

## Type of change

## Change checklist


plugins/messagelayer/consensus_plugin.go
```diff
@@ -459,9 +461,19 @@ func makeStatement(roundStats *vote.RoundStats) {
 			)
 		default:
 		}
+		// limit the size of statements
+		maxSize := payload.MaxSize
+		if (len(conflicts)*statement.ConflictLength + len(timestamps)*statement.TimestampLength) >= int(0.9*float64(maxSize)) {
+			broadcastStatement(conflicts, timestamps)
```
piotrm50, line 467: You should replace this if with a `while` loop that keeps sending sub-slices of `conflicts` and `timestamps` until you run out of elements in both slices. 
Currently this `if` statement doesn't split the conflicts into multiple statements.

piotrm50 (changes requested):
Split statement logic looks broken

plugins/messagelayer/consensus_plugin.go
apenzk: It shouldn't split it into separate statements. Every time the counter reaches the max size of a statement it will send a statement and reset the counters. 

But your approach would also work. Does it have advantages over this one?

piotrm50: Sorry! - I thought it's outside of the `for` loop. It's ok then :)

piotrm50 (approved):

daria305 (approved):

PR #1094, opened 2021-03-18
WebAPI /ledgerstate/outputs & addresses Endpoints

lzpap:
# Endpoints:
 - /ledgerstate/outputs/:outputID 
 - /ledgerstate/outputs/:outputID/consumers 
 - /ledgerstate/outputs/:outputID/metadata
 -  /ledgerstate/addresses/:address
 - /ledgerstate/addresses/:address/unspentOutputs

### DIFF
--- a/plugins/webapi/ledgerstate/output.go
+++ b/plugins/webapi/ledgerstate/output.go
@@ -47,7 +47,7 @@ func GetOutputConsumersEndPoint(c echo.Context) (err error) {
 		}
 	}
 
-	return c.JSON(http.StatusOK, NewConsumers(consumers))
+	return c.JSON(http.StatusOK, NewConsumers(outputID, consumers))
 }
 
 // GetOutputMetadataEndPoint is the handler for the /ledgerstate/outputs/:outputID/metadata endpoint.
@@ -74,10 +74,12 @@ func GetOutputMetadataEndPoint(c echo.Context) (err error) {
 
 // Output represents a JSON model of a ledgerstate.Output.
 type Output struct {
-	ID       string           `json:"id"`
-	Type     string           `json:"type"`
-	Balances []ColoredBalance `json:"balances"`
-	Address  string           `json:"address"`
+	ID            string           `json:"id"`
+	TransactionID string           `json:"transactionID"`
+	OutputIndex   uint16           `json:"outputIndex"`
+	Type          string           `json:"type"`
+	Balances      []ColoredBalance `json:"balances"`
+	Address       string           `json:"address"`
 }
 
 // ColoredBalance represents a JSON model of a single colored balance.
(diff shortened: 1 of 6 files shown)


plugins/webapi/ledgerstate/output.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region Output ///////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Output represents a JSON model of a ledgerstate.Output.
+type Output struct {
+	ID       string           `json:"id"`
+	Type     string           `json:"type"`
```
hmoog, line 78: I think it would be nice to also explicitly show the TransactionID and the OutputIndex here. In code you can derive that from the OutputID so we don't have that as properties in the model but since this logic is not available to a potential API client, it would be nice to just spell these two values out in the JSON model.

plugins/webapi/ledgerstate/output.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region Consumers ////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Consumers is the JSON model of consumers of the output.
+type Consumers struct {
+	ConsumerCount        int                   `json:"consumerCount"`
```
hmoog, line 112: I am not sure if we really need the ConsumerCount here as you can directly derive that from the list of Transactions but maybe it would be nice to have the OutputID named here so the response says which Consumers we are showing.

plugins/webapi/ledgerstate/output.go
```diff
@@ -0,0 +1,168 @@
...
+	ConsumerTransactions []ConsumerTransaction `json:"consumers"`
```
hmoog, line 113: Maybe rename it to "ConsumingTransactions" and the model to "ConsumingTransaction" since the "Consumers" model the relationship between an Output and its consuming Transactions.

hmoog (changes requested):

plugins/webapi/ledgerstate/output.go
```diff
@@ -0,0 +1,168 @@
...
+	return c.JSON(http.StatusOK, NewOutputMetadata(outputMetadata))
+}
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region Output ///////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Output represents a JSON model of a ledgerstate.Output.
+type Output struct {
+	ID       string           `json:"id"`
```
hmoog, line 77: I think it would be nice to also explicitly show the TransactionID and the OutputIndex here by using the `OutputID` model from `NewOutputID` instead of just a string.

In code you can derive that from the OutputID so we don't have that as properties in the model but since this logic is not available to a potential API client, it would be nice to just spell these two values out in the JSON model.

hmoog (changes requested):

plugins/webapi/ledgerstate/output.go
lzpap: I think it doesn't hurt to have `consumerCount`, especially when we will be using this for debugging.
I've added the `outputID`.

plugins/webapi/ledgerstate/output.go
lzpap: done!

PR #1098, opened 2021-03-19
View transaction from explorer 

acha-bill:
View transaction from explorer 
`/explorer/transaction/:id`

Screenshot 2021-03-22 at 14 58 54


jkrvivian:
@acha-bill, it looks really nice! 

It would be nice to have also:
1. the transaction search in the explorer text input and the one on the top right corner.
2. the link to the transaction ID of the message page, same as strong parent and address:
image

lzpap (commented):
great work @acha-bill! 
We should think about harmonizing and improving how we display transactions, becasue at the moment there are two transaction views: one that is displayed if a transaction is displayed on the message query result page, and one that you just introduced. By unifing the JSON structs for a transaction payload (`payload_handler.go`) and the transaction api response (`webapi/ledgerstate/transaction.go`) we can use one single transaction view definition in the frontend. Transaction metadata and attachments components in the frontend can remain the same, only the `ExplorerTransaction.tsx` and `TransactionPayload.tsx` need to be "merged" together. Wdyt?

acha-bill:
@lzpap The merging of the 2 components would better be done in a future refactoring PR. 
This new component is designing around the REST APIs. It's ugly to integrate with the existing logic. 
Same thing goes for the search. @jkrvivian

plugins/dashboard/frontend/src/app/components/ExplorerTransaction.tsx
```diff
@@ -0,0 +1,149 @@
...
+                <p> {txId} </p>
+
+
+                {tx &&
+                    <Row className={"mb-3"}>
+                        <Col>
+                            <ListGroup>
+                                <ListGroup.Item>ID: {txId}</ListGroup.Item>
+                                <ListGroup.Item>Version: {tx.version}</ListGroup.Item>
+                                <ListGroup.Item>Timestamp: {dateformat(new Date(tx.timestamp * 1000), "dd.mm.yyyy HH:MM:ss")}</ListGroup.Item>
```
lzpap, line 71: we shouldn't use `dateformat` becasue it will cause problems in some browsers. use `toLocaleTimeString()`

plugins/dashboard/frontend/src/app/components/ExplorerTransaction.tsx
```diff
@@ -0,0 +1,149 @@
...
+                                      <div style={style}>
+                                            <span>Outputs</span>
+                                            <hr/>
+                                            {tx.outputs.map((output, i) => {
+                                                return (
+                                                    <div className={"mb-2"} key={i}>
+                                                        <span className={"mb-2"}>Index: <Badge variant={"primary"}>{i}</Badge></span>
+                                                        <ListGroup>
+                                                            <ListGroup.Item>ID: <a href={"#"}>{output.outputID.base58}</a></ListGroup.Item>
+                                                            <ListGroup.Item>Address: <a href={"#"}> {output.address}</a></ListGroup.Item>
```
lzpap, line 125: a href to address route!

plugins/dashboard/frontend/src/app/components/ExplorerTransaction.tsx
```diff
@@ -0,0 +1,149 @@
...
+                                                                </div>
+                                                            </ListGroup.Item>
+                                                        </ListGroup>
+                                                    </div>
+                                                )
+                                            })}
+                                        </div>
+                                    </div>
+                                </ListGroup.Item>
+                                <ListGroup.Item>Data payload: {tx.dataPayload}</ListGroup.Item>
```
lzpap, line 141: you can write it like this:
```
{ tx.dataPayload && <ListGroup.Item>Data payload: {tx.dataPayload}</ListGroup.Item>}
```
and then if data payload is null, no empty listgroupitem is displayed.

plugins/dashboard/frontend/src/app/components/ExplorerTransactionMetadata.tsx
```diff
@@ -0,0 +1,81 @@
...
+                                    <td>Branch ID</td>
+                                    <td><a href="#">{txMetadata.branchID}</a></td>
+                                </tr>
+                                <tr>
+                                    <td>Solid</td>
+                                    <td>{txMetadata.solid.toString()}</td>
+                                </tr>
+                                <tr>
+                                    <td>Solidification time</td>
+                                    <td>{dateformat(new Date(txMetadata.solidificationTime * 1000), "dd.mm.yyyy HH:MM:ss")}</td>
```
lzpap, line 64: take care of dateformat!

plugins/dashboard/frontend/src/app/stores/ExplorerStore.tsx
```diff
@@ -15,6 +15,7 @@ import {Link} from 'react-router-dom';
 import {RouterStore} from "mobx-react-router";
 
 export const GenesisMessageID = "1111111111111111111111111111111111111111111111111111111111111111";
+export const GenesisTransactionID = "1111111111111111111111111111111111111111111111111111111111111111"
```
lzpap, line 18: is there such a thing as genesis tx? I thought we only have genesis message and output.

lzpap (changes requested):

plugins/dashboard/frontend/src/app/components/ExplorerTransactionMetadata.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    <p> Metadata for transaction ID {txId} not found.</p>
+                </Container>
+            );
+        }
+        return (
+            <Container>
+            <h4>Metadata</h4>
+                {txMetadata && <Row className={"mb-3"}>
+                    <Col>
+                        <table className={"table table-bordered table-condensed table-sm"}>
```
jkrvivian, line 51: Why table here instead of ListGroup?
ListGroup looks more consistent with other layouts. :stuck_out_tongue:

lzpap (approved):

plugins/dashboard/frontend/src/app/components/ExplorerTransactionMetadata.tsx
acha-bill: coz its compact. :)

PR #1102, opened 2021-03-20
Statement only considered if issued during last round.

piotrm50:
# Description of change

Only consider statement if a node issued a statement during last round.

## Type of change

Integration and unit tests pass. Manually tested by running FPC.

## Change checklist


packages/vote/statement/registry.go
```diff
@@ -181,6 +183,14 @@ func (v *View) AddTimestamps(timestamps Timestamps) {
 	}
 }
 
+// UpdateLastStatementReceivedTime updates last statement issuing time in node's View.
+func (v *View) UpdateLastStatementReceivedTime(statementIssuingTime time.Time) {
+	v.lstMutex.RLock()
+	defer v.lstMutex.RUnlock()
+
+	v.LastStatementReceivedTimestamp = statementIssuingTime
```
daria305, line 191: Maybe it's better to stick with statementReceivingTime instead of statementIssuingTime as we are relaying only on the time when statement was read by our node, so this could be misleading?

plugins/messagelayer/consensus_plugin.go
```diff
@@ -225,7 +226,9 @@ type OpinionGivers map[identity.ID]OpinionGiver
 // Query retrieves the opinions about the given conflicts and timestamps.
 func (o *OpinionGiver) Query(ctx context.Context, conflictIDs []string, timestampIDs []string) (opinions opinion.Opinions, err error) {
 	for i := 0; i < StatementParameters.WaitForStatement; i++ {
-		if o.view != nil {
+		// query statement only if it's been received within the last round interval
```
apenzk, line 229: query statement status only if any statement has been received within the last round interval. This is important. We do want to consider old statements IF the node has been active in the last round.

plugins/messagelayer/consensus_plugin.go
```diff
@@ -225,7 +226,9 @@ type OpinionGivers map[identity.ID]OpinionGiver
...
+		// otherwise node might be down and we don't want to look at old statements in subsequent rounds
```
apenzk, line 230: see above. basically, if the node has been active in this round, we should consider the last statement opinion about a conflict of the node, even if the opinion is already 10 rounds old. this is because the node might have finalized in FPC and doesn't chat about a certain conflict anymore - yet we have to consider its last broadcast opinion.

apenzk: so we have to make sure that the statement opinions dont get emptied, and will be considered in subsequent rounds (if the node is still active)

apenzk (changes requested):

packages/vote/statement/registry.go
piotrm50: Good point ;)

plugins/messagelayer/consensus_plugin.go
piotrm50: I adjusted the comment to your suggestion.

plugins/messagelayer/consensus_plugin.go
piotrm50: That's how it works - conflicts are removed from the statement registry some time after last arrival (5 minutes by default), when the voting is resolved. 
I might have not described it correctly, but it works as you described - in this improvement we check if the node received a statement within last 10 seconds and it works independently of opinion cleaning in registry.

PR #1123, opened 2021-03-24
Feat/fpc mana conditional write statement

piotrm50:
# Description of change

Add condition to only read and write a statement if the issuing node has enough mana.

## Type of change

Manual tests with conflicting transactions and distributed mana, unit tests

## Change checklist


packages/vote/fpc/fpc.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/iotaledger/goshimmer/packages/vote/opinion"
 )
 
+const (
+	tolerance = 0.001
```
apenzk, line 21:
```suggestion
	toleranceTotalMana = 0.001
```

packages/vote/fpc/fpc.go
```diff
@@ -339,7 +342,8 @@ func ManaBasedSampling(opinionGivers []opinion.OpinionGiver, maxQuerySampleSize,
 	}
 
 	// check if total mana is almost zero
-	if math.Abs(totalConsensusMana-0.0) <= 1e-9 {
+
+	if math.Abs(totalConsensusMana) <= tolerance {
```
apenzk, line 346:
```suggestion
	if math.Abs(totalConsensusMana) <= toleranceTotalMana {
```

plugins/messagelayer/consensus_plugin.go
```diff
@@ -453,9 +464,25 @@ type statementLog struct {
 
 // region Statement ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-func makeStatement(roundStats *vote.RoundStats) {
-	// TODO: add check for Mana threshold
+func checkEnoughMana(id identity.ID, threshold float64) bool {
```
apenzk, line 467: description comment missing before function

plugins/messagelayer/consensus_plugin.go
```diff
@@ -255,16 +258,19 @@ func OpinionGiverFunc() (givers []opinion.OpinionGiver, err error) {
...
 	consensusManaNodes, _, err := GetManaMap(mana.ConsensusMana)
-
+	if err != nil {
+		plugin.LogErrorf("Error retrieving consensus mana: %s", err)
+	}
 	for _, v := range Registry().NodesView() {
-		// double check to exclude self
-		if v.ID() == local.GetInstance().ID() {
+		// double check to exclude self and check if node has enough mana to consider its statement
+		if v.ID() == local.GetInstance().ID() || !checkEnoughMana(v.ID(), StatementParameters.ReadManaThreshold) {
```
apenzk, line 266: I dont think we should apply checkEnoughMana here.
a) our perception might be wrong
b) if the node has to didact this from its allowed rate the statement has legitimately passed the rate control
c) if we don't make didact statements from the allowed rate we might have to FPC-vote on whether a statement is ok to be added to the tangle ->circular problem?

plugins/messagelayer/mana_plugin.go
```diff
@@ -296,6 +296,22 @@ func GetHighestManaNodes(manaType mana.Type, n uint) ([]mana.Node, time.Time, er
 	return bmv.GetHighestManaNodes(n)
 }
 
+// GetHighestManaNodesFraction returns the given percent `p` of highest type mana nodes in descending order.
```
apenzk, line 299: This does work well if the mana is following some zipf distribution. But does not work if mana is distributed equally. but we need to cover the latter case as well..

apenzk (changes requested):

apenzk (approved):

PR #1190, opened 2021-04-08
Check consumed bytes in payload unmarshalFunc

lzpap:
# Description of change
Previously, serialized payloads could have trailing, unparsed bytes and were still considered valid.
This PR introduces an extra check that makes sure that the exact number of bytes are unmarshaled, otherwise an error is returned
and the payload is considered to be invalid.

### DIFF
--- a/plugins/syncbeacon/payload/payload.go
+++ b/plugins/syncbeacon/payload/payload.go
@@ -26,7 +26,6 @@ var Type = payload.NewType(200, ObjectName, func(data []byte) (payload payload.P
 		return nil, xerrors.New("not all payload bytes were consumed")
 	}
 	return
-	return
 })
 
 // Payload represents the syncbeacon payload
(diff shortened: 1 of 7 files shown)


plugins/syncbeacon/payload/payload.go
```diff
@@ -16,8 +17,15 @@ const (
...
-
+	var consumedBytes int
+	payload, consumedBytes, err = FromBytes(data)
+	if err != nil {
+		return nil, err
+	}
+	if consumedBytes != len(data) {
+		return nil, xerrors.New("not all payload bytes were consumed")
+	}
+	return
```
capossele, line 28: Is this a Github issue? I see two return

lzpap: my bad, pushed a fix

capossele (approved):

PR #1198, opened 2021-04-13
feat: Introduce FinalizedApprovalWeight flag

jkrvivian:
# Description of change
* add a boolean `FinalizedApprovalWeight` to the message metadata and add the field to API, dashboard
* update `FinalizedApprovalWeight` on the MarkerConfirmed event by walking in the marker’s past cone 

## Type of change
- Enhancement 

## Change checklist

### DIFF
--- a/packages/tangle/approvalweightmanager.go
+++ b/packages/tangle/approvalweightmanager.go
@@ -275,7 +275,7 @@ func branchIDCaller(handler interface{}, params ...interface{}) {
 }
 
 func markerCaller(handler interface{}, params ...interface{}) {
-	handler.(func(*markers.Marker, MessageID))(params[0].(*markers.Marker), params[1].(MessageID))
+	handler.(func(*markers.Marker))(params[0].(*markers.Marker))
 }
 
 // endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/packages/tangle/approvalweightmanager.go
+++ b/packages/tangle/approvalweightmanager.go
@@ -138,10 +138,16 @@ func (a *ApprovalWeightManager) onMarkerConfirmed(marker *markers.Marker) {
 		metadata.SetFinalizedApprovalWeight(true)
 	})
 
-	// start walking from strong parents
-	// TODO: what about weak parents?
 	entryMessageIDs := MessageIDs{}
 	a.tangle.Storage.Message(messageID).Consume(func(message *Message) {
+		// mark weak parents as finalized but not propagate finalized flag to its past cone
+		message.ForEachWeakParent(func(parentID MessageID) {
+			a.tangle.Storage.MessageMetadata(parentID).Consume(func(metadata *MessageMetadata) {
+				metadata.SetFinalizedApprovalWeight(true)
+			})
+		})
+
+		// propagate finalized flag to strong parents' past cone
 		message.ForEachStrongParent(func(parentID MessageID) {
 			entryMessageIDs = append(entryMessageIDs, parentID)
 		})
@@ -161,7 +167,14 @@ func (a *ApprovalWeightManager) propagateFinalizedApprovalWeight(message *Messag
 		return
 	}
 
-	// TODO: what about weak parents?
+	// mark weak parents as finalized but not propagate finalized flag to its past cone
+	message.ForEachWeakParent(func(parentID MessageID) {
+		a.tangle.Storage.MessageMetadata(parentID).Consume(func(metadata *MessageMetadata) {
+			metadata.SetFinalizedApprovalWeight(true)
+		})
+	})
+
+	// propaget finalized to strong parents
 	message.ForEachStrongParent(func(parentID MessageID) {
 		finalizedWalker.Push(parentID)
 	})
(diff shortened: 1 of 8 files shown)


packages/tangle/approvalweightmanager.go
```diff
@@ -86,6 +87,70 @@ func (a *ApprovalWeightManager) weightsPerEpoch(branchID ledgerstate.BranchID) *
 	return weightsPerEpoch
 }
 
+func (a *ApprovalWeightManager) onMarkerConfirmed(marker *markers.Marker, messageID MessageID) {
+	// walk through the marker's pastcone to update FinalizedApprovalWeight
+	finalizedWalker := walker.New()
+
+	// push marker itself
+	finalizedWalker.Push(messageID)
```
jonastheis, line 95: Why do we need to push the message itself if we already do `metadata.SetFinalizedApprovalWeight(true)`?

packages/tangle/approvalweightmanager.go
```diff
@@ -127,7 +192,7 @@ func (a *ApprovalWeightManager) onSequenceSupportUpdated(marker *markers.Marker,
 		a.lastConfirmedMarkers[marker.SequenceID()] = currentMarker.Index()
 
 		fmt.Println("Marker confirmed!!")
-		a.Events.MarkerConfirmed.Trigger(currentMarker)
+		a.Events.MarkerConfirmed.Trigger(currentMarker, a.getMessageIDOfMarker(currentMarker, branchID))
```
jonastheis, line 195: This is my bad, we should just trigger the event with the current marker and use the marker to message mapping as introduced in https://github.com/iotaledger/goshimmer/pull/1200 to get the corresponding message. Then we don't need `getMessageIDOfMarker` as well.

packages/tangle/approvalweightmanager.go
```diff
@@ -86,6 +87,70 @@ func (a *ApprovalWeightManager) weightsPerEpoch(branchID ledgerstate.BranchID) *
 	return weightsPerEpoch
 }
 
+func (a *ApprovalWeightManager) onMarkerConfirmed(marker *markers.Marker, messageID MessageID) {
+	// walk through the marker's pastcone to update FinalizedApprovalWeight
+	finalizedWalker := walker.New()
```
jonastheis, line 92: Instead of doing all this manually you can simply use `WalkMessageAndMetadata`. This should also simplify `propagateFinalizedApprovalWeight` as you don't need to consume.

packages/tangle/approvalweightmanager.go
```diff
@@ -86,6 +87,70 @@ func (a *ApprovalWeightManager) weightsPerEpoch(branchID ledgerstate.BranchID) *
...
+
+	// push marker itself
+	finalizedWalker.Push(messageID)
+	a.tangle.Storage.MessageMetadata(messageID).Consume(func(metadata *MessageMetadata) {
+		metadata.SetFinalizedApprovalWeight(true)
+	})
+	// push parents, keep walking
+	a.tangle.Storage.Message(messageID).Consume(func(message *Message) {
+		message.ForEachParent(func(parent Parent) {
```
jonastheis, line 101: I'm not sure if we can propagate this over all parents. Strong parents for sure, but weak parents might need a special treatment?

jonastheis (changes requested):

packages/tangle/approvalweightmanager.go
jkrvivian: yes, we need to discuss this. I'll replace it with strong parents for now.

jkrvivian: for weak parent: mark finalized, but don't propagate.

packages/tangle/approvalweightmanager.go
```diff
@@ -126,6 +127,57 @@ func (a *ApprovalWeightManager) WeightOfMarker(marker *markers.Marker, anchorTim
...
+			})
+		})
+
+		// propagate finalized flag to strong parents' past cone
+		message.ForEachStrongParent(func(parentID MessageID) {
+			entryMessageIDs = append(entryMessageIDs, parentID)
+		})
+	})
+
+	a.tangle.Utils.WalkMessageAndMetadata(a.propagateFinalizedApprovalWeight, entryMessageIDs, true)
```
jonastheis, line 154: Do we need to revisit elements?

packages/tangle/approvalweightmanager.go
```diff
@@ -126,6 +127,57 @@ func (a *ApprovalWeightManager) WeightOfMarker(marker *markers.Marker, anchorTim
...
+	}
+
+	// mark weak parents as finalized but not propagate finalized flag to its past cone
+	message.ForEachWeakParent(func(parentID MessageID) {
+		a.tangle.Storage.MessageMetadata(parentID).Consume(func(metadata *MessageMetadata) {
+			metadata.SetFinalizedApprovalWeight(true)
+		})
+	})
+
+	// propaget finalized to strong parents
```
jonastheis, line 175:
```suggestion
	// propagate finalized to strong parents
```

jonastheis (changes requested):

PR #1202, opened 2021-04-14
Integrate dRNG with FPC

capossele:


packages/drng/ticker.go
```diff
@@ -0,0 +1,104 @@
...
+			if clock.Since(t.dRNGState().Randomness().Timestamp) < time.Duration(t.awaitOffset)*time.Second {
+				randomness = t.dRNGState().Randomness().Float64()
+				if t.dRNGTicker != nil {
+					t.dRNGTicker.Reset(time.Duration(t.interval) * time.Second)
+				}
+				break
+			}
+			time.Sleep(intervalCheckDRNG * time.Millisecond)
+		}
+	}
```
jkrvivian, line 92: I came up with an idea to improve this waiting loop, but not sure if it's better or I was wrong about something.

here's how it works:
1. When drngTicker ticks, we first check if the randomness is fresh (`clock.Since(t.dRNGState().Randomness().Timestamp) < time.Duration(t.resolution)*time.Second`)
    * if yes, then we send out the randomness and reset drngTicker
    * if no, we reset drngTicker to `awaitOffset`
2. In the for loop, we wait for `awaitOffset`, listen to Randomness event to give us new randomness. If the clock ticks again (no fresh randomness), we use the default value

```suggestion
if t.dRNGState() != nil {
    // check if the randomess is "fresh"
    if t.dRNGTicker != nil && clock.Since(t.dRNGState().Randomness().Timestamp) < time.Duration(t.interval)*time.Second {
	randomness = t.dRNGState().Randomness().Float64()
	// the theoretical time that we should receive a new randomness
	timeToNextDRNG := t.dRNGState().Randomness().Timestamp.Add(time.Duration(t.interval) * time.Second).Sub(clock.SyncedTime())
	t.dRNGTicker.Reset(timeToNextDRNG)
    } else {
	// set ticker to awaitOffset
	t.dRNGTicker.Reset(time.Duration(t.awaitOffset) * time.Second)
    }

out:
    // if fresh randomness is not received, we wait for new randomness for awaitOffset seconds
    for {
	// abort if we already get the latest randomness
	if randomness != t.defaultValue {
		break out
	}

	select {
	    // receive randomness from Randomness event
	    case drngstate := <-t.fromRandamnessEvent:
		// check if the randomness is "fresh"
		if clock.Since(drngstate.Randomness().Timestamp) < time.Duration(t.awaitOffset)*time.Second {
			randomness = t.dRNGState().Randomness().Float64()
			if t.dRNGTicker != nil {
				t.dRNGTicker.Reset(time.Duration(t.interval) * time.Second)
			}
			break out
		}
	    case <-t.dRNGTicker.C:
		// still no new randomness within awaitOffset, take the default value, and reset dRNGTicker
		now := clock.SyncedTime().Unix()
		t.dRNGTicker.Reset(time.Duration(ResolveNextTimePoint(now, t.interval)-now) * time.Second)
		break out
	}
    }
}
```

apenzk (approved):

PR #1205, opened 2021-04-15
always like future time stamps

apenzk:
# Description of change
Always like future transactions

## Type of change
set like to all future messages. This is a temporary solution until the booker/scheduler is updated.

### DIFF
--- a/packages/consensus/fcob/timestamp.go
+++ b/packages/consensus/fcob/timestamp.go
@@ -18,13 +18,23 @@ var (
 // region TimestampQuality /////////////////////////////////////////////////////////////////////////////////////////////
 
 // TimestampQuality returns the TimestampOpinion based on the given times (e.g., arrival and current).
-func TimestampQuality(messageID tangle.MessageID, target, current time.Time) (timestampOpinion *TimestampOpinion) {
-	diff := abs(current.Sub(target))
-
+func TimestampQuality(messageID tangle.MessageID, target, current time.Time) (timestampOpinion *TimestampOpinion, err error) {
 	timestampOpinion = &TimestampOpinion{
 		MessageID: messageID,
 		Value:     opinion.Like,
 	}
+
+	diff := current.Sub(target)
+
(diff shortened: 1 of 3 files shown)


packages/consensus/fcob/timestamp.go
```diff
@@ -18,13 +18,23 @@ var (
 // region TimestampQuality /////////////////////////////////////////////////////////////////////////////////////////////
 
 // TimestampQuality returns the TimestampOpinion based on the given times (e.g., arrival and current).
-func TimestampQuality(messageID tangle.MessageID, target, current time.Time) (timestampOpinion *TimestampOpinion) {
-	diff := abs(current.Sub(target))
-
+func TimestampQuality(messageID tangle.MessageID, target, current time.Time) (timestampOpinion *TimestampOpinion, err error) {
```
capossele, line 21: we do not use the returned error so let's remove it

packages/tangle/scheduler.go
```diff
@@ -98,6 +98,10 @@ func (s *Scheduler) scheduleMessage(messageID MessageID) {
 		return
 	}
 
+	// if !s.isInTheFuture(messageID) {
+	// 	return
+	// }
+
```
capossele, line 104:
```suggestion
```

packages/tangle/scheduler.go
```diff
@@ -127,6 +131,20 @@ func (s *Scheduler) parentsBooked(messageID MessageID) (parentsBooked bool) {
...
+// 		current := clock.SyncedTime()
+// 		// current := time.Now()
+// 		if current.Sub(message.issuingTime) < 0 {
+// 			isInTheFuture = true
+// 		}
+// 	})
+
+// 	return
+// }
+
```
capossele, line 147:
```suggestion
```

capossele (changes requested):

apenzk:
Make TimestampQuality assymetric

PR #1212, opened 2021-04-21
add delayedRoundStart

apenzk:
# Description of change
Subtract the time we are already in a new Round if we are missing a dRNG message

## Type of change

### DIFF
--- a/packages/drng/ticker.go
+++ b/packages/drng/ticker.go
@@ -1,7 +1,6 @@
 package drng
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


packages/drng/ticker.go
```diff
@@ -111,7 +122,8 @@ func (t *Ticker) send() {
 			case randomnessEvent := <-t.fromRandomnessEvent:
 				// check if the randomness is "fresh"
 				if clock.Since(randomnessEvent.Timestamp) < time.Duration(t.awaitOffset)*time.Second {
-					t.delayedRoundStart = 0
+					delay := time.Since(start)
+					t.setDelayedRoundStart(delay)
```
apenzk, line 126: nope. the delay is zero because we reset the timer for the full 10sec here

packages/drng/ticker.go
```diff
@@ -120,8 +132,9 @@ func (t *Ticker) send() {
 				}
 			case <-t.dRNGTicker.C:
 				// still no new randomness within awaitOffset, take the default value, and reset dRNGTicker
+				delay := time.Since(start)
+				t.setDelayedRoundStart(delay)
```
apenzk, line 136: this is not correct, because the ticker is set for 7 second. so the delay in the DelayedRoundStart should be precisely 3sec

capossele (approved):

PR #1216, opened 2021-04-22
Feat/ New scheduler

acha-bill:
Congestion control + Rate setter


packages/tangle/scheduler.go
```diff
@@ -1,133 +1,314 @@
...
 	}
 
-	s.tangle.Storage.MessageMetadata(messageID).Consume(func(messageMetadata *MessageMetadata) {
-		if messageMetadata.SetScheduled(true) {
-			if s.scheduledMessages.Add(messageID) {
-				s.allMessagesScheduledWG.Add(1)
-			}
-			s.Events.MessageScheduled.Trigger(messageID)
+	readyNode := false
+	now := time.Now()
```
jonastheis, line 244: This should use `clock.SyncedTime()` I think.

jonastheis (approved):

tools/integration-tests/tester/tests/common/common_test.go
```diff
@@ -27,36 +31,36 @@ func TestSynchronization(t *testing.T) {
 
 	// 1. issue data messages
 	ids := tests.SendDataMessagesOnRandomPeer(t, n.Peers(), numMessages)
-
-	// wait for messages to be gossiped
-	time.Sleep(10 * time.Second)
+	log.Printf("Issuing %d messages to be synced... done\n", numMessages)
+	// wait for messages to be gossiped and leaf the synced time windows
+	time.Sleep(framework.ParaTangleTimeWindow)
```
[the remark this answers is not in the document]
Wollac: It has to sleep for `ParaTangleTimeWindow` as otherwise the node will not go out of sync.
But we can sure just decrease the value of `ParaTangleTimeWindow` in the integration tests, I think it even used to be 30s at some point

plugins/messagelayer/plugin.go
```diff
@@ -65,6 +69,62 @@ func configure(plugin *node.Plugin) {
...
+	Tangle().Scheduler.Events.MessageScheduled.Attach(events.NewClosure(func(messageID tangle.MessageID) {
+		plugin.LogDebugf("message scheduled in Scheduler: %s", messageID.Base58())
+	}))
+
+	Tangle().Events.MessageInvalid.Attach(events.NewClosure(func(messageID tangle.MessageID) {
+		plugin.LogInfof("message invalid: %s", messageID.Base58())
+	}))
+
+	Tangle().Booker.Events.MessageBooked.Attach(events.NewClosure(func(messageID tangle.MessageID) {
+		plugin.LogDebugf("message booked: %s", messageID.Base58())
```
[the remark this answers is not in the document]
Wollac: We should definitely keep the discarded and blacklisted messages as this should not happen at all in a happy world scenario. I think all the booking messages indeed does not add much information.

tools/integration-tests/tester/tests/common/common_test.go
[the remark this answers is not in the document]
jkrvivian: Ah! I changed it to 2 min for debugging :stuck_out_tongue:

plugins/messagelayer/plugin.go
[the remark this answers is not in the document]
jkrvivian: Exactly what I just pushed :smile:

packages/tangle/tangle.go
```diff
@@ -199,6 +263,9 @@ type Events struct {
 
 	// Error is triggered when the Tangle faces an error from which it can not recover.
 	Error *events.Event
+
+	// Info is triggered when the tangle should log the info provided.
+	Info *events.Event
```
Wollac, line 268: This is never triggered, so why was this added? It is just a left-over debug artifact or is it a valuable addition?

plugins/webapi/jsonmodels/info.go
```diff
@@ -50,3 +56,16 @@ type Mana struct {
 	Consensus          float64   `json:"consensus"`
 	ConsensusTimestamp time.Time `json:"consensusTimestamp"`
 }
+
+// Scheduler is the scheduler details.
+type Scheduler struct {
+	Running        bool           `json:"running"`
+	Rate           string         `json:"rate"`
+	NodeQueueSizes map[string]int `json:"nodeQueueSizes"`
```
Wollac, line 64: Is this really a meaningful metric? Potentially, the number of nodes in the queue could be rather large and is seeing the queue size per cryptic `nodeID` really helpful?

tools/integration-tests/tester/framework/framework.go
```diff
@@ -320,26 +336,28 @@ func (f *Framework) CreateDRNGNetwork(name string, members, peers, minimumNeighb
...
 	}
 	if !config.Mana {
 		return nil, fmt.Errorf("mana plugin is required to load mana snapshot")
 	}
+
+	n, err := f.CreateNetwork(name, peers, minimumNeighbors, config)
+	if err != nil {
+		return nil, err
+	}
+	time.Sleep(15 * time.Second)
```
Wollac, line 350: The sleep was added, what is its purpose? The `CreateNetwork` waits for successful peering, doesn't this also assure that the every node should be running?

tools/integration-tests/tester/framework/parameters.go
```diff
@@ -83,6 +91,8 @@ type GoShimmerConfig struct {
 	DisabledPlugins    string
 	SnapshotFilePath   string
 
+	StartSync bool
```
Wollac, line 94: I believe this should be `StartSynced`

tools/integration-tests/tester/tests/consensus/consensus_conflicts_test.go
```diff
@@ -92,16 +83,18 @@ func TestConsensus(t *testing.T) {
...
+		_, err = n.Peers()[0].SendTransaction(pledgingTxs[receiverId].Bytes())
+		require.NoError(t, err)
 		receiverId++
+		time.Sleep(2 * time.Second)
 	}
 	// sleep 3* the avg. network delay so both partitions confirm their own pledging transaction
 	// and 1 avg delay more to make sure each node has mana
 	log.Printf("waiting 2 * %d seconds avg. network delay + 5s to make the transactions confirmed", framework.ParaFCoBAverageNetworkDelay)
-	time.Sleep(time.Duration(framework.ParaFCoBAverageNetworkDelay)*2*time.Second + 5*time.Second)
+	//time.Sleep(time.Duration(framework.ParaFCoBAverageNetworkDelay)*2*time.Second + 5*time.Second)
```
Wollac, line 94: The Sleep is removed, but the log is still there?

Why has the sleep been moved up? Doing the more sophisticated sleep based on `ParaFCoBAverageNetworkDelay` seems like a good idea...

tools/integration-tests/tester/tests/consensus/consensus_conflicts_test.go
```diff
@@ -92,16 +83,18 @@ func TestConsensus(t *testing.T) {
...
 
 	// Wait for transaction to be finalized via the sync beacons weight
-	time.Sleep(20 * time.Second)
+	//time.Sleep(20 * time.Second)
```
Wollac, line 97: Remove instead of comment, if no longer needed

Wollac:
Why is `Integration tests / value` failing again? Could this be because of 8385fcc?

packages/tangle/storage.go
```diff
@@ -122,6 +122,12 @@ func (s *Storage) Setup() {
 	s.tangle.Parser.Events.MessageParsed.Attach(events.NewClosure(func(msgParsedEvent *MessageParsedEvent) {
 		s.tangle.Storage.StoreMessage(msgParsedEvent.Message)
 	}))
+	s.tangle.Scheduler.Events.MessageDiscarded.Attach(events.NewClosure(func(messageID MessageID) {
+		// make sure the event is triggered when the Scheduler is running
+		if s.tangle.Scheduler.Running() {
```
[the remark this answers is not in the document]
jkrvivian: This event should be triggered in the scheduler only, so this check is unnecessary.

packages/tangle/tangle.go
jkrvivian: I think it's just for debugging. Let's remove it.

plugins/webapi/jsonmodels/info.go
jkrvivian: With this we can observe the congestion of each node, and see how it works.

tools/integration-tests/tester/tests/consensus/consensus_conflicts_test.go
jkrvivian: It's commented during debugging. :stuck_out_tongue:

tools/integration-tests/tester/framework/docker.go
```diff
@@ -129,7 +129,8 @@ func (d *DockerContainer) CreateGoShimmerPeer(config GoShimmerConfig) error {
 			fmt.Sprintf("--drng.xteam.committeeMembers="),
 			fmt.Sprintf("--drng.pollen.committeeMembers="),
 			fmt.Sprintf("--activity.broadcastIntervalSec=%d", config.ActivityInterval),
-			"--messageLayer.startSynced=true",
+			fmt.Sprintf("--messageLayer.StartSynced=%t", config.StartSynced),
```
Wollac, line 132:
```suggestion
			fmt.Sprintf("--messageLayer.startSynced=%t", config.StartSynced),
```

PR #1223, opened 2021-04-23
Add timestamp filter

capossele:
Adds a timed scheduler which keeps messages ordered by their timestamp and releases them according to it for messages with a timestamp in the future.


packages/tangle/parser.go
```diff
@@ -246,6 +256,8 @@ type MessageFilter interface {
 	OnAccept(callback func(msg *Message, peer *peer.Peer))
 	// OnAccept registers the given callback as the rejection function of the filter.
 	OnReject(callback func(msg *Message, err error, peer *peer.Peer))
+	// Close closes the filter.
+	Close()
```
luca-moser, line 260: You can simply embed `io.Closer`. Probably better as it also defines Close() to potentially return an error.

packages/tangle/parser.go
```diff
@@ -514,6 +529,80 @@ func (f *TransactionFilter) getRejectCallback() (result func(msg *Message, err e
...
+// Filter filters up on the given bytes and peer and calls the acceptance callback
+// if the input passes or the rejection callback if the input is rejected.
+func (f *TimestampFilter) Filter(msg *Message, p *peer.Peer) {
+	// if the message has a timestamp in the past or current
+	if clock.Since(msg.IssuingTime()) >= 0 {
+		// bypass the timed executor
+		f.getAcceptCallback()(msg, p)
+		return
+	}
+	// // add the message with timestamp in the future to the timed executor
```
luca-moser, line 574: double slash

luca-moser (approved):
Just to little remarks, otherwise lgtm.

PR #1236, opened 2021-04-30
Use "github.com/cockroachdb/errors" 

georgysavva:
Switch to "github.com/cockroachdb/errors" from the mix of standard "errors" and "golang.org/x/xerrors" libraries.
## Type of change 

### DIFF
--- a/docs/teamresources/guidelines.md
+++ b/docs/teamresources/guidelines.md
@@ -8,12 +8,12 @@
     defer func() {
         cerr := f.Close()
         if err == nil {
-            err = xerrors.Errorf("failed to close file: %w", cerr)
+            err = errors.Wrap(cerr, "failed to close file")
         }
     }()
 ```
-- Wrap errors with `xerrors.Errorf()` when returning them to the caller. It adds the stack trace and a custom message to the error. Without that information investigating an issue is very hard.
-- Use `xerrors.Is()` instead of direct errors comparison. This function unwraps errors recursively. [Example](https://github.com/iotaledger/goshimmer/pull/1113/files#diff-05fdc081489a8d5a61224d812f9bbd7bc77edf9769ed00d95ea024d2a44a699aL62).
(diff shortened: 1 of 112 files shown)


docs/teamresources/guidelines.md
```diff
@@ -46,7 +45,7 @@ var ErrSentinel = errors.New("identifiable error")
 // turn anonymous 3rd party errors into identifiable ones
 func SentinelErrFrom3rdParty() (result interface{}, err error)
     if result, err = 3rdPartyLibrary.DoSomething(); err != nil {
-        err = xerrors.Errorf("failed to do something (%v): %w", err, ErrSentinel)
+        err = errors.Errorf("failed to do something (%v): %w", err, ErrSentinel)
```
capossele, line 48: What's that red `%`?

capossele (approved):

docs/teamresources/guidelines.md
georgysavva: I guess it's because `guidelines.md` isn't a `*.go` file and this how GitHub displays it. In all other places in `*.go` files, there is no red thing.

docs/teamresources/guidelines.md
```diff
@@ -26,18 +26,17 @@
 
 ## Error handling
 
-We use the new error wrapping API and behavior introduced with Go 1.13 but we use the "golang.org/x/xerrors" drop-in replacement which follows the Go 2 design draft and which enables us to have a stack trace for every "wrapping" of the error.
+We use the new error wrapping API and behavior introduced with Go 1.13 but we use the "github.com/cockroachdb/errors" drop-in replacement which follows the Go 2 design draft and which enables us to have a stack trace for every "wrapping" of the error.
 
 Errors should always be wrapped an annotated with an additional message at each step. The following example shows how errors are wrapped and turned into the corresponding sentinel errors.
```
apenzk, line 31:
```suggestion
Errors should always be wrapped and annotated with an additional message at each step. The following example shows how errors are wrapped and turned into the corresponding sentinel errors.
```

jonastheis (approved):

PR #1237, opened 2021-04-30
global snapshot

apenzk:
# Description of change

- Snapshot that resets the network while keeping the last known UTXO, with API call to create snapshot (e.g. for docker). 
- Snapshot tool in tools

## Type of change

Enhancement

- Tested snapshotting from API and tools
- load snapshot in docker network
- check if wallet can load from old snapshot
- check if faucet continues from snapshot

## Change checklist


packages/ledgerstate/utxo_dag.go
```diff
@@ -247,26 +267,18 @@ func (u *UTXODAG) Consumers(outputID OutputID) (cachedConsumers CachedConsumers)
...
-		if stored {
+		if storedTx {
 			cached.Release()
 		}
-		for _, output := range essence.outputs {
+
+		for i, output := range record.Essence.outputs {
+			if !record.UnspentOutputs[i] {
+				continue
+			}
```
lzpap, line 281: Now that we have different types of outputs, the way we store `AddressOutputMapping` is outdated here on line 288.
One output might have more addresses mapped to it, take a look at this:
From `packages/tangle/booker.go` line 247-267:
```
for _, output := range transaction.Essence().Outputs() {
		switch output.Type() {
		case ledgerstate.AliasOutputType:
			castedOutput := output.(*ledgerstate.AliasOutput)
			// if it is an origin alias output, we don't have the aliasaddress from the parsed bytes.
			// that happens in utxodag output booking, so we calculate the alias address here
			b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(castedOutput.GetAliasAddress(), output.ID())
			b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(castedOutput.GetStateAddress(), output.ID())
			if !castedOutput.IsSelfGoverned() {
				b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(castedOutput.GetGoverningAddress(), output.ID())
			}
		case ledgerstate.ExtendedLockedOutputType:
			castedOutput := output.(*ledgerstate.ExtendedLockedOutput)
			if castedOutput.FallbackAddress() != nil {
				b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(castedOutput.FallbackAddress(), output.ID())
			}
			b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(output.Address(), output.ID())
		default:
			b.tangle.LedgerState.UTXODAG.StoreAddressOutputMapping(output.Address(), output.ID())
		}
	}
```

plugins/webapi/snapshot/plugin.go
```diff
@@ -0,0 +1,103 @@
...
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region DumpCurrentLedger ///////////////////////////////////////////////////////////////////////////////////////////////////
+
+// DumpCurrentLedger dumps a snapshot (all unspent UTXO and all of the access mana) from now.
+func DumpCurrentLedger(c echo.Context) (err error) {
+	snapshot := messagelayer.Tangle().LedgerState.SnapshotUTXO()
+
+	aMana, err := snapshotAccessMana()
```
lzpap, line 50: First we fetch the transactions that are older, than `utxoFetchTime-minAge`. Then we get the access mana at `utxoFetchTime+Delta`. But are we sure that this access mana state corresponds to utxo state in `snapshot`? I don't think so.
We are lacking the ablility to tell the exact access mana state with respect to a certain utxo state. It is only a problem if within that small time window `Delta` some big mana moves are happening. And even then, if `Delta` is reasonably small, it doesn't matte rthat much.
Another thing I would liek to discuss is the fact that we should save the base access manas in the snapshot as well, not just the effective value.

packages/ledgerstate/utxo_dag.go
apenzk: done

plugins/webapi/snapshot/plugin.go
apenzk: This is ok because we dont need to be that precise for the global snapshot tool for now. All that should happen is that for the most recent transactins they might not be part of the new network. if value messages become important in Nectar we can even inform the community that any spent outputs in the next 5 min or so might not be account for after the snapshot. 

Eventually we have to replace global snapshot by rolling snapshots anyways, which have to be treated quite different.

packages/tangle/ledgerstate.go
```diff
@@ -179,22 +192,87 @@ func (l *LedgerState) LoadSnapshot(snapshot *ledgerstate.Snapshot) {
 	if attachment != nil {
 		attachment.Release()
 	}
+	return
+}
+
+// SnapshotUTXO returns the UTXO snapshot, which is a list of transactions with unspent outputs.
+func (l *LedgerState) SnapshotUTXO() (snapshot *ledgerstate.Snapshot) {
+	// The following parameter should be much larger than the max allowed timestamp variation, and the required time for confirmation.
+	minAge := 15 * time.Second
```
capossele, line 201: Should we then adjust this value?

apenzk: updated to 120s

capossele (approved):

PR #1243, opened 2021-05-02
Feat/monitor sync status

piotrm50:
# Description of change

add plugin 
## Type of change

Manually with local ELK stash

## Change checklist


georgysavva (approved):
LGTM

packages/remotelogmetrics/events.go
```diff
@@ -0,0 +1,36 @@
+package remotelogmetrics
+
+import (
```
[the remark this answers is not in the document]
georgysavva: @piotrm50 you can run `golangci-lint run --fix` in the project root to automatically format all files and fix those `gofumpt` errors.

plugins/remotelogmetrics/plugin.go
```diff
@@ -0,0 +1,77 @@
...
+func Plugin() *node.Plugin {
+	pluginOnce.Do(func() {
+		plugin = node.NewPlugin(PluginName, node.Enabled, configure, run)
+	})
+	return plugin
+}
+
+func configure(plugin *node.Plugin) {
+	log = logger.NewLogger(PluginName)
+}
```
jonastheis, line 42: The plugin itself has a logger integrated. You can use it with `plugin.Log*`

packages/remotelogmetrics/events.go
```diff
@@ -0,0 +1,36 @@
...
+type CollectionLogEvents struct {
+	// SyncBeaconSyncChanged defines the local sync status change event based on sync beacon.
+	SyncBeaconSyncChanged *events.Event
+	// TangleTimeSyncChanged defines the local sync status change event based on tangle time.
+	TangleTimeSyncChanged *events.Event
+}
+
+func boolCaller(handler interface{}, params ...interface{}) {
+	handler.(func(SyncStatusChangedEvent))(params[0].(SyncStatusChangedEvent))
+}
```
jonastheis, line 18: This is not a `boolCaller` but `SyncStatusChangedEventCaller`

plugins/remotelogmetrics/plugin.go
```diff
@@ -0,0 +1,77 @@
...
+
+func sendSyncStatusChangedEvent() func(syncUpdate remotelogmetrics.SyncStatusChangedEvent) {
+	return func(syncUpdate remotelogmetrics.SyncStatusChangedEvent) {
+		err := remotelog.RemoteLogger().Send(syncUpdate)
+		if err != nil {
+			log.Errorw("Failed to send sync status changed record on sync change event.", "err", err)
+		}
+	}
+}
```
jonastheis, line 77: Why returning a func here instead of just having a func that receives the event as parameter and gets called directly?

jonastheis (changes requested):

jonastheis (commented):
In which index will the messages end up in ElasticSearch? Did you configure logstash accordingly and/or does it create an index automatically now based on the type?

packages/remotelogmetrics/events.go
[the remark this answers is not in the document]
piotrm50: Oh, COOL! I did not know that!

plugins/remotelogmetrics/plugin.go
piotrm50: Good point :)

piotrm50:
> In which index will the messages end up in ElasticSearch? Did you configure logstash accordingly and/or does it create an index automatically now based on the type?

It is configured in the other PR to create new index based on the type, however it is still solved using `if-else` currently because I also assign lifecycle policy by default there and then it's not possible to parametrize index name.

plugins/remotelogmetrics/plugin.go
jonastheis: You can use the integrated logger and simplify the code a bit more.

plugins/remotelogmetrics/plugin.go
```diff
@@ -0,0 +1,79 @@
...
+
+func run(plugin *node.Plugin) {
+	remotelogmetrics.Events().SyncBeaconSyncChanged.Attach(events.NewClosure(func(syncUpdate remotelogmetrics.SyncStatusChangedEvent) {
+		isSyncBeaconSynced.Store(syncUpdate.CurrentStatus)
+	}))
+	remotelogmetrics.Events().TangleTimeSyncChanged.Attach(events.NewClosure(func(syncUpdate remotelogmetrics.SyncStatusChangedEvent) {
+		isTangleTimeSynced.Store(syncUpdate.CurrentStatus)
+	}))
+	remotelogmetrics.Events().TangleTimeSyncChanged.Attach(events.NewClosure(sendSyncStatusChangedEvent()))
+	remotelogmetrics.Events().SyncBeaconSyncChanged.Attach(events.NewClosure(sendSyncStatusChangedEvent()))
```
jonastheis, line 55: Maybe move the setup of the events to the `configure` func.

jonastheis (changes requested):
Please address the comments and the linter warnings.

jonastheis (approved):

PR #1248, opened 2021-05-04
Add TXStream plugin

dessaya:
The TXStream plugin exposes a TCP port for clients to connect and subscribe to real-time
notifications on confirmed transactions targeted to specific addresses. It is
used by Wasp nodes to be notified about incoming requests and updates on a
given chain, but it provides a generic API, not tied to ISCP or Wasp.


lzpap (approved):
Apart from the golangci comments, it looks good to me.

lunfardo314 (commented):
Just needs fixing automated comments

PR #1252, opened 2021-05-06
Orderer after scheduler

capossele:


packages/tangle/orderer.go
```diff
@@ -0,0 +1,147 @@
...
+				if _, exists := o.parentsMap[bookedMessage]; !exists {
+					continue
+				}
+
+				for _, childID := range o.parentsMap[bookedMessage] {
+					o.tryToSchedule(childID)
+				}
+				delete(o.parentsMap, bookedMessage)
+			default:
+			}
```
jkrvivian, line 75: I guess this should be removed? there's the same behaviour in the select below

capossele: No that is done on purpose to always give priority to the code in the first select

acha-bill (approved):
Apart from that `orderer.Shutdown()` is unsed, (supposed to be in `tangle.Shutdown()`, this LGTM.

We need this ASAP. :)

PR #1259, opened 2021-05-07
Add activity plugin

capossele:


plugins/activity/plugin.go
```diff
@@ -0,0 +1,91 @@
...
+		plugin = node.NewPlugin(PluginName, node.Disabled, configure, run)
+	})
+	return plugin
+}
+
+// configure events
+func configure(_ *node.Plugin) {
+	log = logger.NewLogger(PluginName)
+
+	log.Infof("starting node with activity plugin")
```
jonastheis, line 42: You can get rid of the logger as every plugin has an logger integrated. `plugin.LogInfo` or something should do it.

jonastheis (approved):

PR #1272, opened 2021-05-10
Remove syncbeacon (Finally 🥳 )

capossele:


plugins/messagelayer/plugin.go
```diff
@@ -89,6 +89,9 @@ func configure(plugin *node.Plugin) {
 
 func run(*node.Plugin) {
 	if err := daemon.BackgroundWorker("Tangle", func(shutdownSignal <-chan struct{}) {
+		if Parameters.StartSynced {
+			Tangle().TimeManager.Synced()
+		}
```
jonastheis, line 94: What does it do? Can't that just be deleted?

jonastheis (approved):

PR #1279, opened 2021-05-10
Add dRNG remotelog

capossele:


piotrm50 (approved):
I simplified some things to how I envisioned it should look :)

PR #1301, opened 2021-05-15
Client-side Faucet Pow

jonastheis:


plugins/dashboard/faucet_routes.go
```diff
@@ -1,77 +0,0 @@
-package dashboard
```
lzpap, line 1: @jonastheis do we want to get rid of the faucet request functionality all together from the dashboard? Shouldn't we delete the Faucet page as well then? As I see it is only the dashboard local faucet api that got removed.

plugins/webapi/faucet/plugin.go
```diff
@@ -82,10 +68,7 @@ func requestFunds(c echo.Context) error {
 		}
 	}
 
-	faucetPayload, err := faucet.NewRequest(addr, config.Node().Int(faucet.CfgFaucetPoWDifficulty), accessManaPledgeID, consensusManaPledgeID)
-	if err != nil {
-		return c.JSON(http.StatusBadRequest, jsonmodels.FaucetResponse{Error: err.Error()})
-	}
+	faucetPayload := faucet.NewRequest(addr, accessManaPledgeID, consensusManaPledgeID, request.Nonce)
```
lzpap, line 71: can we check that `faucetPayload` is valid with respect to `request.Nonce`? Otherwise there is no point in submitting it to the tangle, no?

capossele (approved):

PR #1304, opened 2021-05-17
Proper Output Consolidation In Wallet Lib

lzpap:
# Description of change

 - Wallet lib (and cli wallet) can consolidate any number of outputs sitting on wallet addresses.
 - In the unlikely scenario that even after the first try, there are still more, than `MaxInputCount` outputs on wallet addresses, just run the command again.
 - Regular wallet functions fail with too many inputs, and notify users to consolidate funds.


client/wallet/wallet.go
```diff
@@ -181,59 +187,68 @@ func (wallet *Wallet) ConsolidateFunds(options ...consolidateoptions.Consolidate
...
-	}
-	if consolidateOptions.WaitForConfirmation {
-		err = wallet.WaitForTxConfirmation(tx.ID())
+		wallet.markOutputsAndAddressesSpent(consumedOutputs)
+		err = wallet.connector.SendTransaction(tx)
+		if err != nil {
+			return nil, err
+		}
+		if consolidateOptions.WaitForConfirmation {
+			err = wallet.WaitForTxConfirmation(tx.ID())
```
capossele, line 248:
```suggestion
			err = wallet.WaitForTxConfirmation(tx.ID())
			if err != nil {
				return nil, err
			}
```

capossele: Double check that this is what should happen in case of error.

client/wallet/output.go
```diff
@@ -167,3 +167,43 @@ func (o OutputsByAddressAndOutputID) ToLedgerStateOutputs() ledgerstate.Outputs
...
+
+// SplitIntoChunksOfMaxInputCount splits the mapping into chunks that contain at most ledgerstate.MaxInputCount outputs.
+func (o OutputsByAddressAndOutputID) SplitIntoChunksOfMaxInputCount() []OutputsByAddressAndOutputID {
+	outputCount := o.OutputCount()
+	if outputCount <= ledgerstate.MaxInputCount {
+		// there is no need to split
+		return []OutputsByAddressAndOutputID{o}
+	}
+	result := make([]OutputsByAddressAndOutputID, outputCount/ledgerstate.MaxInputCount+1)
+	for i, _ := range result {
```
capossele, line 190:
```suggestion
	for i := range result {
```

client/wallet/wallet.go
```diff
@@ -181,59 +187,68 @@ func (wallet *Wallet) ConsolidateFunds(options ...consolidateoptions.Consolidate
...
-	if err != nil {
-		return nil, err
-	}
-	if !ok {
-		return nil, errors.Errorf("created transaction is invalid: %s", tx.String())
-	}
+		// check tx validity (balances, unlock blocks)
+		ok, err := checkBalancesAndUnlocks(inputsAsOutputsInOrder, tx)
+		if err != nil {
+			return nil, err
```
capossele, line 236:
```suggestion
		ok, e := checkBalancesAndUnlocks(inputsAsOutputsInOrder, tx)
		if e != nil {
			return nil, e
```

capossele (changes requested):

client/wallet/wallet.go
lzpap: I've refactored it.

capossele (approved):

PR #1317, opened 2021-05-20
Feat/pow update parents

piotrm50:
# Description of change

PoW in IssuePayload updates timestamp and parents while looking for nonce.

## Type of change

Manual testing by running spammer and faucet requests.

## Change checklist

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -110,7 +110,9 @@
   "pow": {
     "difficulty": 22,
     "numThreads": 1,
-    "timeout": "1m"
+    "timeout": "1m",
+    "parentsRefreshInterval": "1s"
+
   },
   "profiling": {
     "bindAddress": "127.0.0.1:6061"
--- a/config.default.json
+++ b/config.default.json
@@ -110,8 +110,7 @@
   "pow": {
     "difficulty": 22,
     "numThreads": 1,
-    "timeout": "1m",
-    "parentsRefreshInterval": "1s"
+    "timeout": "1m"
 
   },
   "profiling": {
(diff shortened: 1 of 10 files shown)


capossele (approved):

capossele (approved):

config.default.json
```diff
@@ -111,6 +111,7 @@
     "difficulty": 22,
     "numThreads": 1,
     "timeout": "1m"
+
```
jonastheis, line 114: Should this be added here? `"parentsRefreshInterval": "300ms"`
Or do we just rely on the default value?

jonastheis (approved):

tools/docker-network/config.docker.json
```diff
@@ -70,7 +70,8 @@
   "pow": {
     "difficulty": 2,
     "numThreads": 1,
-    "timeout": "10s"
+    "timeout": "10s",
```
apenzk, line 73: lets use 60s as default

packages/tangle/messagefactory_test.go
```diff
@@ -22,6 +23,7 @@ import (
 
 const (
 	targetPOW     = 10
+	powTimeout    = 10 * time.Second
```
apenzk, line 26: lets use 60s here. with this new change we dont have problems with timestamps getting to old anymore

tools/docker-network/config.docker.json
capossele: here 10s is fine as it is for the docker-network where we set the difficulty to be only 2

packages/tangle/messagefactory_test.go
capossele: here 10s is fine as it is for the test where we set the difficulty to be only 2

config.default.json
capossele: Yeah, we thought that by specifying that in the default value was easier so we don't care to set that here

PR #1360, opened 2021-05-26
Simplify manual peering API

georgysavva:
This PR hides the complexity of the `peer.Peer` structure from the end-user of the manual peering module. Now the web API request to add known peers looks like that:
```json
[
    {
      "publicKey": "EYsaGXnUVA9aTYL9FwYEvoQ8d1HCJveQVL7vogu6pqCP",
      "address": "172.19.0.3:14666"
    }
]
```
Instead of the previous one:
```json
[
    {
      "publicKey": "EYsaGXnUVA9aTYL9FwYEvoQ8d1HCJveQVL7vogu6pqCP",
      "ip": "172.19.0.3",
      "services": {
        "peering":{
          "network":"TCP",
          "port":14626
        },
        "gossip": {
          "network": "TCP",
          "port": 14666
        }
      }
    }
]
```
We no longer require the "peering" service port, because it's not needed in both manual peering and gossip. And it might have confused the user.

### DIFF
--- a/packages/manualpeering/manualpeering.go
+++ b/packages/manualpeering/manualpeering.go
@@ -2,10 +2,13 @@ package manualpeering
 
 import (
 	"context"
+	"encoding/json"
+	"net"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/iotaledger/hive.go/autopeering/peer/service"
 	"github.com/iotaledger/hive.go/typeutils"
 
 	"github.com/iotaledger/goshimmer/packages/gossip"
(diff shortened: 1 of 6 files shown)


capossele (approved):

jonastheis (approved):

packages/manualpeering/manualpeering.go
```diff
@@ -342,18 +396,18 @@ func (m *Manager) changeNeighborStatus(neighbor *gossip.Neighbor, connStatus Con
...
 }
 
-func (m *Manager) connectionDirection(p *peer.Peer) (ConnectionDirection, error) {
-	localPK := m.local.PublicKey().String()
-	peerPK := p.PublicKey().String()
-	if localPK < peerPK {
+func (m *Manager) connectionDirection(peerPK ed25519.PublicKey) (ConnectionDirection, error) {
+	localPKStr := m.local.PublicKey().String()
+	peerPKStr := peerPK.String()
+	if localPKStr < peerPKStr {
```
jonastheis, line 402: I'm not a big fan of using the string method to compare as this could have unwanted side effects if we change it. Maybe better to use `bytes.Compare(pk1.Bytes(), pk2.Bytes()`

georgysavva: Yeah, you are right, let's go with bytes

PR #1361, opened 2021-05-26
[docs] create and send txs

acha-bill:


docs/tutorials/send_transaction.md
```diff
@@ -0,0 +1,274 @@
...
+The simplest and easiest way for creating transaction is to use ready solution, such us GUI wallets: [pollen-wallet](https://github.com/iotaledger/pollen-wallet/tree/master) and [Dr-Electron ElectricShimmer](https://github.com/Dr-Electron/ElectricShimmer)
+or command line wallet [Command Line Wallet](./wallet.md). However, there is also an option to create a transaction directly with Go client library, which will be main focus of this tutorial.
+
+For the ready code examples you can go directly to [Code examples](send_transaction.md#ready-examples).
+
+## Funds
+To create a transaction, firstly we need to be in possession of tokens. We can receive them from other network users or request them from the faucet. For more details on how to request funds, see [this](./request_funds.md) tutorial.
+
+## Preparing transaction
+A transaction is build from two main parts: a transaction essence, and an unlock block. The transaction essence contains all crucial information on how many and where the funds should be sent. The unlock block makes sure that only the owner of the destination address will be able to spend outputs of this transaction.
```
capossele, line 11:
```suggestion
A transaction is built from two parts: a transaction essence, and the unlock blocks. The transaction essence contains, among other information, the amount, the origin and where the funds should be sent. The unlock block makes sure that only the owner of the funds being transferred is allowed to successfully perform this transaction.
```

docs/tutorials/send_transaction.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Seed
+In order to send funds we need to have private key that can unlock our funds. If you want to use an existing seed from one of your wallets, just use the backup seed showed during a wallet creation. With this, we can decode the string with `base58` library and create the `seed.Seed` instance. That will allow us to retrieve wallet addresses (`mySeed.Address()`) and their corresponding private and public keys (`mySeed.KeyPair()`).
```
capossele, line 14:
```suggestion
In order to send funds we need to have a private key that can be used to prove that we own the funds and consequently unlock them. If you want to use an existing seed from one of your wallets, just use the backup seed showed during a wallet creation. With this, we can decode the string with the `base58` library and create the `seed.Seed` instance. That will allow us to retrieve the wallet addresses (`mySeed.Address()`) and the corresponding private and public keys (`mySeed.KeyPair()`).
```

docs/tutorials/send_transaction.md
```diff
@@ -0,0 +1,274 @@
...
+```Go
+balance := ledgerstate.NewColoredBalances(map[ledgerstate.Color]uint64{
+							ledgerstate.ColorIOTA: uint64(100),
+						})
+outputs := ledgerstate.NewOutputs(ledgerstate.NewSigLockedColoredOutput(balance, destAddr.Address()))
+```
+The same as in case of inputs we need to adapt it with `ledgerstate.NewOutputs()` before passing to the `NewTransactionEssence` function.
+
+### Signing transaction
+After preparing the transaction essence there is a time to sign it and put the signature to the unlock block part of the transaction.
```
capossele, line 117:
```suggestion
After preparing the transaction essence, we should sign it and put the signature to the unlock block part of the transaction.
```

capossele (changes requested):

docs/tutorials/send_transaction.md
```diff
@@ -0,0 +1,274 @@
...
+```
+Putting it all together, now we are able to create transaction with previously created transaction essence and adapted unlock block.
+
+```Go
+tx := ledgerstate.NewTransaction(txEssence, ledgerstate.UnlockBlocks{unlockBlock})
+```
+
+## Sending transaction
+There are two web API methods that allows us to send the transaction:
+PostTransaction() and IssuePayload(). The second one is more general method that sends the attached payload. We are going to use the first one that will additionally checks the transaction validity before issuing and wait with sending the response until the message is booked.
```
jkrvivian, line 138:
```suggestion
`PostTransaction()` and `IssuePayload()`. The second one is a more general method that sends the attached payload. We are going to use the first one that will additionally check the transaction validity before issuing and wait with sending the response until the message is booked.
```

docs/tutorials/send_transaction.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+#### Post via web API
+First, get the transaction bytes.
+```go
+	// build tx from previous step
+	tx, err := buildTransaction()
+	if err != nil {
+		return
+	}
```
jkrvivian, line 259: Adjust indent.

jkrvivian (changes requested):

jkrvivian (approved):

capossele (approved):

PR #1372, opened 2021-05-28
Manual peering tutorial and docs

georgysavva:
Write a tutorial for manual peering module and document web API.


capossele (approved):

docs/tutorials/manual_peering.md
```diff
@@ -0,0 +1,49 @@
+# Manual Peering
+
+With manual peering, it's possible to connect only with a static list of known peers that the user has provided
+and don't depend on the autopeering. There are two ways to work with the list of known peers of the node:
```
jonastheis, line 4:
```suggestion
Manual peering enables node operators to exchange their nodes' information and let them peer with each other, orthogonal to autopeering. It can be an additional protection against eclipse attacks as the manual peering is completely in the hands of the node operator based on real world trust. Furthermore, it allows to operate nodes without exposing their IP address to the network.
There are two ways to configure the list of known peers of the node:
```

jonastheis (approved):

PR #1398, opened 2021-06-01
[Docs] protocol specification

capossele:


apenzk (approved):
Text looks good. just a few grammar changes and splitting of paras
I have changed Goshimmer to GoShimmer where it was save to do so (i.e. only in the docs or in comments)

PR #1408, opened 2021-06-02
Fix overlocking and sleeping issue

jonastheis:


plugins/webapi/ledgerstate/plugin.go
```diff
@@ -433,9 +433,6 @@ var (
 
 // PostTransaction sends a transaction.
 func PostTransaction(c echo.Context) error {
-	sendTxMu.Lock()
```
jkrvivian, line 436: Remove the `sendTxMu` declaration above as well.

PR #1414, opened 2021-06-07
feat: Move scheduling after booking

jkrvivian:
# Description of change

* Move the scheduler after the booker
* Remove FIFO-scheduler
* Avoid deleting discarded messages from the storage
  This causes panic when inheriting branches in the booker for its children.
* Avoid scheduling old messages (> 30 mins)
  To avoid exploding node queue sizes when syncing
* Make booker running on 1 single go-routine, since it is not thread-safe yet.

Data flow:
```
solidifier -> booker -> scheduler -> orderer -> gossip                                             
                     -> consensus -> tip manager
                     -> chat
                     -> approval weight
```

## Type of change

- Breaking change

## Change checklist


packages/tangle/scheduler.go
```diff
@@ -104,7 +105,16 @@ func (s *Scheduler) Shutdown() {
 }
 
 // Setup sets up the behavior of the component by making it attach to the relevant events of the other components.
-func (s *Scheduler) Setup() {}
+func (s *Scheduler) Setup() {
+	// pass booked messages to the scheduler
+	s.tangle.Booker.Events.MessageBooked.Attach(events.NewClosure(func(messageID MessageID) {
+		if err := s.SubmitAndReady(messageID); err != nil {
+			s.Events.Error.Trigger(errors.Errorf("failed to submit to scheduler: %w", err))
```
Wollac, line 112: An error might be a bit strong here. There are 3 possibilities:

1. `ErrInboxExceeded` this is kind of expected behavior and we are triggering `NodeBlacklisted` anyways. So maybe this isn't even worth to log.
2. `ErrInsufficientMana` will never happen as the min access mana is sufficiently large.
3. Any other error coming from the `ObjectStorage` etc. These are probably actual errors.

packages/tangle/tangle.go
```diff
@@ -52,7 +49,6 @@ type Tangle struct {
 	Events                *Events
 
 	setupParserOnce sync.Once
-	fifoScheduling  typeutils.AtomicBool
 	shutdownSignal  chan struct{}
```
Wollac, line 52: Is `shutdownSignal` still used? It was introduced to cancel the `sleep` in the scheduler switching, which has been removed as part of this PR.
Of course, it also does not hurt either to have this in for future expansions.

Wollac:
Does this new data-flow have any implications on the tip selection:
- Do we need to make sure to only select gossiped tips?
- Could this be causing issues right after syncing when nothing was gossiped?

packages/tangle/booker.go
```diff
@@ -51,7 +51,7 @@ func NewBooker(tangle *Tangle) (messageBooker *Booker) {
 
 // Setup sets up the behavior of the component by making it attach to the relevant events of other components.
 func (b *Booker) Setup() {
-	b.tangle.Orderer.Events.MessageOrdered.Attach(events.NewClosure(func(messageID MessageID) {
+	b.tangle.Solidifier.Events.MessageSolid.Attach(events.NewClosure(func(messageID MessageID) {
```
capossele, line 54: As we typically have multiple go-routines that deal with solidification, wouldn't these go-routines then also continue the execution of the booking?
I think the flow should be interrupted here and have the booker running with only 1 single go-routine. WDYT?

packages/tangle/tangle.go
```diff
@@ -125,6 +117,13 @@ func (t *Tangle) Setup() {
 	t.ConsensusManager.Setup()
 	t.TipManager.Setup()
 
+	// increase scheduler rate if the node is not synced at start
+	if !t.Synced() {
+		rate := t.Scheduler.Rate()
+		rate -= rate / 2 // 50% increase
+		t.Scheduler.SetRate(rate)
+	}
```
capossele, line 125: Considering that we bypass the scheduler for all the messages older than 5 minutes, do we still need this at all?

capossele (approved):
Just a minor change/suggestions, otherwise seems solid

packages/tangle/tangle.go
jkrvivian: Let's remove it!

PR #1419, opened 2021-06-08
Enable Github Actions Caching

georgysavva:
Fixes: https://github.com/iotaledger/goshimmer/issues/890


georgysavva:
After running several benchmarks it turned out that docker layer caching doesn't give any performance benefit due to the fact that we can not cache the Go build. So we should get back this when they fix that in https://github.com/moby/buildkit/issues/1512

Let's merge this PR anyway to at least have a working example with Github Actions cache for docker :)

.github/workflows/docker-develop.yml
```diff
@@ -14,13 +17,40 @@ jobs:
...
+        uses: docker/login-action@v1.9.0
+        with:
+          username: '${{ secrets.IOTALEDGER_HUB_DOCKER_LOGIN }}'
+          password: '${{ secrets.IOTALEDGER_HUB_DOCKER_PASSWORD }}'
+
+      - name: Publish to Docker Hub
+        uses: docker/build-push-action@v2.5.0
+        with:
+          build-args: |
+            mount='id=go_build_cache_mount,type=bind,source=/tmp/foo'
```
jonastheis, line 44: `source=/tmp/foo` does it really not matter? Shouldn't that be maybe `/tmp/.buildx-cache`?

.github/workflows/docker-develop.yml
```diff
@@ -6,6 +6,9 @@ on:
       - develop
     paths-ignore:
       - 'docs/**'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
```
jonastheis, line 11: Why is that added? Was probably just for testing, no?

.github/workflows/docker-develop.yml
georgysavva: Sorry, debug code. I will clean this up

.github/workflows/docker-develop.yml
georgysavva: Sorry, debug code. I will clean this up

jonastheis (approved):

capossele (approved):

PR #1420, opened 2021-06-08
refactor: Remove mana related settings from integration tests

jkrvivian:
# Description of change

* Remove `mana` from Goshimmer config, since mana is enabled by default
* Replace `CreateNetworkWithMana` with `CreateNetwork`. bc each node has 1m to issue messages, no need to wait for faucet pledging mana.

## Type of change

- Enhancement

## Change checklist

### DIFF
--- a/tools/integration-tests/tester/framework/drngnetwork.go
+++ b/tools/integration-tests/tester/framework/drngnetwork.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/docker/docker/client"
 	"github.com/drand/drand/net"
-	walletseed "github.com/iotaledger/goshimmer/client/wallet/packages/seed"
 	"github.com/iotaledger/hive.go/crypto/ed25519"
 	"github.com/iotaledger/hive.go/identity"
 )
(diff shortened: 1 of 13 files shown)


tools/integration-tests/tester/framework/drngnetwork.go
```diff
@@ -49,15 +48,6 @@ func (n *DRNGNetwork) CreatePeer(c GoShimmerConfig, publicKey ed25519.PublicKey)
 		config.FPCTotalRoundsFinalization = 10
 	}
 
-	// create wallet
```
apenzk, line 52: so this wallet is never created for the faucet? for what was the c.Faucet option then actually?

jkrvivian: Yes, in this drng scenario, nodes don't send funds around, so neither the faucet nor a wallet is needed. 
`c.Faucet` means we need one guy in the network to be the faucet :smile:

tools/integration-tests/tester/framework/framework.go
```diff
@@ -294,21 +287,13 @@ func (f *Framework) CreateDRNGNetwork(name string, members, peers int) (*DRNGNet
...
 		DRNGCommittee: drngCommittee,
-		Mana:          true,
 		StartSynced:   true,
 	}
 
 	// create peers/GoShimmer nodes
 	for i := 0; i < peers; i++ {
 		config.Seed = privKeys[i].Seed().String()
-		if _, e := drng.CreatePeer(func() GoShimmerConfig {
-			if i == 0 {
```
apenzk, line 305: the faucetConfig is removed. Why is that - is it no longer needed?

jkrvivian: Because we don't need a faucet sending funds (usually we take `peers[0]` to be the faucet), thus `drng.CreatePeer` will not check `config.Faucet` to specifically set up the faucet wallet.
There's no need to set the `config.Faucet` here, so just take the config above works.

apenzk (approved):

acha-bill (approved):

jonastheis (approved):

PR #1424, opened 2021-06-09
Improve common integration test

jonastheis:
# Description of change
- delete `message` integration test because persistence is also tested in the `common` test
- simplify test to not blindly wait anymore but instead wait for messages to become available on nodes with a max timeout

## Type of change
Enhancement


jkrvivian (changes requested):

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -164,6 +179,49 @@ func CheckForMessageIDs(t *testing.T, peers []*framework.Peer, messageIDs map[st
 	}
 }
 
+// AwaitMessageAvailability awaits until the given message IDs become available on all given peers or
+// the max duration is reached. Returns a map of missing messages per peer. An error is returned if at least
+// one peer does not have all specified messages available.
+func AwaitMessageAvailability(peers []*framework.Peer, messageIDs map[string]DataMessageSent, waitFor time.Duration) (missing map[identity.ID]map[string]types.Empty, err error) {
+	s := time.Now()
+	missing = map[identity.ID]map[string]types.Empty{}
```
jkrvivian, line 187: Need a mutex to protect the concurrent read/write below.
I just noticed that :stuck_out_tongue:

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -164,6 +179,49 @@ func CheckForMessageIDs(t *testing.T, peers []*framework.Peer, messageIDs map[st
...
+
+	for ; time.Since(s) < waitFor; time.Sleep(500 * time.Millisecond) {
+		var wg sync.WaitGroup
+		wg.Add(len(peers))
+
+		for _, p := range peers {
+			go func(p *framework.Peer) {
+				defer wg.Done()
```
jkrvivian, line 195: We can add a check here to avoid requesting messages if a node has them all already. But still need a lock there.
```suggestion
			go func(p *framework.Peer) {
				defer wg.Done()
				
				// TODO: lock !
				if _, has := missing[p.ID()]; !has {
				    return
				} 
```

tools/integration-tests/tester/tests/testutil.go
jonastheis: Do we really need the lock? All threads operate on separate elements of the map. If so we only need it for checking the first level, ie `map[identity.ID]` I think

jonastheis: I've added the lock and the other optimization you suggested as well

acha-bill (approved):
LGTM

tools/integration-tests/tester/tests/testutil.go
jkrvivian: I deleted the other comment simply because it will skip the check in the first round, and I didn't come up with other solution.
But just add the additional check on `i` will work, just realize that from your fix! :clinking_glasses:

jkrvivian (approved):

Issue #1109, opened 2021-03-22
Add global toggle to enable/disable cache time


capossele:
Add a global toggle to enable/disable cache time of the object storage. When `disabled` all storages that are created should be used with a cache time of 0 (cache disabled). The default should be `enabled`. 

Make use of this new functionality:
- [ ] in unit tests
- [ ] add a flag (eg `--database.disableCacheTime`) and set it by default for the integration tests
- [ ] possibly set the flag also for the Docker network?

*EDIT*:
flagi `---globalCacheTime` and any non-negative value will set all the cache to that value. A negative value will keep defaults.

GalRogozinski:
Change in hive.go so that we can configure it for any unit test

GalRogozinski:
#### Tasks
- [ ] add a consts file that will include all current magic numbers for cache
- [ ] Add an `-overrideCacheTime` flag for database plugin
- [ ] Add a utility function that sets the cacheTime according to hardcoded consts, and the flag.
- [ ] Add a TestMain for each package that will point to a common `setup` and `teardown` methods. Make sure `setup` will also load the `-overrideCacheTime` with a 0 default.
- [ ] Ensure integration tests will run with a 0 cache time by default

PR #1440, opened 2021-06-13
Add test mains and test helper

GalRogozinski:
# Description of change

Add a main_test.go to every folder in packages that points to global `Setup` and `Tearup`.
This provides a good framework for current tests.

The motivation for this PR was to easily inject a global cache time for #1109 

## Type of change

- Documentation Fix

Saw in debug that test_main.go is being invoked

## Change checklist

### DIFF
--- a/packages/clock/main_test.go
+++ b/packages/clock/main_test.go
@@ -1,0 +1,16 @@
+package clock
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/iotaledger/goshimmer/packages/testhelper"
+)
+
+// TestMain will setup and teardown any unit test
+func TestMain(m *testing.M) {
+	GlobalSetup()
+	code := m.Run()
+	GlobalTeardown()
+	os.Exit(code)
+}
--- a/packages/clock/main_test.go
+++ b/packages/clock/main_test.go
@@ -4,7 +4,7 @@ import (
 	"os"
 	"testing"
 
-	. "github.com/iotaledger/goshimmer/packages/testhelper"
+	"github.com/iotaledger/goshimmer/packages/testhelper"
 )
 
 // TestMain will setup and teardown any unit test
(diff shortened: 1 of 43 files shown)


GalRogozinski:
When I tired to add a global function for the test I hitted a "circular dependency" error.
Since it seems to be too hard to utilize this idea I am closing the PR

GalRogozinski:
I decided to utilize a different approach that utilizes this PR

packages/clock/main_test.go
```diff
@@ -0,0 +1,16 @@
+package clock
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/iotaledger/goshimmer/packages/testhelper"
```
georgysavva, line 7: It's a good practice to avoid importing with ".":
```suggestion
	"github.com/iotaledger/goshimmer/packages/testhelper"
```

georgysavva (commented):
Could you please explain what those `GlobalSetup()` and `GlobalTeardown()` are going to do?

GalRogozinski:
@georgysavva 
in #1441 I made them load an env variable that tells the code we are running in test mode

packages/clock/main_test.go
GalRogozinski: question before I change:

We should never use `.`?
Can we automate this rule by the linter?

GalRogozinski: Should we add this also to https://goshimmer.docs.iota.org/teamresources/guidelines.html?
Can you give examples of exceptions? (I just thought that if there should be exceptions to the rule, they should be in`test` packages)

georgysavva: A few exceptions that I can come up with:
1. Indeed in test files, but in a different situation: when you have package "foo" and your unit tests are in the same directory but in package "foo_test" (good practice in Go), then you might want to import "foo" with "." to not add the prefix `foo.` to every function call.
2. When you have an internal package

But again, those exceptions are case by case and aren't critical, I would bother adding them as a rule to the guidelines.

GalRogozinski:
@georgysavva 
This PR is no longer a prerequisite of #1441 

But maybe it is still useful? If not I can close it

georgysavva:
I think we can close this one because that functionality is no longer needed

PR #1441, opened 2021-06-13
Global Cache Time for tests should be 0

### LINKED ISSUE #1109: Add global toggle to enable/disable cache time

GalRogozinski:
# Description of change

1. added --database.forceCacheTime that overrides default cache time unless flag is set to -1
2. Utilizes the cache in integration tests
3. unit tests now run with cache disabled

fixes #1109 

Should be merged after ~#1440~ and #1439 

## Type of change

Cache settings were ascertained in debug mode

## Change checklist


georgysavva (changes requested):
I think we approach the problem in reverse order. 
First of all, we shouldn't set an ENV variable at the beginning of every test. They should come from the environment. Because the environment uses them to configure the tests, not vise versa.
So if you want to disable caching you run tests like that:
```
GOSHIMMER_ENV=testing go test ./...
```
But more importantly, I don't think we need an ENV var for that at all. 

An inner package like "database" should know nothing about the environment it is running in. It shouldn't change its behavior depending on the current environment. In other words, it shouldn't have if-branches like that:
```go
	// if test just disable cache	
	if testhelper.IsTest() {		
	return objectstorage.CacheTime(0)
}
```
Instead, it should expose a way to configure itself, and the caller code (unit tests in our case) should be able to configure the package to its needs.

I see it like that:
1. The database package has a default cache time, so the caching is enabled by default
2. The database package allows setting a custom cache time during the instantiation via options. Where 0 or -1 means no caching.
3. We introduce `--database.globalCacheTime` CLI argument that integration tests and node operators will use to control the caching.
4. Unit tests will have a helper factory function that will instantiate a database instance with disable caching and all unit tests that benefit from a non-caching database will utilize it.
@capossele @jonastheis

GalRogozinski:
@georgysavva 

1. currently I think the default cache time in hive is 0... but we change it to hard-coded defaults per use-case. You want to change this?
2.  I agreed with @jonastheis that -1 will mean that we leave defaults alone. Because we didn't want to chage them or make them configurable.
3. Aite, this is how it currently works :-)
4. I definitely agree with and I actually tried to implement as you described...
The problem was how to call this helper factory and I tried to call it from `GlobalSetup`
The problem was that then I hit a circular dependency exception from the database package,,,,

So I am not sure how to go about it... so this is why I did this hacky solution (I didn't like it either),,,

Another option I thought about (and I started doing at the beginning before dumping it) is to always pass caching configuration via injections... but this will be way more code changes

georgysavva:
1. I don't want to change any default behavior 
2. If we don't need to configure the cache time it's fine. Then we just need to pass the only value -1.
3. Cool!
4. You had a circular dependency problem because you made the database package call the test helper package to check the environment. This should be avoided. With my solution, the database package doesn't do that. Only the test helper package knows about the database. 

> Another option I thought about (and I started doing at the beginning before dumping it) is to always pass caching configuration via injections... but this will be way more code changes

Yeah, this is what I actually meant. But I don't think there should be a lot of code changes since the runtime code doesn't need to change they use the default. You need to change only tests but we can refactor them to use the new helper factory functionality.

packages/testhelper/testhelper.go
```diff
@@ -1,27 +1,29 @@
 package testhelper
 
 import (
-	"fmt"
-	"os"
+	"flag"
```
GalRogozinski, line 4: using `spf13/pflag` doesn't work well in tests
this has to do with go using `flag` internally

plugins/database/parameters.go
```diff
@@ -13,8 +14,16 @@ const (
 	CfgDatabaseDirty = "database.dirty"
 )
 
+// Parameters contains configuration parameters used by the storage layer.
+var Parameters = struct {
+	// ForceCacheTime  is a new global cache time in seconds for object storage.
+	ForceCacheTime int `default:"-1" usage:"number of seconds all objects should remain in memory. -1 means use defaults"`
```
georgysavva, line 20: We can use `time.Duration` Instead of an int, the configuration framework will parse it for us.

It's worth updating the description then: 
`default:"-1s" usage:"interval of time objects should remain in memory. Zero time means no caching, negative value means use defaults"`

plugins/database/plugin.go
```diff
@@ -103,6 +103,10 @@ func configure(_ *node.Plugin) {
 	runDatabaseGC()
 }
 
+func run(*node.Plugin) {
+	// placeholder
+}
+
```
georgysavva, line 109: No need to keep unused code

packages/tangle/tangle.go
```diff
@@ -292,6 +294,7 @@ type Options struct {
 	WeightProvider               WeightProvider
 	SyncTimeWindow               time.Duration
 	StartSynced                  bool
+	ForceCacheTime               database.CacheTimeProvider
```
georgysavva, line 297: Name the field as the type `CacheTimeProvider` because `ForceCacheTime` is an implementation detail hidden inside the `database.CacheTimeProvider` type and in the future, we might apply different logic for cache management, but Tangle should know nothing about that.

Also, use a pointer as in any other place. I understand your intention to use value to gracefully handle the case there is no `CacheTimeProvider` passed, but it's very implicit and hard to reason about. Instead, we should use a _default_ `CacheTimeProvider` instance when no one passed such option as it's done for other options in the Tangle.

packages/tangle/tangle.go
```diff
@@ -381,6 +384,13 @@ func StartSynced(startSynced bool) Option {
 	}
 }
 
+// ForceCacheTime is an Option for the Tangle that allows to override hard coded cache time.
+func ForceCacheTime(forceCacheTime int) Option {
+	return func(options *Options) {
+		options.ForceCacheTime = *database.NewCacheTimeProvider(time.Duration(forceCacheTime))
+	}
+}
+
```
georgysavva, line 393: Rename option to CacheTimeProvider and accept a `*database.NewCacheTimeProvider` instance instead of an int, when you initialize the `*Options` use a default `CacheTimeProvider` instance with cache disabled (forceCacheTime=-1)

plugins/messagelayer/plugin.go
```diff
@@ -155,6 +155,7 @@ func Tangle() *tangle.Tangle {
 			}),
 			tangle.SyncTimeWindow(Parameters.TangleTimeWindow),
 			tangle.StartSynced(Parameters.StartSynced),
+			tangle.ForceCacheTime(database.Parameters.ForceCacheTime),
```
georgysavva, line 158: Instead of passing an int, initialize the `database.CacheTimeProvider` instance in a plugin either here or create a singleton in `plugins/database` and import it.

georgysavva (changes requested):
We are close, it looks very good already. There are few minor changes left!

packages/ledgerstate/objectstorage.go
```diff
@@ -38,66 +40,100 @@ const (
...
+		ConsumerPartitionKeys,
+		cacheProvider.CacheTime(consumerCacheTime),
+		objectstorage.LeakDetectionEnabled(false),
+	}
+
+	addressOutputMappingStorageOptions = []objectstorage.Option{
+		cacheProvider.CacheTime(addressCacheTime),
+		objectstorage.PartitionKey(AddressLength, OutputIDLength),
+		objectstorage.LeakDetectionEnabled(false),
+	}
```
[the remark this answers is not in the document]
GalRogozinski: question:
In go we call non-exported variable "global" still?  How would we call exported ones?

The main motivation for the change, as I understand, is for tests correct (even though you can argue it adds flexibility to the prod code)?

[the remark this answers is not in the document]
georgysavva: Sorry, I might misuse the term global variable, let's just call it a package scoped variable. Of course, when it's unexported the situation isn't that bad, but it doesn't solve the original problem: that you have a shared state and it's get mutated for no reason. It's not only related to tests, in production code multiple callers might call our public API concurrently and you will have a data race condition. So again, we should avoid such singleton-ish instances in packages.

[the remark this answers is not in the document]
GalRogozinski: yeah I agree
I initally mapped the concept of go package in my head to the concept of class from java (because this is what it looked like to me from the code)

but the way you think about it sounds better and more proper

plugins/database/plugin.go
GalRogozinski: This is from #1439
We must have this function so the database will configure itself on time

but maybe I will revert this merge and merge back with develop so it won't appear here

plugins/messagelayer/plugin.go
GalRogozinski: but this breaks the pattern...
I don't understand what's wrong with the current code

plugins/database/plugin.go
georgysavva: I thought we decided to close https://github.com/iotaledger/goshimmer/pull/1439?

plugins/messagelayer/plugin.go
georgysavva: Yes, theoretically, singletons are bad, but the whole plugins framework is built on top of them, and creating one more singleton for the `CacheTimeProvider` doesn't break the pattern, but it keeps the code consistent. When we are ready we will refactor the plugin framework and make it use the DI. 

When you pass an int to the `tangle.New()` it makes the function do one more thing: initialize `database.CacheTimeProvider` instance and it violates the single responsibility pattern.
It's better to initialize all Go objects at the beginning of the program and then pass them instead of raw values.

plugins/database/plugin.go
GalRogozinski: I think you got confused with #1440

#1439 is already merged and I think it is needed becuase we didn't configure on time..
Maybe there were more changes than what I did that are needed, but my goal was to do the minimum amount of changes to not break anything else

plugins/messagelayer/plugin.go
GalRogozinski: ok I agree

plugins/database/plugin.go
georgysavva: Ah I see, I am okay with that then

GalRogozinski: I merged develop to my branch to ensure and I saw all was merged well so I am not bothering with the revert

packages/tangle/tangle.go
```diff
@@ -292,6 +294,7 @@ type Options struct {
 	WeightProvider               WeightProvider
 	SyncTimeWindow               time.Duration
 	StartSynced                  bool
+	CacheTimeProvider            database.CacheTimeProvider
```
georgysavva, line 297: Use a pointer here. In `Tangle.configure()` assign a default cache time provider (with force time -1):
```go
// Configure modifies the configuration of the Tangle.
func (t *Tangle) Configure(options ...Option) {
	if t.Options == nil {
		t.Options = &Options{
			Store:                        mapdb.NewMapDB(),
			Identity:                     identity.GenerateLocalIdentity(),
			IncreaseMarkersIndexCallback: increaseMarkersIndexCallbackStrategy,
			CacheTimeProvider: database.NewCacheTimeProvider(-1 /* forceCacheTime */),
		}
	}

	for _, option := range options {
		option(t.Options)
	}

	if t.Options.ConsensusMechanism != nil {
		t.Options.ConsensusMechanism.Init(t)
	}
}
```

tools/integration-tests/tester/framework/docker.go
```diff
@@ -144,6 +145,7 @@ func (d *DockerContainer) CreateGoShimmerPeer(config GoShimmerConfig) error {
 			fmt.Sprintf("--mana.allowedConsensusPledge=%s", func() string {
 				return strings.Join(config.ManaAllowedConsensusPledge[:], ",")
 			}()),
+			fmt.Sprintf("--database.forceCacheTime=%d", config.ForceCacheTime),
```
georgysavva, line 148: Since now `database.forceCacheTime` expects a `time.Duration` type which looks like `10s` instead of just an int you should use "%s" instead of "%d"
```suggestion
			fmt.Sprintf("--database.forceCacheTime=%s", config.ForceCacheTime),
```

tools/integration-tests/tester/framework/parameters.go
```diff
@@ -124,6 +125,7 @@ type GoShimmerConfig struct {
 	WriteManaThreshold         float64
 	ReadManaThreshold          float64
 	SnapshotResetTime          bool
+	ForceCacheTime             int
```
georgysavva, line 128: replace with time.Duration:

```suggestion
	ForceCacheTime             time.Duration
```

georgysavva (commented):
A few small improvements needed, but looks great overall!

packages/tangle/tangle.go
GalRogozinski: ok,
I deleted the `/* forceCacheTime */` comment because people using IDEs can just read the go docs

tools/integration-tests/tester/framework/docker.go
```diff
@@ -65,6 +65,7 @@ func (d *DockerContainer) CreateGoShimmerEntryNode(name string, seed string) err
 			fmt.Sprintf("--node.disablePlugins=%s", disabledPluginsEntryNode),
 			"--autopeering.entryNodes=",
 			fmt.Sprintf("--autopeering.seed=base58:%s", seed),
+			fmt.Sprintf("--database.forceCacheTime=%d", noCache),
```
georgysavva, line 68: Why do you have a separate configuration for EntryNode? 
Why don't just do as for peer:
```go
			fmt.Sprintf("--database.forceCacheTime=%s", config.ForceCacheTime)
```
With `ForceCacheTime=0`?

GalRogozinski: I noticed that there is a current pattern in the code where entry node has different hard-coded configs

I didn't want to break this pattern, even though I don't really understand why it exists...
So before I change anyhting can you tell me why all the other peer configs are not passed to entry node?

packages/ledgerstate/objectstorage.go
```diff
@@ -38,66 +40,104 @@ const (
...
-	objectstorage.CacheTime(10 * time.Second),
-	objectstorage.LeakDetectionEnabled(false),
-}
+	// outputStorageOptions contains a list of default settings for the Output object storage.
+	outputStorageOptions []objectstorage.Option
+
+	// outputMetadataStorageOptions contains a list of default settings for the OutputMetadata object storage.
+	outputMetadataStorageOptions []objectstorage.Option
+
+	// consumerStorageOptions contains a list of default settings for the OutputMetadata object storage.
```
jonastheis, line 78:
```suggestion
	// consumerStorageOptions contains a list of default settings for the Consumer object storage.
```

jonastheis (commented):
Great work, looks awesome!

tools/integration-tests/tester/framework/docker.go
georgysavva: Ah okay, I the entry node doesn't work as a general Goshimmer node, most of the plugins are disabled. 
As you can see in the list `disabledPluginsEntryNode` the `messagelayer` plugin is disabled, so the database won't be used at all. You can just remove the cache configuration at all for an entry node.

packages/tangle/tangle.go
```diff
@@ -95,6 +97,7 @@ func (t *Tangle) Configure(options ...Option) {
 			Store:                        mapdb.NewMapDB(),
 			Identity:                     identity.GenerateLocalIdentity(),
 			IncreaseMarkersIndexCallback: increaseMarkersIndexCallbackStrategy,
+			CacheTimeProvider:            database.NewCacheTimeProvider(-1),
```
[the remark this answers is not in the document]
georgysavva: Yeah, I also thought about that, here is what stopped me:
1. I think that the default behavior should be the same as what we have now (cache is enabled)
2. It would be weird that the tangle has default cache time 0, but the GoShimmer node itself has cache time enabled (-1) in case no one passed a CLI arg or config file.
3. The fact that we initialize the tangle in a single place won't stay forever, I would like to refactor our plugin system and get rid of singletons in the future. So an inner-level code like the tangle package shouldn't make assumptions that It gets initialized from a single place. 
4. In case a developer forgot to pass the option it's better to run unit tests with cache than to run production code without cache.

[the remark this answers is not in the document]
GalRogozinski: However, it is also confusing for people who read only production code to see that the default is 0 and later gets overridden... people who see this won't know it is for tests..

I would even prefer to not have a default for this, and test writers who forget about the CacheTimeProvider will be reminded by an exception... But you can argue that this is also ugly...

I think the better solution would be to always use a testutil to initialize the tangle and the ideal solution is to mock the tangle in unit tests that just need it as a dependency (do we want to do this)?... But this is for another PR...

So as for right now, I would vote to just remove the default option because I think code is read more often than it is written.
I just think it is better to force test writers to think about using `newTestTangle` or mocks, rather to just have them incorrectly initialize the tangle in tests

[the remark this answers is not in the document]
georgysavva: I am okay with both options: remove the option completely or keep it with cache enabled (-1). @GalRogozinski up to you!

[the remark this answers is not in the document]
jonastheis: I'd favor @GalRogozinski's suggestion to leave it away. So it is a coding error when you forget to specify it and it can't be forgotten in unit tests :)

georgysavva (approved):
Great job!

GalRogozinski:
@georgysavva and @jonastheis 
all unit tests passed and no integration test is constatnly failing (only intermittently) 

you guys want to approve and merge?

jonastheis (approved):

PR #1444, opened 2021-06-14
Fix memory leak in pow plugin/pkg

lzpap:
# Description of change

`hash.New()` implementation  failed to release memory when computing pow for messages.


GalRogozinski (approved):
Looks Good!

PR #1435, opened 2021-06-11
refactor: Refactor value and faucet integration tests

jkrvivian:
# Description of change

Part of #1187 
* Remove TestFaucetPersistence test
* Rename `faucetmoveallfunds_test.go ` to `preparefaucet_test.go`
* Combine `TestTransactionsPersistence` and `TestValueColoredPersistence`
* Add `AwaitPeerGetFundsFromFaucet`
* Fixes on other `Await...` functions
* Replace most of the sleeps with `Await...` functions

## Type of change

- Bug fix 
- Enhancement 

## Change checklist

- [x] My code follows the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes

PR #1478, opened 2021-06-17
Refactor Test Framework

Wollac:
## Description of change

- Introduce a `health` endpoint that only returns true, if all plugins have been started.
- Creating a new node in the test frameworks waits for the node to report healthy. This eliminates the need for sleep after a start.
- Refactor `testutil` to not query nodes in parallel: As the CI usually only provides 1 or 2 cores this extra parallelism caused rare timeouts without any real benefits.
- Introduced a reflection-based docker configuration automatically translating config structs into the corresponding command line flags. 
- Autopeering waits for a connected network; only requiring minimum amount of neighbors could lead to a network split in rare cases
- Autopeering waits until each peer as discovered the entire network
- Refactor `common` test to make it more robust
- Refactor `mana` test to make it more robust
- Refactor `DRNG` test to make it more robust
- Integrate #1435 
- Disable the PoW check in the faucet API, as this depends on a configuration parameter of a disabled plugin.
- Upgrade [hive.go](https://github.com/iotaledger/hive.go)


tools/integration-tests/tester/framework/framework.go
```diff
@@ -40,298 +40,188 @@ func Instance() (f *Framework, err error) {
...
+			end = numPeers
 		}
-		_, err = network.createPartition(network.peers[i:end])
+		err = network.createPartition(ctx, network.peers[i:end])
 		if err != nil {
 			return nil, err
 		}
 	}
 	// wait until pumba containers are started and block traffic between partitions
 	time.Sleep(5 * time.Second)
```
jonastheis, line 145: Why not this?
```suggestion
	time.Sleep(graceTimePumba)
```

tools/integration-tests/tester/tests/common/common_synchronization_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// 1. issue data messages
+	log.Printf("Issuing %d messages to sync...", numMessages)
+	ids := tests.SendDataMessages(t, n.Peers(), numMessages)
+	log.Println("Issuing messages... done")
+
+	// 2. spawn peer without knowledge of previous messages
+	log.Println("Spawning new node to sync...")
+	newPeer, err := n.CreatePeer(ctx, createNewPeerConfig())
+	require.NoError(t, err)
+	err = n.DoManualPeering(context.Background())
```
jonastheis, line 43: Why does this not use `ctx` as well?

tools/integration-tests/tester/tests/common/common_synchronization_test.go
```diff
@@ -0,0 +1,99 @@
...
+	require.False(t, tests.Synced(t, newPeer))
+
+	// 7. issue some messages on old peers so that new peer can sync again
+	log.Printf("Issuing %d messages on the %d initial peers...", numSyncMessages, initialPeers)
+	ids = tests.SendDataMessages(t, n.Peers()[:initialPeers], numSyncMessages, ids)
+	log.Println("Issuing messages... done")
+
+	// 9. check whether all issued messages are available on all nodes
+	tests.RequireMessagesAvailable(t, []*framework.Node{newPeer}, ids, time.Minute, tests.Tick)
+	tests.RequireMessagesEqual(t, []*framework.Node{newPeer}, ids)
```
jonastheis, line 86:
```suggestion
	tests.RequireMessagesAvailable(t, n.Peers(), ids, time.Minute, tests.Tick)
	tests.RequireMessagesEqual(t, n.Peers(), ids)
```
The comment above says this.

jonastheis (changes requested):
Awesome stuff! I left a few minor comments. 

I think it would be great if we can come up with a dynamic configuration so that we don't have to redefine the config in the integration tests but simply reuse the configuration structs that are (not yet) existing for each plugin.

tools/integration-tests/tester/framework/docker.go
```diff
@@ -203,68 +136,86 @@ func (d *DockerContainer) CreatePumba(name string, containerName string, targetI
...
 	if err != nil {
 		return -1, err
 	}
 
 	return resp.State.ExitCode, nil
 }
 
+// Shutdown stops the container and stores its log. It returns the exit status of the process.
+func (d *DockerContainer) Shutdown(ctx context.Context, optionalTimeout ...time.Duration) (exitStatus int, err error) {
+	err = d.Stop(ctx, optionalTimeout...)
```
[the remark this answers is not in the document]
Wollac: If the container does not stop gracefully within the timeout it gets killed. Thus, the `Stop` should always return until something is really amiss with the docker framework. In which case getting a log would most probably also fail.
This for example happens for the `drand` instances, they need to get killed but we still get the logs.

tools/integration-tests/tester/framework/partition.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, peer := range p.peers {
+		ids = append(ids, peer.ID().String())
+	}
+	return ids
+}
+
+// deletePartition deletes a partition, all its Pumba containers and creates logs for them.
+func (p *Partition) deletePartition(ctx context.Context) error {
+	// stop containers
+	for _, pumba := range p.pumbas {
```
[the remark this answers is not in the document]
Wollac: In contrast to GoShimmer shutting down pumba takes neglectable time, but sure, it should not hurt to shut them down in parallel.

tools/integration-tests/tester/tests/common/common_synchronization_test.go
Wollac: The comment above says `check whether all issued messages are available on to the new peer` which is exactly what the existing code is checking.
This is a sync test and for this the status on the syncing is the most relevant, so I think it makes sense to keep the test as narrow as possible and only check for `newPeer`

jonastheis: `// 9. check whether all issued messages are available on all nodes` ?

tools/integration-tests/tester/framework/network.go
```diff
@@ -28,558 +28,491 @@ type Network struct {
...
 
-// deletePartition deletes a partition, all its Pumba containers and creates logs for them.
-func (p *Partition) deletePartition() error {
-	// stop containers
-	for _, pumba := range p.pumbas {
-		err := pumba.Stop()
-		if err != nil {
-			return err
+	// simple DFS to determine reachable nodes
+	var u string
```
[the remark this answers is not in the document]
Wollac: hmm, I mean you could do
```go
missing := make(map[string]struct{}, len(nodes))
for _, node := range nodes {
	missing[node.ID().String()] = struct{}{}
}

// simple DFS to determine unreachable nodes
dfs := walker.New(false)
dfs.Push(nodes[0].ID().String())

for dfs.HasNext() {
	u := dfs.Next().(string)
	delete(missing, u)

	for v := range connections[u] {
		dfs.Push(v)
	}
}
return len(missing) == 0
```
would you consider this simplified? I am not sure tbh, maybe the previous code was even easier to understand

tools/integration-tests/tester/framework/partition.go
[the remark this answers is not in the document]
Wollac: :+1:

tools/integration-tests/tester/tests/common/common_synchronization_test.go
Wollac: Then I'll change that comment

jonastheis (approved):

PR #1515, opened 2021-06-25
Exit gossip worker functions on error

jonastheis:
# Description of change
The gossip workers continue in their flow even when there are errors creating unnecessary overhead. This PR exits out of these functions if an error is detected. 

## Type of change
- Enhancement

Executed unit and integration tests.


Wollac (approved):
Lgtm

PR #1522, opened 2021-06-29
Debugging/several fixes

hmoog:


jkrvivian (approved):

packages/gossip/manager.go
```diff
@@ -220,7 +224,10 @@ func (m *Manager) getNeighborsByID(ids []identity.ID) []*Neighbor {
 }
 
 func (m *Manager) send(b []byte, to ...identity.ID) {
-	neighbors := m.getNeighborsByIDOrRandom(to...)
+	neighbors := m.getNeighborsByID(to)
```
apenzk, line 227: Doing this leaves the function getNeighborsByIDOrRandom and other functions related to this as a waste product. 

We should either change the const randomNeighborsNum to the neighbor number len(allNeighbors) 
or 
delete getNeighborsByIDOrRandom and getRandomNeighbors

apenzk (changes requested):

PR #1524, opened 2021-06-30
Check synced state at least every sync time window interval

### LINKED ISSUE #1477: Make sync status consistent with Tangle Time
Make the `TimeManager`'s `lastSynced` field consistent over calls of `Synced()` as otherwise the node might think it is in sync just because it never got new messages confirmed and thus `updateTime()` never called.

luca-moser:
# Description of change

The `TimeManager` keeps a flag `lastSynced` representing the synced state of the node. This flag only ever gets updated when a corresponding `MarkerConfirmation` event is triggered leading to the flag never being updated in relation to the sync time window if no marker gets confirmed; this means a node might think that it is synced, even if it did not confirm any new marker within the sync time window (which is our definition of being "synced").

This PR adds a ticker which schedules the "synced check" at least every sync time window interval to keep the flag consistent even if no markers got confirmed.

Closes #1477

## Type of change

Integration tests.

## Change checklist

### DIFF
--- a/packages/tangle/timemanager.go
+++ b/packages/tangle/timemanager.go
@@ -4,15 +4,15 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cockroachdb/errors"
 	"github.com/iotaledger/goshimmer/packages/clock"
 	"github.com/iotaledger/goshimmer/packages/markers"
-
-	"github.com/cockroachdb/errors"
 	"github.com/iotaledger/hive.go/cerrors"
 	"github.com/iotaledger/hive.go/events"
 	"github.com/iotaledger/hive.go/kvstore"
 	"github.com/iotaledger/hive.go/marshalutil"
 	"github.com/iotaledger/hive.go/stringify"
+	"github.com/iotaledger/hive.go/timeutil"
 )
 
 const (
@@ -32,7 +32,11 @@ type TimeManager struct {
 
 	lastConfirmedMutex   sync.RWMutex
 	lastConfirmedMessage LastConfirmedMessage
+	lastSyncedMutex      sync.RWMutex
 	lastSynced           bool
+
+	shutdownSignal chan struct{}
+	shutdownOnce   sync.Once
 }
 
 // NewTimeManager is the constructor for TimeManager.
@@ -41,8 +45,9 @@ func NewTimeManager(tangle *Tangle) *TimeManager {
 		Events: &TimeManagerEvents{
 			SyncChanged: events.NewEvent(SyncChangedCaller),
 		},
-		tangle:      tangle,
-		startSynced: tangle.Options.StartSynced,
+		tangle:         tangle,
+		startSynced:    tangle.Options.StartSynced,
+		shutdownSignal: make(chan struct{}),
 	}
 
 	// initialize with Genesis
@@ -68,6 +73,11 @@ func NewTimeManager(tangle *Tangle) *TimeManager {
 	return t
 }
 
+// Start starts the TimeManager.
+func (t *TimeManager) Start() {
+	go t.mainLoop()
+}
+
 // Setup sets up the behavior of the component by making it attach to the relevant events of other components.
 func (t *TimeManager) Setup() {
 	t.tangle.ApprovalWeightManager.Events.MarkerConfirmation.Attach(events.NewClosure(t.updateTime))
@@ -82,6 +92,10 @@ func (t *TimeManager) Shutdown() {
 		t.tangle.Events.Error.Trigger(errors.Errorf("failed to persists LastConfirmedMessage (%v): %w", err, cerrors.ErrFatal))
 		return
 	}
+
+	t.shutdownOnce.Do(func() {
+		close(t.shutdownSignal)
+	})
 }
 
 // LastConfirmedMessage returns the last confirmed message.
@@ -103,9 +117,8 @@ func (t *TimeManager) Time() time.Time {
 // Synced returns whether the node is in sync based on the difference between TangleTime and current wall time which can
 // be configured via SyncTimeWindow.
 func (t *TimeManager) Synced() bool {
-	t.lastConfirmedMutex.RLock()
-	defer t.lastConfirmedMutex.RUnlock()
-
+	t.lastSyncedMutex.RLock()
+	defer t.lastSyncedMutex.RUnlock()
 	return t.lastSynced
 }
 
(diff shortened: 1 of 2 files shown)


packages/tangle/timemanager.go
```diff
@@ -32,7 +33,15 @@ type TimeManager struct {
 
 	lastConfirmedMutex   sync.RWMutex
 	lastConfirmedMessage LastConfirmedMessage
+	lastSyncedMutex      sync.RWMutex
 	lastSynced           bool
+
+	mu             sync.Mutex
+	ticker         *time.Ticker
+	started        typeutils.AtomicBool
+	stopped        typeutils.AtomicBool
```
Wollac, line 42: Are `started` and `stopped` actually used anywhere?

packages/tangle/timemanager.go
```diff
@@ -103,9 +127,8 @@ func (t *TimeManager) Time() time.Time {
 // Synced returns whether the node is in sync based on the difference between TangleTime and current wall time which can
 // be configured via SyncTimeWindow.
 func (t *TimeManager) Synced() bool {
-	t.lastConfirmedMutex.RLock()
-	defer t.lastConfirmedMutex.RUnlock()
-
+	t.lastSyncedMutex.RLock()
+	defer t.lastSyncedMutex.RUnlock()
 	return t.lastSynced
```
Wollac, line 132: Alternatively, we could make `lastSynced` an atomic. Or could this cause any issues that when `Synced` is called at the same time of `updateSyncedState`, it wouldn't wait for the update?

packages/tangle/timemanager.go
```diff
@@ -117,33 +140,59 @@ func (t *TimeManager) synced() bool {
...
+
+loop:
+	for {
+		select {
+		case <-t.ticker.C:
+			t.updateSyncedState()
+		case <-t.shutdownSignal:
+			break loop
+		}
+	}
```
Wollac, line 193: There is `timeutil.Ticker` which is supposed to do exactly that. Maybe you could have a look at that, but tbh I think the current code is more explicit and easier to understand than a util function although it would maybe safe a few lines of code.

packages/tangle/timemanager.go
```diff
@@ -117,33 +140,59 @@ func (t *TimeManager) synced() bool {
...
+	defer t.lastSyncedMutex.Unlock()
+	if newSynced := t.synced(); t.lastSynced != newSynced {
+		t.lastSynced = newSynced
+		// trigger the event inside the lock to assure that the status is still correct
+		t.Events.SyncChanged.Trigger(&SyncChangedEvent{Synced: newSynced})
+	}
+}
+
 // updateTime updates the last confirmed message.
 func (t *TimeManager) updateTime(marker markers.Marker, newLevel int, transition events.ThresholdEventTransition) {
```
Wollac, line 156: Is it an issue, if `updateTime` gets called after `Shutdown` do we need to check for that or can it be ignored?

luca-moser: Since the `TimeManager` gets shutdown after the components triggering the `MarkerConfirmation` event of the Tangle, `updateTime` should not be called. But even if it would, it would only cause the persisted "last seen confirmed msg" to be inconsistent with what was actually lastly confirmed.

packages/tangle/timemanager.go
luca-moser: You are right that an atomic would work too but it seems that handlers of the `SyncChanged` event are supposed to be executed with the assurance that the given updated sync flag is valid during their execution. Of course, examining the event handlers for `SyncChanged` reveals that for them it wouldn't really matter since this flag can change shortly after anyway. But lets keep the lock for now to keep that property.

packages/tangle/timemanager.go
luca-moser: Changed it to use that ticker; note that the `Scheduler` exhibits a similar loop which can be changed to it.

packages/tangle/timemanager.go
luca-moser: I've removed them.

packages/tangle/timemanager.go
```diff
@@ -82,6 +93,12 @@ func (t *TimeManager) Shutdown() {
 		t.tangle.Events.Error.Trigger(errors.Errorf("failed to persists LastConfirmedMessage (%v): %w", err, cerrors.ErrFatal))
 		return
 	}
+
+	t.shutdownOnce.Do(func() {
+		t.mu.Lock()
```
Wollac, line 98: Is there any reason for this lock? Shouldn't the `Once` assure that the function is not called twice at the same time?

packages/tangle/timemanager.go
```diff
@@ -4,15 +4,15 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cockroachdb/errors"
```
Wollac, line 7: I don't think this is formatted according to `goimports -local github.com/iotaledger/goshimmer`, so I am wondering why the Dog is not complaining :thinking:

packages/tangle/timemanager.go
```diff
@@ -117,33 +133,54 @@ func (t *TimeManager) synced() bool {
...
 
+// the main loop runs the updateSyncedState at least every synced time window interval to keep the synced state updated
+// even if no updateTime is ever called.
+func (t *TimeManager) mainLoop() {
+	timeutil.NewTicker(t.updateSyncedState, func() time.Duration {
+		if t.tangle.Options.SyncTimeWindow == 0 {
+			return DefaultSyncTimeWindow
+		}
+		return t.tangle.Options.SyncTimeWindow
+	}()).WaitForShutdown()
```
Wollac, line 181: `t.shutdownSignal` parameter is missing

packages/tangle/timemanager.go
```diff
@@ -117,33 +133,54 @@ func (t *TimeManager) synced() bool {
...
 		}
+
+		t.updateSyncedState()
 	})
 }
 
+// the main loop runs the updateSyncedState at least every synced time window interval to keep the synced state updated
+// even if no updateTime is ever called.
+func (t *TimeManager) mainLoop() {
+	timeutil.NewTicker(t.updateSyncedState, func() time.Duration {
```
Wollac, line 176: I personally think:
```go
interval := DefaultSyncTimeWindow
if t.tangle.Options.SyncTimeWindow != 0 {
	interval = t.tangle.Options.SyncTimeWindow
}
```
is a bit nicer than the `func` construct, but I am happy with both

packages/tangle/timemanager.go
luca-moser: Executing `goimports` against the file doesn't yield any changes. 🤷🏻

Wollac: Even when doing `goimports -local github.com/iotaledger/goshimmer`? it should group all the goshimmer imports...
But who am I to complain, if the Dog isn't :shrug:

Wollac (approved):

Issue #1507, opened 2021-06-24
Unify parameter names


Wollac:
Currently, the plugin and parameter names are inconsistent:
 - Parameter names, including the namespaces, should all follow `lowerCamelCase`, e.g. `messageLayer`, `webAPI`, `instanceID`,`basicAuth`,...
 - Use the `time.Duration` type instead of specifying seconds in integers.
 - Consistently use bind address instead of port.
- Plugin names should match the corresponding parameter namespace, if they have parameters.
- We sometimes have sub plugins (e.g. `Gossip[TipsBroadcaster]`, `WebAPI tools Endpoint`). These names should be use a consistent scheme, that is compatible with Go type names, e.g. `CamelCase`
- Assure that parameters are only used in that plugin: e.g. `autopeering.Seed` is used at various places, (also when autopeering is disabled) so it should be renamed to `identity.Seed` or similar.

All these changes impact configuration for users, devnet, tests so they should all be deployed at the same time.

PR #1525, opened 2021-06-30
Use ParametersDefinition structs in integration test framework

jonastheis:
# Description of change
This PR introduces the new way of defining config parameters for each plugin. We define a `ParametersDefinition` type to describe the flags and their default values and a `Parameters` variable to bind and use the flags. 

In the integration testing framework the manually defined configs for plugins are modified to embed the `ParametersDefinition` to avoid recreating the config from the integration tests. The `default` values of the parameters are read via reflection and written to the config struct within the integration testing framework. 

## Type of change
- Enhancement 

Ran the tests.


tools/integration-tests/tester/framework/config/config.go
```diff
@@ -31,76 +49,181 @@ type GoShimmer struct {
...
+			}
+			valueField.Set(reflect.ValueOf(defaultValue))
+		case []string:
+			if defaultValue, err = csv.NewReader(strings.NewReader(tagDefaultValue)).Read(); err != nil {
+				panic(err)
+			}
+			valueField.Set(reflect.ValueOf(defaultValue))
+		default:
+			panic(fmt.Sprintf("type `%s` not implemented", reflect.TypeOf(defaultValue)))
+		}
```
Wollac, line 161:
```suggestion
		switch valueField.Interface().(type) {
		case string: // no conversion needed
			valueField.SetString(tagDefaultValue)
		case []string: // parse comma separated strings instead of JSON-style lists
			defaultValue, err := csv.NewReader(strings.NewReader(tagDefaultValue)).Read()
			if err != nil {
				panic(err)
			}
			valueField.Set(reflect.ValueOf(defaultValue))
		case time.Duration: // Duration does not implement encoding.TextUnmarshaler, so we have to do it manually
			defaultValue, err := time.ParseDuration(tagDefaultValue)
			if err != nil {
				panic(err)
			}
			valueField.Set(reflect.ValueOf(defaultValue))
		default: // use the JSON unmarshaler for everything else
			if err := json.Unmarshal([]byte(tagDefaultValue), valueField.Addr().Interface()); err != nil {
				panic(err)
			}
		}
```

This uses json instead of adding the switch complexity here.

tools/integration-tests/tester/framework/config/config.go
```diff
@@ -31,76 +49,181 @@ type GoShimmer struct {
...
+// NewGoShimmer creates a GoShimmer config initialized with default values.
+func NewGoShimmer() (config GoShimmer) {
+	config = GoShimmer{}
+	fillStructFromDefaultTag(&config)
+	return
+}
+
+// fillStructFromDefaultTag recursively explores the given struct pointer and sets values of fields to the `default` as
+// specified in the struct's tags.
+func fillStructFromDefaultTag(s interface{}) {
```
Wollac, line 61: I would make this `fillStructFromDefaultTag(val reflect.Value)` and move the single `reflect` conversion inside the `NewGoShimmer`

tools/integration-tests/tester/framework/config/flags.go
```diff
@@ -47,11 +49,22 @@ func (s GoShimmer) CreateFlags() []string {
 	flags = append(
 		[]string{
 			"--node.enablePlugins=Webapi tools Endpoint",
-			fmt.Sprintf("--autopeering.seed=base58:%s", base58.Encode(s.Seed)),
 			fmt.Sprintf("--node.enablePlugins=%s", setToString(enabledPlugins)),
 			fmt.Sprintf("--node.disablePlugins=%s", setToString(disabledPlugins)),
 		},
 		flags...)
+
+	// manually add seed to flags if autopeering is disabled
```
Wollac, line 57: This is super ugly, the seed has nothing to do with the autopeering anymore! However unless we rename this as part of #1507 I guess there is no way around it.

plugins/autopeering/parameters.go
```diff
@@ -14,7 +14,10 @@ var Parameters = struct {
 	Ro float64 `default:"2." usage:"Ro parameter"`
```
Wollac, line 14:
```suggestion
	Ro float64 `default:"2" usage:"Ro parameter"`
```
`2.` is strange let's either go for `2` or `2.0`

tools/integration-tests/tester/framework/config/flags.go
Wollac: Can we maybe add a comment explaining that this is only because the seed is in the wrong struct?

tools/integration-tests/tester/framework/config/config.go
Wollac: Should we maybe move that function into a different file to keep config.go concise as this probably requires updates more frequently.

plugins/autopeering/parameters.go
apenzk: interestingly in older versions of python   `2` would define an integer and `2.` would define a float. Thus in Python i would define float variables as `someVariable = 2.`

plugins/autopeering/local/parameters.go
```diff
@@ -1,23 +1,34 @@
...
+
+// ParametersNetwork contains the configuration parameters of the local peer's network.
+var ParametersNetwork = ParametersDefinitionNetwork{}
 
 func init() {
-	flag.String(CfgBind, "0.0.0.0", "bind address for global services such as autopeering and gossip")
-	flag.String(CfgExternal, "auto", "external IP address under which the node is reachable; or 'auto' to determine it automatically")
-	flag.Int(CfgPort, 14626, "UDP port for incoming peering requests")
-	flag.String(CfgSeed, "", "private key seed used to derive the node identity; optional base58 or base64 encoded 256-bit string. Prefix with 'base58:' or 'base64', respectively")
+	configuration.BindParameters(&ParametersLocal, "autopeering")
```
apenzk, line 32: I noticed several parameter instances are mapped to "autopeering" (e.g. see also discovery/parameters.go). What happens if we incidently create two parameters with the same name. Do they write to the same parameter in "autopeering" ?

plugins/manaeventlogger/parameters.go
```diff
@@ -1,18 +1,22 @@
...
 )
 
+// ParametersDefinition contains the definition of the parameters used by the manaeventlogger plugin.
+type ParametersDefinition struct {
+	// CSV defines the file path to store mana events.
+	CSV string `name:"csv" default:"/tmp/consensusManaEvents.csv" usage:"file to store mana events"`
+	// Buffersize defines the events' buffer size.
+	BufferSize int `default:"100" usage:"event logs buffer size"`
+	// CheckBufferIntervalSec defines interval between buffer checks.
+	CheckBufferIntervalSec int `default:"5" usage:"check buffer interval secs"`
```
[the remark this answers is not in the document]
apenzk: i absolutely agree we should move away from defining second intervals as integers. This should be time.Duration but how is the default then given ? `default:"5*time.Second"` ?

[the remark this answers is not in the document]
apenzk: btw already sometimes durations are defined as time.Duration and sometimes as int. So we should clean this inconsistency up. please add me to the review or task when we do this.

plugins/networkdelay/parameters.go
```diff
@@ -0,0 +1,18 @@
...
+type ParametersDefinition struct {
+	// OriginPublicKey defines the default issuer node public key in base58 encoding.
+	OriginPublicKey string `default:"9DB3j9cWYSuEEtkvanrzqkzCQMdH1FGv3TawJdVbDxkd" usage:"default issuer node public key"`
+}
+
+// Parameters contains the configuration used by the networkdelay plugin.
+var Parameters = ParametersDefinition{}
+
+func init() {
+	configuration.BindParameters(&Parameters, "networkdelay")
```
apenzk, line 17: In https://goshimmer.docs.iota.org/implementation_design/configuration_parameters.html we state that the optionalNamespace is lowerCamelCase but in the code we use lowereverything. Which one is the correct one?

func init() {
    configuration.BindParameters(&Parameters, "customPlugin")
}

tools/integration-tests/tester/framework/config/config.go
```diff
@@ -31,76 +49,181 @@ type GoShimmer struct {
...
+	val := reflect.ValueOf(s).Elem()
```
apenzk, line 62: https://goshimmer.docs.iota.org/implementation_design/configuration_parameters.html should include a discussion about `reflect`

apenzk (changes requested):

plugins/autopeering/local/parameters.go
jonastheis: Yeah this is not optimal. However, we didn't want to break the interface with this PR. I think cleaning up the configuration parameters should be done with https://github.com/iotaledger/goshimmer/issues/1507

plugins/drng/parameters.go
```diff
@@ -1,60 +1,60 @@
...
-	CfgDRNGCustomDistributedPubKey = "drng.custom.distributedPubKey"
-	// CfgDRNGCustomCommitteeMembers defines the config flag of the DRNG committee members identities.
-	CfgDRNGCustomCommitteeMembers = "drng.custom.committeeMembers"
-)
+// ParametersDefinition contains the definition of configuration parameters used by the drng plugin.
+type ParametersDefinition struct {
+	// Pollen contains the configuration parameters of GoShimmer dRNG committee.
+	Pollen struct {
+		// InstanceID defines the config flag of the DRNG instanceId.
+		InstanceID int `name:"instanceId" default:"1" usage:"instance ID of the GoShimmer drng instance"`
```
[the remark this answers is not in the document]
jonastheis: I'd just leave it for now and clean it up with https://github.com/iotaledger/goshimmer/issues/1507

plugins/manaeventlogger/parameters.go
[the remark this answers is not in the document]
Wollac: The `default` can only be a string. As such it must be the string representation of `time.Duration`, i.e. `30s` etc

plugins/gossip/plugin.go
```diff
@@ -46,7 +42,7 @@ func configure(*node.Plugin) {
 
 func run(*node.Plugin) {
 	if err := daemon.BackgroundWorker(PluginName, start, shutdown.PriorityGossip); err != nil {
-		log.Panicf("Failed to start as daemon: %s", err)
+		plugin.Logger().Panicf("Failed to start as daemon: %s", err)
```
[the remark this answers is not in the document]
jonastheis: Since it always should be initialized I think this is the easier option

plugins/networkdelay/parameters.go
jonastheis: This should also be part of https://github.com/iotaledger/goshimmer/issues/1507.

tools/integration-tests/tester/framework/config/flags.go
jonastheis: Done. Yeah it's ugly and should be removed once we do #1507.

tools/integration-tests/tester/framework/config/config.go
jonastheis: What exactly do you mean?

plugins/drng/parameters.go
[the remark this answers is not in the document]
Wollac: :+1:

plugins/gossip/plugin.go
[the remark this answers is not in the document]
Wollac: Still, that seems very strange to me as we even have this (otherwise unused) `plugin *node.Plugin` parameter two lines above.
However, if you don't like using that how about using `Plugin()` to indicate that this will always be initialized?

tools/integration-tests/tester/framework/config/config.go
```diff
@@ -31,76 +49,116 @@ type GoShimmer struct {
...
+// NewGoShimmer creates a GoShimmer config initialized with default values.
+func NewGoShimmer() (config GoShimmer) {
+	config = GoShimmer{}
+	fillStructFromDefaultTag(&config)
+	return
+}
+
+// fillStructFromDefaultTag recursively explores the given struct pointer and sets values of fields to the `default` as
+// specified in the struct's tags.
+func fillStructFromDefaultTag(s interface{}) {
```
GalRogozinski, line 61: doesn't similar logic exist in hive.go?
don't we have sime duplication?

plugins/faucet/plugin.go
```diff
@@ -93,90 +62,85 @@ func Plugin() *node.Plugin {
 // Faucet gets the faucet component instance the faucet plugin has initialized.
 func Faucet() *StateManager {
 	faucetOnce.Do(func() {
-		base58Seed := config.Node().String(CfgFaucetSeed)
-		if base58Seed == "" {
-			log.Fatal("a seed must be defined when enabling the faucet plugin")
+		if Parameters.Seed == "" {
+			plugin.LogFatal("a seed must be defined when enabling the faucet plugin")
```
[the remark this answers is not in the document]
jonastheis: The thing is that every plugin has already in integrated logger, created with the plugin's name. This avoids creating the a logger within every plugin over and over again. So we should also use that.

tools/integration-tests/tester/framework/config/config.go
jonastheis: It is indeed similar but not exactly the same as this function is actually filling in the struct with the default values. In hive.go it uses them to create flags.

GalRogozinski: still
shouldn't it be in hive?

GalRogozinski:
Something that bothers me with golang is that due to the package scope files can access the config and overwrite their values.
In Java I would've made them private and only expose their values via getters

I guess such a thing will be hard to achieve in golang?

plugins/clock/parameters.go
```diff
@@ -0,0 +1,18 @@
+package clock
+
+import (
+	"github.com/iotaledger/hive.go/configuration"
+)
+
+// ParametersDefinition contains the definition of configuration parameters used by the clock plugin.
+type ParametersDefinition struct {
+	// NtpPools defines the config flag of the NTP pools.
+	NtpPools []string `default:"0.pool.ntp.org,1.pool.ntp.org,2.pool.ntp.org" usage:"list of NTP pools to synchronize time from"`
```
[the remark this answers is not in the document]
karimodm: True: looking at the `lowerCamelCase` better, `NTPPools` will be converted to `ntpPools`! :cool:

plugins/clock/plugin.go
```diff
@@ -37,14 +30,9 @@ func Plugin() *node.Plugin {
...
 
-func init() {
-	flag.StringSlice(CfgNTPPools, []string{"0.pool.ntp.org", "1.pool.ntp.org", "2.pool.ntp.org"}, "list of NTP pools to synchronize time from")
-}
-
 func configure(plugin *node.Plugin) {
-	ntpPools = config.Node().Strings(CfgNTPPools)
-	if len(ntpPools) == 0 {
-		plugin.LogFatalf("%s needs to provide at least 1 NTP pool to synchronize the local clock.", CfgNTPPools)
+	if len(Parameters.NtpPools) == 0 {
```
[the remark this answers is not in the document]
karimodm: If a `--clock.ntpPools=""` would be passed to goshimmer cli the default value would be overwritten with an empty array. So this check is actually still necessary,

plugins/activity/parameters.go
```diff
@@ -2,15 +2,18 @@ package activity
...
 	BroadcastIntervalSec int `default:"3" usage:"the interval at which the node will broadcast its activity message"`
 	// ParentsCount is the number of parents that node will choose for its activity messages.
 	ParentsCount int `default:"8" usage:"the number of parents that node will choose for its activity messages"`
 	// DelayOffset is the maximum for random initial time delay before sending the activity message.
 	DelayOffset int `default:"10" usage:"the maximum for random initial time delay before sending the activity message"`
-}{}
+}
+
+// Parameters contains the configuration parameters of the activity plugin.
+var Parameters = ParametersDefinition{}
```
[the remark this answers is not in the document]
karimodm: Yep, makes sense to change this.

plugins/config/parameters.go
```diff
@@ -1,18 +1,26 @@
...
+	"github.com/iotaledger/hive.go/configuration"
 )
 
 const (
-	// CfgDisablePlugins contains the name of the parameter that allows to manually disable node plugins.
 	CfgDisablePlugins = "node.disablePlugins"
-
-	// CfgEnablePlugins contains the name of the parameter that allows to manually enable node plugins.
-	CfgEnablePlugins = "node.enablePlugins"
+	CfgEnablePlugins  = "node.enablePlugins"
```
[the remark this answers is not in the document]
karimodm: Yeah, this is actually very ugly and even ambiguous: now you could set a Plugin config struct as `Enabled=true` but still have the plugin in the `node.disablePlugins` list; what would take precedence?
Despite being necessary, this change probably does not belong in this PR as it involves changing docker compose files and quite a few other things.

tools/integration-tests/tester/framework/config/config.go
karimodm: The documentation should be changed to reflect that parameters should be defined in a type `ParametersDefinition` and `Parameters` now becomes a var of that type. But documenting on how the internals work with `reflect` I think will go in too much detail imho.

tools/integration-tests/tester/framework/config/config.go
karimodm: This struct filling code is very specific to the integration framework itself; and probably there is no general use-case for this that would justify placing this functionality in hive.go.

karimodm:
@GalRogozinski it is true that we could avoid exporting these configuration parameters and provide getters instead.
This main issue there is that the parameters are sometimes accessed from outside the package itself: if we expose the `Parameters` variable itself, which is a pointer, that would still allow other packages to not only access but also change the parameters' values.
So you have a point, but right now I don't think we should do it.

plugins/config/parameters.go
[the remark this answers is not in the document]
Wollac: I agree, let's keep them as a special case and clean them up later

Wollac (approved):

plugins/autopeering/local/parameters.go
apenzk: Ok

plugins/gossip/plugin.go
```diff
@@ -44,7 +44,7 @@ func configure(*node.Plugin) {
 
 func run(*node.Plugin) {
 	if err := daemon.BackgroundWorker(PluginName, start, shutdown.PriorityGossip); err != nil {
-		log.Panicf("Failed to start as daemon: %s", err)
+		Plugin().Logger().Panicf("Failed to start as daemon: %s", err)
```
jkrvivian, line 47: Let's remove var `log` above.

jkrvivian (changes requested):

jkrvivian (approved):

PR #1550, opened 2021-07-08
Do not force a static goshimmer compile for Docker

karimodm:
# Description of change

Do not force a static goshimmer compile for Docker. Go knows better what to do.

Nodes still starts and syncs.


Dockerfile
```diff
@@ -4,7 +4,7 @@
 # golang 1.16.3-buster amd64
 FROM golang@sha256:dfa3cef088454200d6b48e2a911138f7d5d9afff77f89243eea6342f16ddcfb0 AS build
 
-ARG BUILD_TAGS=rocksdb,builtin_static,netgo
+ARG BUILD_TAGS=rocksdb,builtin_static
```
capossele, line 7: Why do you remove `netgo` here? We had issues with some architectures in the past when this was not present

karimodm:
Removed `netgo` once again ofter some testing.

capossele (approved):

Issue #1493, opened 2021-06-22
Re-design and apply data flow changes to reflect solidification of transactions


luca-moser:
- Add multithreading to Tangle (Booker)
- Break Tangle into components until Booker
- Add Transaction solidification flow

PR #1556, opened 2021-07-12
Eligibility Manager 

### LINKED ISSUE #1494: Add eligibility flag to transactions
Which do not reference via weak parents their dependencies until those get confirmed

daria305:
This PR is adding an `EligibilityManager` into the tangle package that is responsible for changing eligible flag of the meassage. Fixes #1494 

It is attached to two events:
 - MessageSolid 
 - TransactionConfirmed

It emits `MessageEligible` event

Dependencies are stored in `unconfirmedTransactionDependencies` objectStorage that maps dependency txID to the map of all dependent on it transactions IDs

After message solidification event is triggered the message is eligible:
 - if it is not a transaction
 - if all of its dependencies are confirmed
 - if all dependencies are directly approved by the message

If none of these conditions is met, dependencies are saved in object storage

Whenever transaction is confirmed Eligibility manager checks if there are any transactions dependent on the message that triggered the event and updates their eligibility status. After this it removes confirmed transactionID from the objectStorage

## Type of change

We added unit tests fro the manager and object storage, (listed in the issue comment)

## Change checklist

### DIFF
--- a/packages/consensus/fcob/consensusmechanism_test.go
+++ b/packages/consensus/fcob/consensusmechanism_test.go
@@ -219,7 +219,6 @@ func TestOpinionFormer_Scenario2(t *testing.T) {
 	testTangle.ConsensusManager.Events.MessageOpinionFormed.Attach(events.NewClosure(func(messageID tangle.MessageID) {
 		t.Logf("MessageOpinionFormed for %s", messageID)
 
-		assert.True(t, testTangle.ConsensusManager.MessageEligible(messageID))
 		assert.Equal(t, payloadLiked[messageID], testTangle.ConsensusManager.PayloadLiked(messageID))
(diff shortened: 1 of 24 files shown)


packages/consensus/fcob/consensusmechanism_test.go
```diff
@@ -219,7 +219,6 @@ func TestOpinionFormer_Scenario2(t *testing.T) {
 	testTangle.ConsensusManager.Events.MessageOpinionFormed.Attach(events.NewClosure(func(messageID tangle.MessageID) {
 		t.Logf("MessageOpinionFormed for %s", messageID)
 
-		assert.True(t, testTangle.ConsensusManager.MessageEligible(messageID))
```
GalRogozinski, line 222: This has been deleted since consensus mechanism is not in charge of eligibility and it has no place in the unit test

jonastheis:
#1493 might reuse some of this code.

PR #1577, opened 2021-07-27
Message syntactical validation

### LINKED ISSUE #1570: Syntactical Message validation with dislike switch
## Description

Upon receiving a Message, it must be syntactically validated following the changes in the message layout introduced with the dislike switch, which has been now made implicit. Refer to the [Message Structure Layout](https://github.com/iotaledger/goshimmer/blob/feat/dislike-parent-types/docOps/docs/protocol_specification/tangle.md#message-structure).
Rules:

1. A Strong Block parent must exist.
2. ParentBlock types cannot repeat.
3. Parent count per block 0 <= x <= 8.
4. Parents unique within block.
5. Parents lexicographically sorted within block.
6. ParentsBlockCount must match blocks actually parsed.
7. ParentsCount within each block must match the parents actually parsed within the block.
8. References need to be valid MessageIDs.
9. A Parent(s) repetition is only allowed when it occurs across Strong and Like parents. To support cases like this:
   
OTV-testcases-Page-1 (2)

GalRogozinski:
# Description of change

Add syntactic validation to messages parsed via gossip 
fixes #1570 

## Type of change

unit test added for the new filter

## Change checklist


packages/tangle/message.go
```diff
@@ -40,11 +40,11 @@ const (
...
 	MaxParentsCount = 8
 
-	// defines the minimum number of parents each parents block must have.
+	// MinParentsBlocksCount defines the minimum number of parents each parents block must have.
 	MinParentsBlocksCount = 1
 
-	// defines the maximum number of parents each parents block must have.
-	MaxParentsBlocksCount = 4
+	// MaxParentsBlocksCount defines the maximum number of parents each parents block must have.
+	MaxParentsBlocksCount = 8
```
karimodm, line 47: I think this variable should be renamed. Isn't it duplicated with `MaxParentsCount`?

packages/tangle/message.go
```diff
@@ -40,11 +40,11 @@ const (
 	// MaxParentsCount defines the maximum number of parents each parents block must have.
 	MaxParentsCount = 8
 
-	// defines the minimum number of parents each parents block must have.
+	// MinParentsBlocksCount defines the minimum number of parents each parents block must have.
```
karimodm, line 43: Rename the variable. Isn't it duplicated with `MinParentsCount`?

packages/tangle/parser.go
```diff
@@ -49,6 +51,7 @@ func NewParser() (result *Parser) {
 
 	// add builtin filters
 	result.AddBytesFilter(NewRecentlySeenBytesFilter())
+	result.AddMessageFilter(NewMessageSyntacticalValidator())
```
karimodm, line 54: Why introducing a filter to do this kind of job? Syntactical checks on the previous message layout were performed without.

karimodm (changes requested):
I am not sure if creating a filter is the right way to implement this: how were syntactical checks performed before? And why that code has not been replaced by this new one?

packages/tangle/parser.go
```diff
@@ -248,6 +251,157 @@ type MessageFilter interface {
...
+/**
+Validates:
+1. Block Types don't repeat themselves
+2. MessageIDs don't repeat themselves in a block
+3. MessageIDS don't repeat across blocks, except for Strong and Liked parents
+*/
+func (f *MessageSyntacticalValidator) validateUniqueness(msg *Message, peer *peer.Peer) bool {
+	// maps block type to the messageIDs it contains
+	blockMap := make(map[ParentsType]map[MessageID]types.Empty, 4)
+	for _, block := range msg.parentsBlocks {
```
karimodm, line 340: Using `ForEachParentByType` and `ForEachParent` would be more redeable.

packages/tangle/parser.go
```diff
@@ -248,6 +251,157 @@ type MessageFilter interface {
...
+		f.onRejectCallback(msg, ErrBlockCountMismatch, peer)
+		return
+	}
+
+	if !f.validateUniqueness(msg, peer) {
+		return
+	}
+
+	// Parent count per block 0 <= x <= 8
+	for _, block := range msg.parentsBlocks {
```
karimodm, line 299: I think you could use `ForEachParentByType` and `ForEachParent` to make the code more readeable.

packages/tangle/parser.go
```diff
@@ -533,6 +687,34 @@ var (
 
 	// ErrInvalidMessageAndTransactionTimestamp is returned when the message its transaction timestamps are invalid.
 	ErrInvalidMessageAndTransactionTimestamp = fmt.Errorf("invalid message and transaction timestamp")
+
+	// ErrNoStrongParents is returned when strong blocks are missing in messages
+	ErrNoStrongParents = errors.New("Missing strong messages in block")
```
karimodm, line 692: Error strings should start lowercase.

packages/tangle/message.go
GalRogozinski: seems like it...

Git blame tells me you added it (or modified it last) :detective: 
you want me to fix this as part of this PR?

packages/tangle/parser.go
GalRogozinski: The reasons are:
1. Currently other validation checks are done via filters. A maintainer would expect all validations to be done this way.
2. Were you referring to the validation done while parsing the message (were you?) imo adding checks there would just conflate between parsing and validation.
3. The filter framework currently creates events for rejection that we can hook into
4. If for some reason an api would be added to send a message to node directly (not just payload) this would become handy.

packages/tangle/message.go
karimodm: I think we should get rid of all unused and badly-named vars yeah! Especially if I wrote them :P

packages/tangle/parser.go
karimodm: Checks before were performed without the use of a filter; so the code is now checking twice: with the "old" mechanism and this new filter. I am not sure that the filter is the best approach, because the checks must be performed on parsing anyway, so these are essentially duplicated checks.

jonastheis: I understand syntactical validation to be part of the parsing to make sure that the bytes you're reading actually correspond to the bytes you're expecting to read (ie you know that what you're reading is a correct message). This includes loading an object from disk as well as receiving it from a neighbor and API. As such, this needs to happen somewhere within a call of `MessageFromBytes` to cover all these cases. A filter then simply does these checks redundantly, and is thus not necessary anymore.

jonastheis (changes requested):
As part of this PR also the `message_test.go` needs to be adjusted and probably new tests added to test the new message layout constraints.

packages/tangle/parser.go
GalRogozinski: I just solved a real silly bug that drove me crazy yesterday. It was really an obvious one in retrospect but it did cost me an entire day of work, There is a good chance that if I would've continued to mismash parsing and validation together this bug would've been left undiscovered...

The bug was that the parsed `parentsBlocksCount` was never assigned back to the `message` struct.
It was a bit elusive to me because since I saw it correctly parsed I assumed it was assigned to the message,

Now if we add more validation logic to the parsing code we will make the function longer, thus harder to maintain.
If I would do it the way you guys instructed me now then I might have added a call at the end of `MessageFromBytes` called `validate` to account for all the less trivial rules (uniqueness, lexicographical order), so I won't have reptitions. This approach is likely to lead to more bugs! This would be because `validate` would be missing validation logic that is implicit in `parseMessageBytes`.

Since `validate` is in package scope and we like to have flat design where everything is accessible it may get reused in the future. *Currently* you may say there is no need for it to get reused so it doesn't matter. But is it *guaranteed* to  stay this way? One should always assume he won't know the code requirements in the future.

So I believe that parsing and validation are actually two seperate things and we should resist the temptation of unifying them. As long as we adhere to certain principles (keeping functions short, with 1 clear-cut responsibility, and reusable) we will probably end up with a more maintainable and less bug-prone code...

Sorry for such a long writeup, but I think this kind of discussion saves time in the end :-)

packages/tangle/parser.go
```diff
@@ -533,6 +687,34 @@ var (
...
+	ErrParentsNotLexicographicallyOrdered = errors.New("Messages within blocks must be lexicographically ordered")
+
+	// ErrRepeatingBlockTypes is returned when the same block type appears more than once in a message
+	ErrRepeatingBlockTypes = errors.New("Block types within a message must not repeat")
+
+	// ErrRepeatingReferencesInBlock is returned when a parent appears more than once in a block
+	ErrRepeatingReferencesInBlock = errors.New("Duplicate parents in a message block")
+
+	// ErrRepeatingMessagesAcrossBlocks is returned when a parent appears more than once across blocks
+	ErrRepeatingMessagesAcrossBlocks = errors.New("Different blocks have repeating messages")
```
luca-moser, line 717: Error msgs should be lowercased.

luca-moser (changes requested):

packages/tangle/parser.go
```diff
@@ -248,6 +251,157 @@ type MessageFilter interface {
...
+		return
+	}
+
+	if !f.validateUniqueness(msg, peer) {
+		return
+	}
+
+	// Parent count per block 0 <= x <= 8
+	for _, block := range msg.parentsBlocks {
+		if block.ParentsCount < MinParentsCount || block.ParentsCount > MaxParentsCount {
```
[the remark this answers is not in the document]
GalRogozinski: Will be done as part of #1593 

Currents solution will be validation withing the constructor as discussed

packages/tangle/parser.go
GalRogozinski: using forEachParent now complicates the code

This is because it has a functional style that utilizes a consumer.
I want now to return from my upper-level function in the middle and this makes it hard w.o doing some tricks

besides as far as I understnad golang, the ethos is to keep things simple, and for loops are simple as it gets

packages/tangle/parser.go
GalRogozinski: done

packages/tangle/parser.go
```diff
@@ -248,6 +251,157 @@ type MessageFilter interface {
...
+**/
+func (f *MessageSyntacticalValidator) Filter(msg *Message, peer *peer.Peer) {
+	// A strong Block must exist
+	if len(msg.ParentsByType(StrongParentType)) == 0 {
+		f.onRejectCallback(msg, ErrNoStrongParents, peer)
+		return
+	}
+
+	// validate block count
+	if int(msg.parentsBlocksCount) != len(msg.parentsBlocks) {
```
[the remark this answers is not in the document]
GalRogozinski: Hmm it only holds it implicitely (you need to get the length of the slice)

The reason it is there as far as I undestand is because `parentsBlockCount` is part of the message layout and how it is serialized, and we want the struct to reflect that


I guess once we do validation via the serializer we can give up on this field and use only the slice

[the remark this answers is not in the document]
GalRogozinski: looking at the code again, I actually agree that it can be removed now already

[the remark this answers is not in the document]
GalRogozinski: I said I will remove it but:
1. There are too many changes in this PR
2. The current comment in the code that says that the struct should reflect what is sent over the wire made me feel a bit uneasy

packages/tangle/testutils.go
```diff
@@ -574,47 +574,58 @@ func nextSequenceNumber() uint64 {
...
-	return NewMessage(strongParents, weakParents, dislikeParents, likeParents, timestamp, ed25519.PublicKey{}, nextSequenceNumber(), payload.NewGenericDataPayload([]byte(payloadString)), 0, ed25519.Signature{})
+func newTestParentsDataMessageTimestampIssuer(payloadString string, strongParents, weakParents, dislikeParents, likeParents []MessageID, issuer ed25519.PublicKey, timestamp time.Time) *Message {
+	message, _ := NewMessage(strongParents, weakParents, dislikeParents, likeParents, timestamp, issuer, nextSequenceNumber(), payload.NewGenericDataPayload([]byte(payloadString)), 0, ed25519.Signature{})
+	return message
 }
 
-func newTestParentsDataMessageTimestampIssuer(payloadString string, strongParents, weakParents, dislikeParents, likeParents []MessageID, issuer ed25519.PublicKey, timestamp time.Time) *Message {
-	return NewMessage(strongParents, weakParents, dislikeParents, likeParents, timestamp, issuer, nextSequenceNumber(), payload.NewGenericDataPayload([]byte(payloadString)), 0, ed25519.Signature{})
+func newTestParentsDataWithTimestamp(payloadString string, strongParents, weakParents, dislikeParents, likeParents []MessageID, timestamp time.Time) *Message {
```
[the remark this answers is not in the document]
GalRogozinski: This is how it was in dev/consensus branch

not sure if this should be fixed in this PR or another

packages/tangle/message.go
```diff
@@ -1232,3 +1299,18 @@ func (c *CachedMessageMetadata) Consume(consumer func(messageMetadata *MessageMe
...
+var (
+	errNoStrongParents                    = errors.New("Missing strong messages in first parent block")
+	errBlocksNotOrderedByType             = errors.New("Blocks should be ordered in ascending order according to their type")
+	errBlockTypeIsUnknown                 = errors.Errorf("Block types must range from %d-%d", 0, numberOfBlockTypes-1)
+	errParentsCountMismatch               = errors.New("Number of parents in a message doesn't match parent count")
+	errParentsOutOfRange                  = errors.Errorf("A block must have at least %d-%d parents", MinParentsCount, MaxParentsCount)
+	errParentsNotLexicographicallyOrdered = errors.New("Messages within blocks must be lexicographically ordered")
+	errRepeatingBlockTypes                = errors.New("Block types within a message must not repeat")
+	errRepeatingReferencesInBlock         = errors.New("Duplicate parents in a message block")
+	errRepeatingMessagesAcrossBlocks      = errors.New("Different blocks have repeating messages")
```
luca-moser, line 1313: Error msgs need to be lowercased

luca-moser (changes requested):

packages/tangle/parser.go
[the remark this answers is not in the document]
luca-moser: Tracked here: https://github.com/iotaledger/goshimmer/issues/1607, can be done at a later point in time.

packages/tangle/message.go
```diff
@@ -41,14 +41,14 @@ const (
...
-	// defines the minimum number of parents each parents block must have.
-	MinParentsBlocksCount = 1
-
-	// defines the maximum number of parents each parents block must have.
-	MaxParentsBlocksCount = 4
-
 	// MinStrongParentsCount defines the minimum number of strong parents a message must have.
 	MinStrongParentsCount = 1
+
+	// numberOfBlockTypes counts StrongParents, WeakParents, DislikeParents, LikeParents
```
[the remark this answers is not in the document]
GalRogozinski: done

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+
+	// Block types must be ordered in ASC order and not repeat
+	for i := 0; i < len(parentsBlocks)-1; i++ {
+		if parentsBlocks[i].ParentsType == parentsBlocks[i+1].ParentsType {
+			return nil, errRepeatingBlockTypes
+		}
+		if parentsBlocks[i].ParentsType > parentsBlocks[i+1].ParentsType {
+			return nil, errBlocksNotOrderedByType
+		}
+		// we can skip the first block because we already acertained it is of StrongParentType
```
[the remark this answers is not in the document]
GalRogozinski: done

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+		if block.ParentsCount > MaxParentsCount || block.ParentsCount < MinParentsCount {
+			return nil, errParentsOutOfRange
+		}
+		// The lexicographical order check also makes sure there are no duplicates
+		for i := 0; i < len(block.References)-1; i++ {
+			if block.References[i].CompareTo(block.References[i+1]) == 0 {
+				return nil, errRepeatingReferencesInBlock
+			}
+			if block.References[i].CompareTo(block.References[i+1]) != -1 {
+				return nil, errParentsNotLexicographicallyOrdered
```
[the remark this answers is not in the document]
GalRogozinski: done

it was an if because originally there was only one case

packages/tangle/message.go
GalRogozinski: done

before I thought you just meant to make the vars unexported :-P
can we have our dog bark at this? I would expect this to be automated since making errors lower case is a golang thing..

luca-moser: I've opened https://github.com/iotaledger/goshimmer/issues/1608 to track this. Can you please also make the error variables exported again? This would allow to match against these error cases from outside this package.

packages/tangle/message.go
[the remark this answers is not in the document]
luca-moser: You can remove the `compare` var from the switch statement.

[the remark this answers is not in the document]
GalRogozinski: that's weird I actually originally did it w.o `compare` but the go linter went mad at me

now that you asked I tried again and now it is quiet :man_shrugging:

packages/tangle/message.go
GalRogozinski: We need this for integration tests maybe?

Because with our flat design it is anyhow pretty accessible

luca-moser: I don't know where these errors would be examined right now but I could imagine that some situation might arise. For example integration tests as you mentioned :)

luca-moser (approved):

packages/tangle/message.go
```diff
@@ -344,90 +408,93 @@ func MessageFromBytes(bytes []byte) (result *Message, consumedBytes int, err err
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse signature of the message: %w", err)
 	}
 
 	// retrieve the number of bytes we processed
 	readOffsetEnd := marshalUtil.ReadOffset()
 
 	// store marshaled version as a copy
-	result.bytes, err = marshalUtil.ReadBytes(readOffsetEnd-readOffsetStart, readOffsetStart)
+	msgBytes, err := marshalUtil.ReadBytes(readOffsetEnd-readOffsetStart, readOffsetStart)
```
karimodm, line 485: Do we check that there are no bytes left to process at this point?

karimodm (changes requested):

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+	return newMessageWithValidation(MessageVersion, parentsBlocks, issuingTime, issuerPublicKey, msgPayload, nonce, signature, sequenceNumber)
+}
+
+/**
+newMessageWithValidation creates a new message while performing ths following syntactical checks:
+1. A Strong Parents Block must exist.
+2. Parents Block types cannot repeat.
+3. Parent count per block 1 <= x <= 8.
+4. Parents unique within block.
+5. Parents lexicographically sorted within block.
```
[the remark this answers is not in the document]
GalRogozinski: 5 is about the order of the parents
8 is about the order of the blocks themselves

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+	// Block types must be ordered in ASC order and not repeat
+	for i := 0; i < len(parentsBlocks)-1; i++ {
+		if parentsBlocks[i].ParentsType == parentsBlocks[i+1].ParentsType {
+			return nil, ErrRepeatingBlockTypes
+		}
+		if parentsBlocks[i].ParentsType > parentsBlocks[i+1].ParentsType {
+			return nil, ErrBlocksNotOrderedByType
+		}
+		// we can skip the first block because we already ascertained it is of StrongParentType
+		if parentsBlocks[i+1].ParentsType >= numberOfBlockTypes {
```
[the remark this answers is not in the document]
GalRogozinski: I have an idea on how to create an automatic update for this variable..

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+5. Parents lexicographically sorted within block.
+6. ParentsCount within each block must match the parents actually parsed within the block.
+7. A Parent(s) repetition is only allowed when it occurs across Strong and Like parents
+8. Blocks should be ordered by type in ascending order
+**/
+func newMessageWithValidation(version uint8, parentsBlocks []ParentsBlock, issuingTime time.Time,
+	issuerPublicKey ed25519.PublicKey, msgPayload payload.Payload, nonce uint64,
+	signature ed25519.Signature, sequenceNumber uint64) (result *Message, err error) {
+	// Validate strong parent block
+	if len(parentsBlocks) == 0 || parentsBlocks[StrongParentType].ParentsType != StrongParentType ||
```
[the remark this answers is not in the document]
GalRogozinski: Unlike the other constructor `NewMessage`, this constructor doesn't sort or changes the contents of the inputs. It just checks their validity. If for example a deserialized message doesn't have StrongParentType as the first block it is invalid.

So I don't think I am making any assumptions here. I am simply checking if indeed we have a strong type and it is the first block.

packages/tangle/message.go
```diff
@@ -41,14 +41,14 @@ const (
...
-	// defines the maximum number of parents each parents block must have.
-	MaxParentsBlocksCount = 4
-
 	// MinStrongParentsCount defines the minimum number of strong parents a message must have.
 	MinStrongParentsCount = 1
+
+	// numberOfBlockTypes counts StrongParents, WeakParents, DislikeParents, LikeParents
+	numberOfBlockTypes = 4
+
+	// numberOfUniqueBlocks counts WeakParents and DislikeParents block.
```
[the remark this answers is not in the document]
GalRogozinski: done

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+		if parentsBlocks[i].ParentsType > parentsBlocks[i+1].ParentsType {
+			return nil, ErrBlocksNotOrderedByType
+		}
+		// we can skip the first block because we already ascertained it is of StrongParentType
+		if parentsBlocks[i+1].ParentsType >= numberOfBlockTypes {
+			return nil, ErrBlockTypeIsUnknown
+		}
+	}
+
+	for _, block := range parentsBlocks {
```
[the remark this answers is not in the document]
GalRogozinski: done

packages/tangle/message.go
```diff
@@ -290,20 +273,101 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
 		nonce:              nonce,
 		signature:          signature,
+	}, nil
+}
+
+// validate messagesIDs are unique across blocks
+// there may be repetition across strong and like parents
+func referencesUniqueAcrossBlocks(parentsBlocks []ParentsBlock) bool {
+	combinedMessageMap := make(map[MessageID]types.Empty, numberOfBlockTypes*MaxParentsCount)
+	parentsArray := make(MessageIDs, 0, MaxParentsCount*numberOfUniqueBlocks)
```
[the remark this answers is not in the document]
GalRogozinski: yes, but in the end I put everything in combined message map

What names do you want me to give? I can maybe have strongLikedMessageMap? uniqueParentsArray? and combinedMessageMap?
If you like the names give :thumbsup: and i will change.

Or offer different names :-)

packages/tangle/message.go
GalRogozinski: I didn't add it..

But if you run the test that checks just this on line 890 of message_test it passes just fine

packages/tangle/message.go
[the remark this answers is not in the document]
karimodm: You are not necessarily checking that is the first block: you are assuming that `StrongParentType` will always have index 0; that's why I find that second condition strange to read: it seems you are using `parentBlocks` as a map.
Since on line 316 you are checking for a count mismatch you should also check the parentBlockCount and parentsBlock index matches the ParentsType of the block.

packages/tangle/message.go
```diff
@@ -290,20 +275,104 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+
+/**
+newMessageWithValidation creates a new message while performing ths following syntactical checks:
+1. A Strong Parents Block must exist.
+2. Parents Block types cannot repeat.
+3. Parent count per block 1 <= x <= 8.
+4. Parents unique within block.
+5. Parents lexicographically sorted within block.
+6. ParentsCount within each block must match the parents actually parsed within the block.
+7. A Parent(s) repetition is only allowed when it occurs across Strong and Like parents
```
karimodm, line 291:
```suggestion
7. A Parent(s) repetition is only allowed when it occurs across Strong and Like parents.
```

packages/tangle/message.go
```diff
@@ -290,20 +275,104 @@ func NewMessage(strongParents, weakParents, dislikeParents, likeParents MessageI
...
+8. Blocks should be ordered by type in ascending order
```
karimodm, line 292:
```suggestion
8. Blocks should be ordered by type in ascending order.
```

packages/tangle/message.go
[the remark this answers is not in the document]
karimodm: Yep, you are right.

packages/tangle/message.go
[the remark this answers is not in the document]
karimodm: I would propose: `combinedParents` and `uniqueParents` or something like that. Pointing out the fact that you use the former to make repeated parents unique, and the latter to eventually count that the count of the parents is what you expect.

packages/tangle/message.go
karimodm: As far as I understand line 870 of message_test is not the same test: in that you are checking the the amount of bytes processed equals the amount of bytes in the message. I would like the parser to fail if the message has unprocessed trailing bytes, with a corresponding test.
I think it will make the parser more robust! What do you think?

packages/tangle/message.go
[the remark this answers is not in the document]
GalRogozinski: I can change StrongParentType to 0..
The linter might give me a magic number error but I can ignore it (which is why I didn't do so in the first place)..

I didn't understand what you wanted me to do on line 316 (now 318)..
`parentBlockCount` is not a parameter of the constructor. The deserializer uses it for parsing so it can't have a mismatch.
I don't think the parentBlock index must match the block type..

packages/tangle/message.go
GalRogozinski: opened https://github.com/iotaledger/goshimmer/issues/1617

karimodm (approved):
Looks good to me now, just this last pending point if you want to address it:
> As far as I understand line 870 of message_test is not the same test: in that you are checking the the amount of bytes processed equals the amount of bytes in the message. I would like the parser to fail if the message has unprocessed trailing bytes, with a corresponding test.
I think it will make the parser more robust! What do you think?

packages/tangle/message.go
GalRogozinski: I didn't forget, was just getting delayed...

Currently there is the following code:
```
// retrieve the number of bytes we processed
	readOffsetEnd := marshalUtil.ReadOffset()

	// store marshaled version as a copy
	msgBytes, err := marshalUtil.ReadBytes(readOffsetEnd-readOffsetStart, readOffsetStart)
	if err != nil {
		return nil, fmt.Errorf("error trying to copy raw source bytes: %w", err)
	}
```

It makes sure that we only consider the bytes that we read as the real bytes, meaning if there are more bytes in the buffer we simply trim them and never use them.
I think this works.. I wouldn't have doen it this way, but not sure if I should make more changes to this PR, especially since we will be soon have a PR rewriting the deserializer...

PR #1600, opened 2021-08-04
Improve spammer precision

### LINKED ISSUE #1592: Improve spammer precision
Use a ticker to drive the spammer.

daria305:
# Description of change
Fixes #1592 

 - Change the spammer tool to use Ticker instead of waiting to control the rate
 - Add optional parameter to set the rate unit, default is mps
`curl --location 'http://localhost:8080/spammer?cmd=start&rate=1000&imif=poisson&unit=mpm'`
`curl --location 'http://localhost:8080/spammer?cmd=start&rate=100'`
 - Add missing imif flag to web API ToggleSpammer() client API

## Type of change
- Documentation Update

HTTP, web API, spammer go script tested on local docker net

### DIFF
--- a/client/spammer.go
+++ b/client/spammer.go
@@ -12,19 +12,19 @@ const (
 )
 
 // ToggleSpammer toggles the node internal spammer.
-func (api *GoShimmerAPI) ToggleSpammer(enable bool, rate int, timeUnit string, imif string) (*jsonmodels.SpammerResponse, error) {
+func (api *GoShimmerAPI) ToggleSpammer(enable bool, rate int, unit, imif string) (*jsonmodels.SpammerResponse, error) {
 	// set default imif in case of incorrect imif value
 	if imif != "poisson" {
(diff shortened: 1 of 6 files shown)


capossele (changes requested):
Looks good overall. One comment though related to the rate now expressed in seconds. This limits the rate of the spammer as it cannot issue slower than 1 MPS and, depending on the aMana, that could be too much. Should we maybe find a solution to be more flexible regarding the setting of the rate?

daria305:
> Looks good overall. One comment though related to the rate now expressed in seconds. This limits the rate of the spammer as it cannot issue slower than 1 MPS and, depending on the aMana, that could be too much. Should we maybe find a solution to be more flexible regarding the setting of the rate?

I've added an optional parameter `unit` that by default is mps but can be changed to mpm.

jkrvivian (approved):

capossele (approved):

PR #1613, opened 2021-08-10
Feat/otv metrics

piotrm50:
# Description of change

Add new branch and message metrics to Grafana and Elasticsearch.

## Type of change

Manual tests in Docker network.

## Change checklist


plugins/metrics/message.go
```diff
@@ -70,9 +106,33 @@ var (
 	// current number of missing messages in missingMessageStorage
 	missingMessageCountDB atomic.Uint64
 
+	// current number of finalized messages
+	finalizedMessageCount = make(map[MessageType]uint64)
+
+	// protect map from concurrent read/write.
+	finalizedMessageCountMutex syncutils.RWMutex
+
+	// total time it took all messages to finalize. unit is milliseconds!
```
apenzk, line 115: we choose to go for ns everywhere
to be more precise we should use time.Duration where possible when discussing time elapsed or intervals, and use uint64 in nanoseconds where not possible

plugins/metrics/message.go
```diff
@@ -70,9 +106,33 @@ var (
...
+	messageFinalizationTotalTime = make(map[MessageType]uint64)
+
+	// protect map from concurrent read/write.
+	messageFinalizationTotalTimeMutex syncutils.RWMutex
+
+	// current number of confirmed  branches
+	confirmedBranchCount atomic.Uint64
+
+	// total time it took all branches to finalize. unit is milliseconds!
```
apenzk, line 124: same.. use nanosec or time.Duration

plugins/metrics/plugin.go
```diff
@@ -166,6 +168,46 @@ func registerLocalMetrics() {
...
+
+		var issuingTime time.Time
+		messagelayer.Tangle().Storage.Message(messageID).Consume(func(message *tangle.Message) {
+			message.ForEachParent(func(parent tangle.Parent) {
+				increasePerParentType(parent.Type)
+			})
+			issuingTime = message.IssuingTime()
+		})
+		if messagelayer.Tangle().Storage.MessageMetadata(messageID).Consume(func(messageMetadata *tangle.MessageMetadata) {
+			messageFinalizationTotalTime[messageType] += uint64(messageMetadata.FinalizedTime().Sub(issuingTime).Milliseconds())
```
apenzk, line 185: this should be done in nanoseconds

apenzk (changes requested):
you should check if we can use ns instead.. since ns is the unit of time.Duration. time.Duration is what we want to use for defining any time interval

plugins/metrics/message.go
piotrm50: Nanoseconds could overflow for uint64 and for metrics we don't need that much precision anyway.

plugins/metrics/plugin.go
piotrm50: Nanoseconds could overflow for uint64 and for metrics we don't need that much precision anyway.

plugins/metrics/message.go
```diff
@@ -48,6 +74,21 @@ var (
...
 	initialMessageTotalCountDB uint64
 
+	// total number of branches in the database at startup
+	initialBranchTotalCountDB uint64
+
+	// total number of finalized branches in the database at startup
+	initialFinalizedBranchCountDB uint64
+
+	// total number of confirmed branches in the database at startup
+	initialConfirmedBranchCountDB uint64
```
apenzk, line 84: i am not sure i understand this metric.. is this loaded from the snapshot? shouldnt the number of branches be zero on startup?

packages/remotemetrics/events.go
```diff
@@ -0,0 +1,97 @@
...
+	IssuedTimestamp    time.Time `json:"issuedTimestamp" bson:"issuedTimestamp"`
+	SolidTimestamp     time.Time `json:"solidTimestamp,omitempty" bson:"solidTimestamp"`
+	ScheduledTimestamp time.Time `json:"scheduledTimestamp" bson:"scheduledTimestamp"`
+	BookedTimestamp    time.Time `json:"bookedTimestamp" bson:"bookedTimestamp"`
+	ConfirmedTimestamp time.Time `json:"confirmedTimestamp" bson:"confirmedTimestamp"`
+	DeltaSolid         int64     `json:"deltaSolid,omitempty" bson:"deltaSolid"`
+	DeltaScheduled     int64     `json:"deltaArrival" bson:"deltaArrival"`
+	DeltaBooked        int64     `json:"deltaBooked" bson:"deltaBooked"`
+	DeltaConfirmed     int64     `json:"deltaConfirmed" bson:"deltaConfirmed"`
+	StrongParentCount  int       `json:"strongParentCount" bson:"strongParentCount"`
```
apenzk, line 54: If i understood correct this is counting the edges (strrong and below weak). If this is the case consider calling it StrongEdgeCount, since ParentCount could be confusing since a message would be considered multiple times as a parent.

apenzk (approved):
Seems fine. Only two comments: one regarding naming and one regarding a definition

plugins/metrics/message.go
piotrm50: It doesn't have to be, for example when a node is restarted with the same database. When running from a snapshot with a clean database then it will be zero.

PR #1615, opened 2021-08-10
Makes golint report upper case error msgs

luca-moser:
# Description of change

What's in the title.

## Type of change

Ran `golangci-lint` locally with and without the change on a file with an upper cased error msg and it did report when the confidence level was reduced.


karimodm (approved):
Default is 0.8 I see. Ok :cool:

PR #1618, opened 2021-08-11
Implements handle branch in FG

luca-moser:
Adds the "`FinalityGadget`" implementation to handle branch approval weight translation to grade of finality to the branch and its contained txs. Also modifies the `ApprovalWeightManager` to update branch weight of parents, even if the supporter was already present, in order to update the approval weight on parent branches when a supporter was added on a child branch.


karimodm (approved):
lgtm!

PR #1620, opened 2021-08-11
Refactor docker private network tool

luke-thorne:
# Description of change

Improve user experience with the Docker private network tool.

Extend the use of Docker Compose to better leverage configuration options available. Environment variables can be set to configure different aspects of Docker compose and [profiles](https://docs.docker.com/compose/profiles/) are implemented to group containers by capability, reducing the complexity of `run.sh`. Users are able to easily run a local GoShimmer network that is extensible. 

Extends documentation to cover how to run and configure a local network without the help of `run.sh`.

Exposes the dashboard from the `faucet` container instead of `peer_master` because running both the Analysis Dashboard and the old Dashboard causes the old Dashboard to malfunction. 

## Type of change

Tested by running docker compose without Grafana or DRNG containers, with only Grafana OR DRNG enabled, and with both Grafana AND DRNG enabled.

When Grafana is enabled, it creates a race condition where `peer_master` attempts to connect before MongoDB is available and crashes. This has been alleviated by having Docker Compose restart `peer_master` when it crashes with `restart: always`.

Updated `get_approval_csv.sh` to work with the new docker-compose configuration and verified all approvals are collected.

## Change checklist


documentation/docs/tooling/docker_private_network.md
```diff
@@ -106,4 +109,157 @@ Here is an example config that can be added:
...
+Docker Compose creates a `shimmerdb` volume to maintain a tangle even after tearing down the containers. Run `docker-compose down -v` to clear the volume.
+
+##### Example
+
+You can set the environment variable configuration inline as seen in this example.
+
+```bash
+GOSHIMMER_TAG=develop docker-compose up -d
+```
+
```
luca-moser, line 152: This bit appears 3x; either delete the other two sections or adjust the examples accordingly.

luca-moser (changes requested):

documentation/docs/tooling/docker_private_network.md
luke-thorne: I moved the sections so the example appears with the configuration options instead of with each container.

luca-moser (approved):

PR #1627, opened 2021-08-13
Refactors onTransactionConfirmed

luca-moser:
# Description of change

Makes it more readable by extracting the functionality of gathering the input infos into another function and also **fixes a bug** where the `totalAmount` was incorrectly incremented by the summed up `amount` of inputs instead of with the `balance` of the input.

## Type of change

I watch the CI pass.

### DIFF
--- a/plugins/messagelayer/mana_plugin.go
+++ b/plugins/messagelayer/mana_plugin.go
@@ -125,50 +125,9 @@ func onTransactionConfirmed(transactionID ledgerstate.TransactionID) {
 	Tangle().LedgerState.Transaction(transactionID).Consume(func(transaction *ledgerstate.Transaction) {
 		// holds all info mana pkg needs for correct mana calculations from the transaction
 		var txInfo *mana.TxInfo
+
 		// process transaction object to build txInfo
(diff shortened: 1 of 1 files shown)


plugins/messagelayer/mana_plugin.go
```diff
@@ -187,6 +146,40 @@ func onTransactionConfirmed(transactionID ledgerstate.TransactionID) {
...
+	for _, input := range transaction.Essence().Inputs() {
+		var inputInfo mana.InputInfo
+
+		Tangle().LedgerState.CachedOutput(input.(*ledgerstate.UTXOInput).ReferencedOutputID()).Consume(func(o ledgerstate.Output) {
+			inputInfo.InputID = o.ID()
+
+			// first, sum balances of the input, calculate total amount as well for later
+			o.Balances().ForEach(func(color ledgerstate.Color, balance uint64) bool {
+				inputInfo.Amount += float64(balance)
+				totalAmount += float64(balance)
```
capossele, line 160: good catch!

capossele (approved):

PR #1637, opened 2021-08-18
Remove ConfirmedConsumer from OutputMetadata

### LINKED ISSUE #1633: Remove ConfirmedConsumer from outputs
Remove the field `ConfirmedConsumer` from `OutputMetadata` (check json models, APIs, diagnostics, etc). Since this information is used when creating a global snapshot it needs to be adjusted as to retrieve this value on demand.

karimodm:
The `ConfirmedConsumer` property has been removed (along with `firstConsumer`) for the `ledgerstate.OutputMetadata` struct.
For the `jsonmodels.OutputMetadata` struct the  field is still present so it can be used from the Explorer; it is although filled using the new `LedgerState.ConfirmedConsumer` method.

Fixes #1633


packages/tangle/ledgerstate.go
```diff
@@ -266,6 +269,26 @@ func (l *LedgerState) Consumers(outputID ledgerstate.OutputID) (cachedTransactio
...
 }
 
+// ConfirmedConsumer returns the (cached) confirmed consumer of the given outputID.
+func (l *LedgerState) ConfirmedConsumer(outputID ledgerstate.OutputID) (confirmedConsumer *ledgerstate.CachedConsumer) {
+	found := false
+	for _, cachedConsumer := range l.Consumers(outputID) {
+		cachedConsumer.Consume(func(consumer *ledgerstate.Consumer) {
+			// We already identified a confirmed Consumer
+			l.CachedTransactionMetadata(consumer.TransactionID()).Consume(func(transactionMetadata *ledgerstate.TransactionMetadata) {
+				if l.tangle.ConfirmationOracle.IsBranchConfirmed(transactionMetadata.BranchID()) {
```
jonastheis, line 279: I'm not sure if that always works. What happens if the transaction is in the master branch? It might be necessary to add another method `IsTransactionConfirmed` to the `ConfirmationOracle`.

packages/tangle/ledgerstate.go
```diff
@@ -266,6 +269,26 @@ func (l *LedgerState) Consumers(outputID ledgerstate.OutputID) (cachedTransactio
 	return l.UTXODAG.CachedConsumers(outputID)
 }
 
+// ConfirmedConsumer returns the (cached) confirmed consumer of the given outputID.
+func (l *LedgerState) ConfirmedConsumer(outputID ledgerstate.OutputID) (confirmedConsumer *ledgerstate.CachedConsumer) {
+	found := false
+	for _, cachedConsumer := range l.Consumers(outputID) {
+		cachedConsumer.Consume(func(consumer *ledgerstate.Consumer) {
+			// We already identified a confirmed Consumer
```
jonastheis, line 277: ? The comment seems out of place.

packages/tangle/ledgerstate.go
```diff
@@ -266,6 +269,26 @@ func (l *LedgerState) Consumers(outputID ledgerstate.OutputID) (cachedTransactio
...
+			l.CachedTransactionMetadata(consumer.TransactionID()).Consume(func(transactionMetadata *ledgerstate.TransactionMetadata) {
+				if l.tangle.ConfirmationOracle.IsBranchConfirmed(transactionMetadata.BranchID()) {
+					found = true
+					confirmedConsumer = cachedConsumer
```
jonastheis, line 281: You already consumed `cachedConsumer` and therefore it might be not in cache anymore when it is returned. I'd suggest changing the method to return a `transactionID`.

packages/jsonmodels/ledgerstate.go
```diff
@@ -484,32 +484,21 @@ type OutputMetadata struct {
 	Solid              bool                `json:"solid"`
 	SolidificationTime int64               `json:"solidificationTime"`
 	ConsumerCount      int                 `json:"consumerCount"`
-	FirstConsumer      string              `json:"firstConsumer,omitempty"`
 	ConfirmedConsumer  string              `json:"confirmedConsumer,omitempty"`
 	GradeOfFinality    gof.GradeOfFinality `json:"gradeOfFinality"`
 }
 
 // NewOutputMetadata returns the OutputMetadata from the given ledgerstate.OutputMetadata.
 func NewOutputMetadata(outputMetadata *ledgerstate.OutputMetadata) *OutputMetadata {
```
jonastheis, line 492: add `confirmedConsumer TransactionID` as a parameter here instead of the need to implicitly set it.

jonastheis (changes requested):

packages/tangle/ledgerstate.go
```diff
@@ -194,17 +194,15 @@ func (l *LedgerState) SnapshotUTXO() (snapshot *ledgerstate.Snapshot) {
...
-					tx, exist := copyLedgerState[outputMetadata.ConfirmedConsumer()]
-					// ignore consumers that are not confirmed long enough or even in the future.
-					if !exist || startSnapshot.Sub(tx.Essence().Timestamp()) < minAge {
-						unspentOutputs[i] = true
-						includeTransaction = true
-					}
+			unspentOutputs[i] = true
+			includeTransaction = true
+
+			l.ConfirmedConsumer(output.ID()).Consume(func(consumer *ledgerstate.Consumer) {
```
luca-moser, line 200: If I read `ConfirmedConsumer()` right, it could be that the `*ConfirmedConsumer` returned is nil if there is none, in which case this `Consume` would cause a panic?

plugins/dashboard/explorer_routes.go
```diff
@@ -250,10 +250,15 @@ func findAddress(strAddress string) (*ExplorerAddress, error) {
 		// how much pending mana the output has?
 		pendingMana, _ := messagelayer.PendingManaOnOutput(output.ID())
 
+		jsonMetadata := jsonmodels.NewOutputMetadata(metaData)
+		messagelayer.Tangle().LedgerState.ConfirmedConsumer(output.ID()).Consume(func(consumer *ledgerstate.Consumer) {
```
luca-moser, line 254: Same here, could be that there is no `ConfirmedConsumer` causing a nil pointer panic.

luca-moser (changes requested):
The places where `LedgerState.ConfirmedConsumer` is used are susceptible to nil pointer panics as you can not assume that there is a confirmed consumer of an output.

packages/tangle/ledgerstate.go
karimodm: I have to have a serious look on how this ObjectStorage thing works!

packages/tangle/ledgerstate.go
```diff
@@ -266,6 +270,23 @@ func (l *LedgerState) Consumers(outputID ledgerstate.OutputID) (cachedTransactio
 	return l.UTXODAG.CachedConsumers(outputID)
 }
 
+// ConfirmedConsumer returns the (cached) confirmed consumer of the given outputID.
```
jonastheis, line 273: Update comment

packages/jsonmodels/ledgerstate.go
```diff
@@ -484,32 +484,20 @@ type OutputMetadata struct {
...
-	// omit firstConsumer field if it hasn't been consumed yet
-	if outputMetadata.FirstConsumer().Base58() != ledgerstate.GenesisTransactionID.Base58() {
-		firstConsumer = outputMetadata.FirstConsumer().Base58()
-	}
-	confirmedConsumer := ""
-	// omit confirmedConsumer field if it hasn't been consumed yet by a confirmed transaction
-	if outputMetadata.ConfirmedConsumer().Base58() != ledgerstate.GenesisTransactionID.Base58() {
-		confirmedConsumer = outputMetadata.ConfirmedConsumer().Base58()
-	}
+func NewOutputMetadata(outputMetadata *ledgerstate.OutputMetadata, confirmedConsumer *ledgerstate.Consumer) *OutputMetadata {
```
jonastheis, line 493: I don't think this is practical at all. The consumer object contains redundant information as this is the output metadata. It should be sufficient to just pass the `transactionID` and take the base58 representation for this json model.

packages/tangle/ledgerstate.go
```diff
@@ -266,6 +270,23 @@ func (l *LedgerState) Consumers(outputID ledgerstate.OutputID) (cachedTransactio
...
+func (l *LedgerState) ConfirmedConsumer(outputID ledgerstate.OutputID) (consumerID ledgerstate.TransactionID) {
+	found := false
+	// default to no consumer, i.e. Genesis
+	consumerID = ledgerstate.GenesisTransactionID
```
jonastheis, line 277: Maybe compare the consumerID instead of having an additional bool?

plugins/webapi/ledgerstate/plugin.go
```diff
@@ -334,7 +335,13 @@ func GetOutputMetadata(c echo.Context) (err error) {
 	}
 
 	if !messagelayer.Tangle().LedgerState.CachedOutputMetadata(outputID).Consume(func(outputMetadata *ledgerstate.OutputMetadata) {
-		err = c.JSON(http.StatusOK, jsonmodels.NewOutputMetadata(outputMetadata))
+
+		confirmedConsumerID := messagelayer.Tangle().LedgerState.ConfirmedConsumer(outputID)
+		confirmedConsumer := ledgerstate.NewConsumer(outputID, confirmedConsumerID, types.True)
```
jonastheis, line 340: Simplify with passing just transactionID to the json model.

plugins/dashboard/explorer_routes.go
```diff
@@ -250,10 +251,14 @@ func findAddress(strAddress string) (*ExplorerAddress, error) {
 		// how much pending mana the output has?
 		pendingMana, _ := messagelayer.PendingManaOnOutput(output.ID())
 
+		// obtain information about the consumer of the output being considered
+		confirmedConsumerID := messagelayer.Tangle().LedgerState.ConfirmedConsumer(output.ID())
+		confirmedConsumer := ledgerstate.NewConsumer(output.ID(), confirmedConsumerID, types.True)
```
jonastheis, line 256: Simplify with passing just transactionID to the json model.

jonastheis (changes requested):

jonastheis (approved):

luca-moser (approved):

PR #1639, opened 2021-08-18
Adds finality gadget unit test

luca-moser:
# Description of change

Moves the `ApprovalWeightManager`'s `ProcessMessage` test into a `Scenario` struct abstracting the steps taken during such tests. Due to import cycle and the Tangle's package structure, the scenario code had to be moved into `testutil.go`. The `SimpleFinalityGadget` test then builds on top of the `ProcessMessage` scenario used to test the `ApprovalWeightManager` by asserting that the correct events got fired and that the messages, branches, transactions and outputs got the right grade of finality set on each step of the test scenario.

This PR also adds:
- the typical way using `Option` funcs to configure the `SimpleFinalityGadget`
- triggering of `BranchConfirmed` events
- a bug fix where before GoF updates on a marker would leak to parents which themselves were markers
- a bug fix where the origin branch's transaction id was used to adjust output GoF instead of the candidate transaction's during forward propagation
- triggering of `TransactionConfirmed` events when the forward propagation of a branch's GoF update causes the tx to also gain a high enough GoF. This checks beforehand however whether the attachments of the txs do have a high enough GoF as to not make it so that a tx can have a higher GoF than the msg containing it.
- Bumps AW threshold for what is considered `High` to `>50`

## Type of change

The test I wrote.


packages/consensus/finality/finality_gadget_test.go
```diff
@@ -0,0 +1,393 @@
...
+			t.Log(err)
+		}
+	}))
+
+	// we need to update the WeightProvider on confirmation
+	fg.Events().MessageConfirmed.Attach(events.NewClosure(func(messageID tangle.MessageID) {
+		testTangle.Storage.Message(messageID).Consume(func(message *tangle.Message) {
+			testTangle.WeightProvider.Update(message.IssuingTime(), identity.NewID(message.IssuerPublicKey()))
+		})
+	}))
```
jonastheis, line 392: Do we really need to do this? Doesn't it just use the static weight defined in the scenario?

packages/consensus/finality/finality_gadget.go
```diff
@@ -152,8 +209,8 @@ func (s *SimpleFinalityGadget) HandleMarker(marker *markers.Marker, aw float64)
 	}
 
 	propagateGoF := func(message *tangle.Message, messageMetadata *tangle.MessageMetadata, w *walker.Walker) {
-		// stop walking to past cone if reach a marker
-		if messageMetadata.StructureDetails().IsPastMarker && messageMetadata.GradeOfFinality() == gradeOfFinality {
+		// stop walking to past cone if reach a marker with a higher grade of finality
+		if messageMetadata.StructureDetails().IsPastMarker && messageMetadata.GradeOfFinality() > gradeOfFinality {
```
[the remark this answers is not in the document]
luca-moser: good catch, you are right.

packages/consensus/finality/finality_gadget_test.go
luca-moser: Removed as no longer necessary

jonastheis (approved):

PR #1655, opened 2021-08-24
Consensus bugfixes

jonastheis:
# Description of change
This PR fixes a number of issues related to finality and the new consensus:
- Fix bug where a branch's weight is not yet tracked and MaxFloat64 is returned as its weight
- Propagate GoF also via liked parents
- Add only relevant supporters as supporters for markers
- Fix and simplify logic when setting a payload's GoF to always use `min(branchGoF,messageGoF)`
- Update branch GoF (on the corresponding transaction) before propagating it in the UTXO future cone
- Temporary fix to allow gaps in marker sequences 
- Avoid revisiting of markers when propagating AW

## Type of change

Manual debugging and testing. Also, all test are still passing.


packages/consensus/finality/finality_gadget.go
```diff
@@ -336,10 +342,15 @@ func (s *SimpleFinalityGadget) setMessageGoF(messageMetadata *tangle.MessageMeta
...
-				gradeOfFinality = s.opts.BranchTransFunc(branchID, s.tangle.ApprovalWeightManager.WeightOfBranch(branchID))
+			// A transaction can't have a higher GoF than its branch, thus we need to evaluate based on min(branchGoF,messageGoF).
+			// This also works for transactions in MasterBranch since it has gof.High and we then use the messageGoF.
+			branchGoF, err := s.tangle.LedgerState.UTXODAG.BranchGradeOfFinality(transactionMetadata.BranchID())
+			if err != nil {
+				// TODO: properly handle error
+				panic(err)
+			}
+			if branchGoF < gradeOfFinality {
+				gradeOfFinality = branchGoF
```
luca-moser, line 353: Couldn't this cause that a message's GoF overrides a branch GoF which is higher? Imagine a scenario where a branch collected enough weight in other parts of the Tangle through reattachments of the same transaction,

packages/markers/sequence.go
```diff
@@ -164,7 +164,8 @@ func (s *Sequence) IncreaseHighestIndex(referencedMarkers *Markers) (index Index
 		panic("tried to increase Index of wrong Sequence")
 	}
 
-	if increased = referencedSequenceIndex == s.highestIndex; increased {
+	// TODO: this is a quick'n'dirty solution and should be revisited.
```
luca-moser, line 167: Can you expand the comment what is happening and what it is trying to solve?

luca-moser (changes requested):

packages/consensus/finality/finality_gadget.go
jonastheis: Ok, I adjusted this to take reattachments into account via `min(branchGoF,max(messageGoF,transactionGoF))`. This works because the `transactionGoF` should be greater than the current `messageGoF` if there is another attachment with a higher GoF.

packages/markers/sequence.go
jonastheis: Done.

luca-moser (approved):
Lets introduce a panic if we hit the invalid invariant of `transactionGoF` > `branchGoF`. Other than that LGTM.

packages/consensus/finality/finality_gadget.go
```diff
@@ -354,14 +370,6 @@ func (s *SimpleFinalityGadget) setPayloadGoF(messageID tangle.MessageID, gradeOf
 						outputMetadata.SetGradeOfFinality(gradeOfFinality)
```
karimodm, line 370: Can we use `adjustOutputGoF`?

karimodm (changes requested):

karimodm (approved):

Issue #1660, opened 2021-08-27
Remove MongoDB 


piotrm50:
MongoDB is only used to collect data about FPC conflicts. For new consensus mechanism it will not be needed and can be entirely removed from the code, docker-compose and existing deployments. 

TODO:

- [x] remove mongoDB from local docker network scripts
- [x] remove `peer-entry.sh` and `wait-for.sh` scripts in docker network [here](https://github.com/iotaledger/goshimmer/blob/develop/tools/docker-network/builder) and update docker network's `docker-compose.yaml` accordingly.
- [x] remove mongoDB from predevnet deployment [here](https://github.com/iotaledger/goshimmer/tree/dev/consensus/deploy/ansible)
- [x] remove mongodb from the devnet deployment and internal testnet running on `ressims.iota.cafe` - I think this needs to be done after the release of the new consensus mechanism.

jonastheis:
1 and 2 out of the list above have been already addressed with https://github.com/iotaledger/goshimmer/pull/1657 due to incompatibility with the `distroless` image.

piotrm50:
Addressed the third point from the list with https://github.com/iotaledger/goshimmer/pull/1703
The last one should be resolved after the release of the new consensus.

PR #1661, opened 2021-08-27
Fix: plugin were skipped in docker network

daria305:
# Description of change
After introducing waiting for MongoDB script in 1e39278709dcabb297432dfe18f01a5aac9e6c7e there was a problem that some pulgins were not starting in the docker network

```
command: >
      /run/entrypoint.sh /go/bin/goshimmer
```
The script did not handle the parameters correctly because of the spaces in the plugin name: "webapi tools endpoint" this PR is the temporary solution. It just moves the webapi plugin name at the end, so plugins listed after that are not skipped.

Proper solution - removing MongoDB will be addressed by issue: #1660

## Type of change

### DIFF
--- a/tools/docker-network/docker-compose.yml
+++ b/tools/docker-network/docker-compose.yml
@@ -38,7 +38,7 @@ services:
       --metrics.local=false
       --metrics.manaResearch=false
       --node.disablePlugins=clock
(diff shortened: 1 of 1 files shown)


jonastheis:
The `wait-for` script was already removed with https://github.com/iotaledger/goshimmer/pull/1657 due to incompatibility with the `distroless` image. So I think this can be closed.

PR #1662, opened 2021-08-27
Broadcast Plugin

arne-fuchs:
# Description of change

Added a plugin to broadcast every message on the message layer to connections over a server socket on port 5050. The messages will be plain send so they can be processed in a dedicated application. 

## Type of change

The repo is online since weeks and has been cloned and used several times. I am writing on Java Library for it right now.

The install and usage instructions and project can be found here:
https://github.com/arne-fuchs/broadcast

## Change checklist

- [X] My code follows the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes //Never tested

### DIFF
--- a/plugins/broadcast/README.md
+++ b/plugins/broadcast/README.md
@@ -1,0 +1,36 @@
+# broadcast
+A plugin for IOTA's GoShimmer Node to broadcast every message on the message layer an write it to active tcp connections over port 5050
+
+## Installation
+Move the project's folder into your goshimmer/plugins/ folder.
+
+In goshimmer/plugins/research.go add the following line:
+```
+broadcast.Plugin(),
+```
+in the node.Plugins(...) list.
+
+You may need to recompile the goshimmer software.
+
+In the config.json you need to add "broadcast" to the "node" sections as followed:
+```
+"node": {
+    "disablePlugins": [],
+    "enablePlugins": ["broadcast"]
+},
+```
+
+## Usage
+Just connect to the plugin's port 5050 and you get the messages in real time as long as you are connected.
+A maximum of 256 Connections are possible before it throws errors.
+
+## Donations
+If you want to keep me motivated to do more open source stuff you can donate me some IOTA's. Even very small amounts makes me happy:
+
+```
+iota1qqvrqjfscx5ax7vnt8mmtmzj30af3xf7zfm8t7lnaxyrt73awgqckz02upv
+```
+
+## GitHub Project
+
+https://github.com/arne-fuchs/broadcast
--- a/plugins/broadcast/plugin.go
+++ b/plugins/broadcast/plugin.go
@@ -3,9 +3,9 @@ package broadcast
 import (
 	"github.com/iotaledger/goshimmer/plugins/broadcast/server"
 	"github.com/iotaledger/goshimmer/plugins/config"
-	flag "github.com/spf13/pflag"
 	"sync"
 
+	"github.com/iotaledger/hive.go/configuration"
 	"github.com/iotaledger/hive.go/daemon"
 	"github.com/iotaledger/hive.go/events"
 	"github.com/iotaledger/hive.go/logger"
@@ -17,12 +17,11 @@ import (
 )
 
 const (
-	pluginName = "Broadcast"
+	pluginName  = "Broadcast"
 	bindAddress = "broadcast.bindAddress"
 )
 
 var (
-	// plugin is the plugin instance of the activity plugin.
 	plugin *node.Plugin
 	once   sync.Once
 
@@ -32,20 +31,26 @@ var (
 // Plugin gets the plugin instance.
 func Plugin() *node.Plugin {
 	once.Do(func() {
-		plugin = node.NewPlugin(pluginName, node.Enabled, configure, run)
+		plugin = node.NewPlugin(pluginName, node.Disabled, configure, run)
 	})
 	return plugin
 }
 
+// ParametersDefinition contains the configuration parameters used by the plugin
+type ParametersDefinition struct {
+	Port int `default:"5050" usage:"port for broadcast plugin connections"`
+}
+
+var parameters = &ParametersDefinition{}
+
 // Handler functions
 func init() {
-	flag.String(bindAddress, ":5050", "the bind address for the broadcast plugin")
+	configuration.BindParameters(parameters, "Broadcast")
 }
 
 // Configure events
 func configure(_ *node.Plugin) {
 	plugin.LogInfof("starting node with broadcast plugin")
-	log = logger.NewLogger(pluginName)
 }
 
 //Run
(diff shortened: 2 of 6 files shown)


plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+
+	bufferedConnDataReceived, bufferedConnClosed := connectionList[idx-1].readLoop()
+
+	select {
+	case data := <-bufferedConnDataReceived:
+		//No input required. For debugging it will be printed
+		connectionList[idx-1].log.Infof("Data received:%v", data)
+		connectionList[idx-1].active = false
+	case <-shutdownSignal:
+		connectionList[idx-1].log.Infof("Shutdown signal received")
```
luca-moser, line 93: Why does the shutdown signal trigger this for loop? Won't you end up doing the same for X amount of connections while the "shutdown"? Perhaps this is simply a miss copy-and-paste from the cleanup code from `Listen()`?

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+		for i := 0; i < idx; i++ {
+			connectionList[i].active = false
+		}
+		return
```
luca-moser, line 97: unneeded return statement

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+	case <-bufferedConnClosed:
+		connectionList[idx-1].log.Errorf("Connection lost")
+		connectionList[idx-1].active = false
```
luca-moser, line 100: This line can basically be put after this switch statement as you are doing it for every case.

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+		return
+	}
+}
+
+func (connection *Connection) readLoop() (chan []byte, chan bool) {
+	bufferedConnDataReceived := make(chan []byte)
+	bufferedConnClosed := make(chan bool)
+
+	go func() {
+		{
```
luca-moser, line 110: Remove these blocks and give distinct names for the created closures.

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+				d := make([]byte, len(data))
+				copy(d, data)
+				bufferedConnDataReceived <- d
+			})
+			connection.bufferedConn.Events.ReceiveMessage.Attach(cl)
+			defer connection.bufferedConn.Events.ReceiveMessage.Detach(cl)
+		}
+
+		if err := connection.bufferedConn.Read(); err != nil {
+			if err != io.EOF && !strings.Contains(err.Error(), "Use of closed network connection") {
```
luca-moser, line 127: Instead of checking the string of the error, use `!errors.Is(nestedErr, ErrClosed)` where `ErrClosed` is from the `net` package.

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+	return bufferedConnDataReceived, bufferedConnClosed
+}
+
+func Broadcast(data []byte) {
+	connectionListMutex.Lock()
+	defer connectionListMutex.Unlock()
+
+	idx := int(index.Load())
+	for i := 0; i < idx; i++ {
+		if connectionList[i].active {
```
luca-moser, line 143: negate and use `continue`

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+			_, err := connectionList[i].bufferedConn.Write(data)
+			if err != nil {
```
luca-moser, line 145: Truncate to one line

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,165 @@
...
+		newIndex := 0
+		var newConnectionList = [256]Connection{}
+		for i := 0; i < idx; i++ {
+			if connectionList[i].active {
+				newConnectionList[newIndex] = connectionList[i]
+				newIndex++
+			}
+		}
+		index.Store(uint32(newIndex))
+		connectionList = newConnectionList
```
luca-moser, line 163: refactor this cleanup into a separate function

plugins/broadcast/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	once.Do(func() {
+		plugin = node.NewPlugin(pluginName, node.Enabled, configure, run)
+	})
+	return plugin
+}
+
+// Handler functions
+func init() {
+	flag.String(bindAddress, ":5050", "the bind address for the broadcast plugin")
+}
```
luca-moser, line 43: Should be using a `Parameters` struct as in other plugins.

plugins/broadcast/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	plugin *node.Plugin
+	once   sync.Once
+
+	log *logger.Logger
+)
+
+// Plugin gets the plugin instance.
+func Plugin() *node.Plugin {
+	once.Do(func() {
+		plugin = node.NewPlugin(pluginName, node.Enabled, configure, run)
```
luca-moser, line 35: This plugin should be disabled per default.

luca-moser (changes requested):
Hey @arne-fuchs, thanks for this PR.
After you've fixed the review comments we will be merging this plugin. However please note that it will likely be removed in the future as we implement event based publishing of data through MQTT.

arne-fuchs:
Ok! I'll get it done until next Week.

plugins/broadcast/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Handler functions
+func init() {
+	flag.String(bindAddress, ":5050", "the bind address for the broadcast plugin")
+}
+
+// Configure events
+func configure(_ *node.Plugin) {
+	plugin.LogInfof("starting node with broadcast plugin")
+	log = logger.NewLogger(pluginName)
```
jonastheis, line 48: Please use the integrated logger everywhere instead of creating a `log`.

plugins/broadcast/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+	if err != nil {
+		log.Errorf("Failed to start Broadcast daemon: %v", err)
+	}
+
+
+	//Get Messages from node
+	notifyNewMsg := events.NewClosure(func(messageID tangle.MessageID) {
+		messagelayer.Tangle().Storage.Message(messageID).Consume(func(message *tangle.Message) {
+			server.Broadcast(message.Bytes())
```
jonastheis, line 72: Make sure that this is **not** blocking. Events are synchronous calls and if the plugin blocks the goroutine, the message can't continue in the data flow.

plugins/broadcast/plugin.go
arne-fuchs: I am already using the logger from "github.com/iotaledger/hive.go/logger" as just like in other plugins

arne-fuchs: Ok, correction: using the new logger now.
Not many plugins have that implemented that yet

plugins/broadcast/server/server.go
arne-fuchs: I see no use in that; doesn't it make it way more complex than it is?

plugins/broadcast/server/server.go
arne-fuchs: It only happens 2 out of 3 cases.

arne-fuchs:
It should be ok now. Plaese review.

plugins/broadcast/plugin.go
arne-fuchs: made method call asynchronous - reolving conversation

plugins/broadcast/README.md
```diff
@@ -0,0 +1,36 @@
+# broadcast
+A plugin for IOTA's GoShimmer Node to broadcast every message on the message layer an write it to active tcp connections over port 5050
```
luca-moser, line 2:
```suggestion
A plugin for IOTA's GoShimmer Node to broadcast every message on the message layer and write it to active TCP connections over port 5050.
```
```suggestion
A plugin for IOTA's GoShimmer Node to broadcast every message on the message layer an write it to active tcp connections over port 5050
```

plugins/broadcast/server/server.go
```diff
@@ -0,0 +1,164 @@
...
+		return fmt.Errorf("failed to start Broadcast daemon: %w", err)
+	}
+
+	go func() {
+		for {
+			connection, err := listener.Accept()
+			if err != nil {
+				if connection != nil {
+					log.LogInfof("Couldn't accept connection: %s", err)
+				}
```
luca-moser, line 44: `connection` will probably ending up `nil` if `err` is `nil`, so remove this if check on the `connection`

plugins/broadcast/server/server.go
luca-moser: You did not remove the blocks.

luca-moser (changes requested):

luca-moser (approved):

PR #1667, opened 2021-08-31
Feat: New solidification logic

hmoog:
# Description of change

This PR modifies the solidification logic to not solidify past weak parents. This allows us to use the weak parents as a way to exclude undesired messages from the tangle (e.g. as part of the congestion control).

It introduces the following changes to the definition of being solid:

1. A message is solid, if its strong parents are solid and its weak parents are at least weakly solid.
2. A message is weakly solid if it either has a data payload or if its weak parents are at least weakly solid.
3. A message is invalid if any of its weak parents doesn't pass the parents age check, or if it doesn't have weak parents and its strong parents don't pass the parents age check.

If the Solidifier requests missing weak parents, then he skips downloading their strong past cone unless another message references the same message with a strong parent.

The UTXODAG now also has a concept of Solidity and instead of booking the Transactions, we just store them and let them wait to become solid (we do not actively request transactions).

Messages only become solid after their transaction has become solid, which means that by the time we reach the Booker, the Transaction is already fully booked and associated to a branch.

## Type of change

## Change checklist

### DIFF
--- a/packages/ledgerstate/temp.go
+++ b/packages/ledgerstate/temp.go
@@ -1,0 +1,69 @@
+package ledgerstate
+
+import (
+	"sync"
+
+	"github.com/iotaledger/hive.go/events"
+)
+
+// region eventsQueue //////////////////////////////////////////////////////////////////////////////////////////////////
+
+// eventsQueue represents an Event
+type eventsQueue struct {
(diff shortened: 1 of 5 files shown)


hmoog:
Had to revert last commit

PR #1681, opened 2021-09-09
Adds consensus integration test

luca-moser:
- Adds a simple consensus integration test where two nodes with 40%/20% consensus mana weight respectively issue conflicting txs and see that over time the one issued by the 40% node gains high GoF, while the other one results in "none" GoF.


tools/integration-tests/tester/tests/consensus/consensus_test.go
```diff
@@ -0,0 +1,150 @@
...
+		res1, err := node1.GetTransactionMetadata(tx1.ID().Base58())
+		require.NoError(t, err)
+		res2, err := node2.GetTransactionMetadata(tx2.ID().Base58())
+		require.NoError(t, err)
+		return res1.BranchID != ledgerstate.MasterBranchID.String() &&
+			res2.BranchID != ledgerstate.MasterBranchID.String()
+	}, tests.Timeout, tests.Tick)
+
+	// we issue msgs on both nodes so the txs' GoF can change, given that they are dependent on their
+	// attachments' GoF. if msgs would only be issued on node 2 or 1, they weight would never surpass 50%.
```
jonastheis, line 106: Is that true? If node 2 behaves honest it'd issue a message and if the branch is not liked it should set a like switch to the message of tx1 so that the branch would end up having 60% of weight, no? Either way issuing a bunch of messages on both nodes is also fine, just maybe update the comment.

luca-moser: You are correct, that the **branch** of `tx1` would indeed gain 60% AW if only node 2 would issue messages and not both nodes. However, the GoF of a transaction is dependent on the attachment sites, therefore `tx1-msg` would not gain more AW since we only update message GoF based on markers. So in this case, future markers of `tx1-msg` would never have more than 20% AW, hence not allowing the AW of `tx1-msg` to gain over 40% and subsequent `tx1` to adjust its GoF.

PR #1691, opened 2021-09-17
Fix cMana weight provider

jonastheis:
# Description of change
This PR introduces an `activityLog` per node for the cMana weight provider. In that way we can track all node activity (with a specified `granularity`) in the relevant time (from `TT-activeTimeThreshold` until `TT`). In case there is node activity in the future from the perspective of the `TangleTime`, it is stored in the `activityLog` and once the `TT` moves forward the node can stay `active` according to information in the log.

In the previous implementation we lost this information since we only stored a single time per node and did not store any active times in the future of the current `TT`. In this way nodes joining the network later could end up with a bug where 
1. no node is seen as active anymore if the first confirmed message was created by a node without cMana
2. a single node has all active cMana (as interpreted by the weight provider) if the first confirmed message was created by a node with **any cMana**. This node would be seen as having all the active weight in this case.

Add activity log to cMana weight provider to avoid bug where node activity is not properly tracked due to only storing one timestamp per node

## Type of change

## Change checklist

### DIFF
--- a/packages/tangle/cmanaweightprovider.go
+++ b/packages/tangle/cmanaweightprovider.go
@@ -1,11 +1,13 @@
 package tangle
 
 import (
+	"container/heap"
 	"sync"
 	"time"
 
 	"github.com/cockroachdb/errors"
 	"github.com/iotaledger/hive.go/cerrors"
+	"github.com/iotaledger/hive.go/datastructure/set"
 	"github.com/iotaledger/hive.go/identity"
 	"github.com/iotaledger/hive.go/kvstore"
 	"github.com/iotaledger/hive.go/marshalutil"
@@ -24,7 +26,7 @@ const (
 type CManaWeightProvider struct {
 	store             kvstore.KVStore
 	mutex             sync.RWMutex
-	activeNodes       map[identity.ID]time.Time
+	activeNodes       map[identity.ID]*activityLog
 	manaRetrieverFunc ManaRetrieverFunc
 	timeRetrieverFunc TimeRetrieverFunc
 }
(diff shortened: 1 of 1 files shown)


packages/tangle/cmanaweightprovider.go
```diff
@@ -88,14 +99,19 @@ func (c *CManaWeightProvider) WeightsOfRelevantSupporters() (weights map[identit
...
+
+		// Skip node if it does not fulfill minimumManaThreshold.
+		if nodeMana <= minimumManaThreshold {
 			continue
 		}
 
-		nodeMana := mana[nodeID]
-		if nodeMana <= minimumManaThreshold {
+		// Determine whether node was active in time window.
+		if active, empty := al.Active(targetTime.Add(-activeTimeThreshold), targetTime); !active {
```
luca-moser, line 111: Extract `targetTime.Add(-activeTimeThreshold)` to a var before the `for` loop as it is going to be the same for each iteration.

packages/tangle/cmanaweightprovider.go
```diff
@@ -177,3 +195,105 @@ func activeNodesToBytes(activeNodes map[identity.ID]time.Time) []byte {
...
+
+// Active returns true if the node was active between lower and upper bound.
+// It cleans up the log on the fly, meaning that old/stale times are deleted.
+// If the log ends up empty after cleaning up, empty is set to true.
+func (a *activityLog) Active(lowerBound, upperBound time.Time) (active, empty bool) {
+	lb, ub := timeToUnixGranularity(lowerBound), timeToUnixGranularity(upperBound)
+
+	for a.times.Len() > 0 {
+		// Get the lowest element of the min-heap = the earliest time.
+		activity := (*a.times)[0]
```
luca-moser, line 247: Rename this `activity` to `earliestActivity`

luca-moser (changes requested):

karimodm (approved):
We should extend the test set, but it looks good to me!

PR #1717, opened 2021-09-23
LB role using Traefik

karimodm:


luca-moser (approved):

deploy/ansible/roles/lb/templates/asset-registry.yml.j2
```diff
@@ -0,0 +1,29 @@
+#jinja2: trim_blocks:True, lstrip_blocks:True
+http:
+  routers:
+    asset-registry-router:
+      rule: "Host(`asset-registry.tokenizedassetsdemo.iota.cafe`) || Host(`asset-registry.nectar.iota.cafe`)"
```
georgysavva, line 5: Why does it say `nectar` in the `asset-registry.nectar.iota.cafe` domain name?
We have an asset-registry service on devnet. Besides that shouldn't it be templated instead?

PR #1723, opened 2021-09-27
Use multi-arch 1.17.1 base image

luca-moser:
We had the 1.17-buster linux/amd64 image defined in combination with GOOS=linux/GOARCH=amd64; this however caused my Macbook Air M1 to no longer be able to compile/run the tools/docker-network as it would lead to a qemu segmentation fault. Changing our base image to the multi-arch 1.17.1-buster image and removing the explicit GOOS/GOARCH envs in the build RUN step, resolved these issues for me.

In theory there never was a need to have GOOS/GOARCH defined in the build RUN step, since it would be set correctly automatically by whatever golang image would be pulled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 # syntax = docker/dockerfile:1.2.1
 
 ############################
-# golang 1.17-buster amd64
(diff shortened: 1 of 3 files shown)


karimodm (approved):
Yep, runtime is multiarch too :heavy_check_mark:

PR #1730, opened 2021-10-05
Fixes nil pointer when an AW reorg is detected

luca-moser:
Check beforehand whether the remotelogconn is available.


plugins/remotelog/plugin.go
```diff
@@ -17,11 +17,8 @@ import (
 	"github.com/iotaledger/hive.go/node"
 	"github.com/iotaledger/hive.go/workerpool"
 	"go.uber.org/dig"
-	"gopkg.in/src-d/go-git.v4"
```
karimodm, line 20: Package should not have been removed. Used in `getGitInfo`

karimodm (approved):

PR #1779, opened 2021-11-05
Fix parameters defaults & deploy code

karimodm:
# Description of change

Change a bunch  of deploy stuff and defaults to launch OTV in good shape.

### DIFF
--- a/plugins/messagelayer/parameters.go
+++ b/plugins/messagelayer/parameters.go
@@ -20,7 +20,7 @@ type ParametersDefinition struct {
 	}
 
 	// TangleTimeWindow defines the time window in which the node considers itself as synced according to TangleTime.
-	TangleTimeWindow time.Duration `default:"2m" usage:"the time window in which the node considers itself as synced according to TangleTime"`
+	TangleTimeWindow time.Duration `default:"10m" usage:"the time window in which the node considers itself as synced according to TangleTime"`
 
 	// StartSynced defines if the node should start as synced.
 	StartSynced bool `default:"false" usage:"start as synced"`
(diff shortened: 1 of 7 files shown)


plugins/pow/parameters.go
```diff
@@ -9,7 +9,7 @@ import (
 // ParametersDefinition contains the definition of the parameters used by the pow plugin.
 type ParametersDefinition struct {
 	// Difficulty defines the PoW difficulty.
-	Difficulty int `default:"22" usage:"PoW difficulty"`
```
jonastheis, line 12: This should also be changed in `config.default.json` and possibly across documentation

plugins/messagelayer/parameters.go
```diff
@@ -20,7 +20,7 @@ type ParametersDefinition struct {
 	}
 
 	// TangleTimeWindow defines the time window in which the node considers itself as synced according to TangleTime.
-	TangleTimeWindow time.Duration `default:"2m" usage:"the time window in which the node considers itself as synced according to TangleTime"`
```
jonastheis, line 23: This should only be changed for our deployment. In general I think 2min is okay

plugins/pow/parameters.go
karimodm: Done, I couldn't find references to this number in the docs.

plugins/messagelayer/parameters.go
karimodm: Done

jonastheis (approved):

jonastheis (approved):

PR #1783, opened 2021-11-06
Update setup.md

MaKla89:
# Description of change

Added info on potentially necessary configuration for Docker on IPv6-only-networks as a warning INFO box. Docker does not support IPv6 containers per default, IPv6 support needs to be enabled manually as explained in the Docker documentation.


karimodm (approved):
Thanks for the contribution!

PR #1794, opened 2021-11-11
Remove orderer and add dispatcher

### LINKED ISSUE #1841: Mark messages ready
Enqueue messages to buffer (Scheduler.Submit(id)) after booking.
Mark messages ready (Scheduler.Ready(id)), when their parents are scheduled or confirmed.

capossele:
# Description of change

Fixes #1841, #1849 and #1839

- Remove orderer
- Add dispatcher that uses a workerPool to dispatch scheduled messages to gossip and TSA
- Make use of Ready flag in the scheduler to schedule messages as described in #1841

## Type of change


capossele:
Fixes #1841 and #1849

packages/tangle/scheduler.go
```diff
@@ -122,15 +122,34 @@ func (s *Scheduler) Setup() {
...
+
+		if s.isReady(messageID) {
+			if err := s.Ready(messageID); err != nil {
+				s.Events.Error.Trigger(errors.Errorf("failed to mark %s as ready: %w", messageID, err))
+			}
+		}
 	}))
 
+	onMessageScheduledOrConfirmed := func(messageID MessageID) {
+		s.tangle.Storage.Approvers(messageID).Consume(func(approver *Approver) {
```
piotrm50, line 141: Check if the message has been scheduled already to avoid iterating through approvers twice - after scheduling and confirmation.

piotrm50 (changes requested):

piotrm50 (approved):

piotrm50 (approved):

packages/tangle/scheduler.go
```diff
@@ -122,15 +126,32 @@ func (s *Scheduler) Setup() {
...
+		s.updateDependencies(messageID, s.tryReady(messageID))
 	}))
 
+	onMessageScheduled := func(messageID MessageID) { s.updateDependencies(messageID, nil) }
+	s.tangle.Scheduler.Events.MessageScheduled.Attach(events.NewClosure(onMessageScheduled))
+
+	onMessageConfirmed := func(messageID MessageID) {
+		var scheduled bool
+		s.tangle.Storage.MessageMetadata(messageID).Consume(func(messageMetadata *MessageMetadata) {
+			scheduled = messageMetadata.Scheduled() || messageMetadata.ScheduledBypass() // TODO: remove bypass check
```
GalRogozinski, line 146: A good rule on a PR that should be merged is to remove all TODOs

if this is supposed to be done as part of another issue maybe do the "ISSUE xxx:" comment

packages/tangle/scheduler.go
```diff
@@ -245,6 +266,73 @@ func (s *Scheduler) Clear() {
 	}
 }
 
+// isEligible returns true if the given messageID has either been scheduled or confirmed.
+func (s *Scheduler) isEligible(messageID MessageID) (eligible bool) {
+	s.tangle.Storage.MessageMetadata(messageID).Consume(func(messageMetadata *MessageMetadata) {
+		eligible = messageMetadata.Scheduled() ||
+			s.tangle.ConfirmationOracle.IsMessageConfirmed(messageID) ||
+			messageMetadata.ScheduledBypass() // TO BE REMOVED
```
GalRogozinski, line 274: ditto

packages/tangle/scheduler.go
```diff
@@ -245,6 +266,73 @@ func (s *Scheduler) Clear() {
...
+
+	if len(dependencies) == 0 {
+		if _, exists := s.parentsMap[messageID]; !exists {
+			return
+		}
+
+		for _, childID := range s.parentsMap[messageID] {
+			s.tryReady(childID)
+		}
+		delete(s.parentsMap, messageID)
```
GalRogozinski, line 309: This is the only place where we delete from the map correct?
Is it possible to have a memory overflow attack? 

For example, let say someone sends unsolid messages to the node and overflows the map?
Maybe there are also other conditions in which the map won't be pruned...

Btw when we delete from a map in go, it will actually evict the memory? I recall we had some issues with that...

GalRogozinski (commented):
Looks nice

Just answer my comments so I can give an approval :-)

packages/tangle/scheduler.go
```diff
@@ -122,15 +126,32 @@ func (s *Scheduler) Setup() {
 			return
 		}
 
-		if err := s.SubmitAndReady(messageID); err != nil {
+		if err := s.Submit(messageID); err != nil {
 			if !errors.Is(err, schedulerutils.ErrBufferFull) &&
 				!errors.Is(err, schedulerutils.ErrInboxExceeded) &&
 				!errors.Is(err, schedulerutils.ErrInsufficientMana) {
```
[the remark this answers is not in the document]
piotrm50: AFAIK those are already logged when they're thrown. However, I removed most of those in a separate PR, as currently the only possible error will be `schedulerutils.ErrInsufficientMana` and it will be logged immediately.

packages/tangle/scheduler.go
piotrm50: Good point. The idea was to merge everything and then remove the scheduler bypassed flag to avoid potential conflicts if we did that before merging everything to `dev/drop-head`

packages/tangle/dispatcher.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"sync"
+
+	"github.com/iotaledger/hive.go/events"
+	"github.com/iotaledger/hive.go/workerpool"
+)
+
+const (
+	messageWorkerCount     = 1
+	messageWorkerQueueSize = 1024
```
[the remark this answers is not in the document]
capossele: That's a good question! Looking at what we did in the gossip queue, we have 1000 there, so maybe we should do the same here? wdyt?

packages/tangle/scheduler.go
capossele: Unsolid messages would not reach the scheduler. This map should always be very small as messages are in general processed already in order. However, if the map should get big, we might want to re-think to use an iteration over the approver instead of using this reverse-map

capossele: Anyway, I've reverted to use an iteration over the approvers as it seems from the failing tests that the parentsMap was buggy

packages/tangle/dispatcher.go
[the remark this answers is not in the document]
GalRogozinski: Hmm first I need to ask why we define a size in the first place...
I am guessing the size in the gossip queue achieves 2 purposes:
1. Rate limiter - if messages arrive faster than they are processed the workerpool blocks
2. Saves us from memory overflow

But we also have rate setting mechanism as a function of mana correct? Is it implemented yet?
So the size of the queue just have to be large enough to accomodate that rate

Now looking at the code I understand that all the work in the gossip queue is to trigger an event and you have several go routines that do that for some reason...

The dispatcher is farhter down the pipeline and we don't want it to jam it. So it makes sense it is a bit bigger than the gossip queue. Maybe I would decrease the number of workers of the gossip message worker, but not so important for the prototype I guess...

packages/tangle/scheduler.go
```diff
@@ -334,6 +290,23 @@ func (s *Scheduler) unreadyParents(messageID MessageID) (parents []MessageID) {
...
+		}
+	}
+}
+
+// updateApprovers iterates over the direct approvers of the given messageID and
+// tries to mark them as ready.
+func (s *Scheduler) updateApprovers(messageID MessageID) {
+	s.tangle.Storage.Approvers(messageID).Consume(func(approver *Approver) {
+		s.tryReady(approver.approverMessageID)
+	})
```
GalRogozinski, line 307: maybe the approver has one parent confirmed and the other not confirmed? This is enough gor try ready?

packages/tangle/scheduler.go
```diff
@@ -135,21 +131,23 @@ func (s *Scheduler) Setup() {
...
 		s.tangle.Storage.MessageMetadata(messageID).Consume(func(messageMetadata *MessageMetadata) {
-			scheduled = messageMetadata.Scheduled() || messageMetadata.ScheduledBypass() // TODO: remove bypass check
+			scheduled = messageMetadata.Scheduled() || messageMetadata.ScheduledBypass() // Issue #1855 TODO: remove bypass check
 		})
 		if scheduled {
 			return
 		}
-		onMessageScheduled(messageID)
+		s.tangle.Storage.Message(messageID).Consume(func(message *Message) {
+			s.unsubmit(message)
```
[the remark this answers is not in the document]
capossele: If a message has been confirmed and not scheduled yet but about to be marked as ready, it should just be unsubmitted as it would be filter out from the dispatcher anyway

packages/tangle/scheduler.go
capossele: tryReady calls isReady that checks ALL of its parents

GalRogozinski (approved):

piotrm50 (approved):

karimodm:
This fixes #1839

PR #1801, opened 2021-11-11
Fix: correct & expose API port for automated DB, snapshot dumps and dRANDs

karimodm:


deploy/ansible/roles/snapshotter/templates/goshimmer_snapshot.bin_dump.sh.j2
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-HTTP_STS_CODE="$(/usr/bin/curl -s -o /dev/null -w '%{http_code}\n' http://ressims.iota.cafe:9101/healthz)"
-LATEST_SNAPSHOT="http://127.0.0.1:9101/snapshot"
+HTTP_STS_CODE="$(/usr/bin/curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/healthz)"
```
georgysavva, line 3: Why it's the localhost instead of `ressims.iota.cafe`?

capossele: I don't think we use ressims for anything DevNet related don't we?

capossele (approved):
This PR looks very **random**

deploy/ansible/roles/snapshotter/templates/goshimmer_snapshot.bin_dump.sh.j2
karimodm: We want to check that the node we are going to get the snapshot from is in sync: localhost. Ressims does not run anything related to our networks anymore.

georgysavva (approved):

Issue #1460, opened 2021-06-17
Clean-up/merging branches


capossele:
Clean-up the branches once they get confirmed or rejected

- [ ] Re-map the branches and update metadata

hmoog:
```go
func OnBranchConfirmed(branchID BranchID) {
    for oldBranch, newBranch := range branchDAG.MergeToMaster(branchID) {
        // remap branches using markers
        // remap branches using individually mapped messages (example: introduce a conflict)
    }
}
```

PR #1846, opened 2021-11-16
Conflict Spam Integration Test

### LINKED ISSUE #1595: Adjust consensus integration tests
## Simple double spend with spends in UTXO future cone
1. create deterministic partition (check with active cMana weights)
1. issue conflict in each partition with transactions spending from them
1. merge partitions
1. check that conflict is resolved based on higher weights
2. verify that all transactions in winning UTXO future cone have high GoF others have no GoF

## Spamming of conflicts
1. Spam 100 conflicts (double spends, triple, pairwise conflicting, other more sophisticated conflict structures, ...)
2. Check that all conflicts are consistently resolved in all nodes

GalRogozinski:
# Description of change

Fixes #1595

Adds pairwise and triplets conflicts

## Type of change

- Integration Test

## Change checklist


tools/integration-tests/tester/tests/testutil.go
```diff
@@ -231,6 +233,59 @@ func SendFaucetRequest(t *testing.T, node *framework.Node, addr ledgerstate.Addr
...
+		outputs = append(outputs, output)
+	}
+
+	// create tx essence
+	txEssence := ledgerstate.NewTransactionEssence(0, time.Now(), manaPledgeID,
+		manaPledgeID, inputs, ledgerstate.NewOutputs(outputs...))
+
+	// create signatures
+	unlockBlocks := []ledgerstate.UnlockBlock{}
+	for i := 0; i < len(inputs); i++ {
```
jonastheis, line 276: Same here: Since we know the length and iterate over the index it'd be better to initialize with make imo and then assign to the corresponding index instead of using append.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/iotaledger/goshimmer/tools/integration-tests/tester/tests"
+	"github.com/stretchr/testify/require"
+)
+
+// constant var, shouldn't be changed
+var tokensPerRequest int
+
+/**
+TestConflictSpam spams a node with conflicts and makes sure the GoFs are the same across the network
+*/
```
jonastheis, line 23: Please use line comment style.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+		addresses[i] = &address
+		keyPairs[address.String()] = peer1.KeyPair(uint64(i))
+	}
+
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
+	const conflictRepetitions = 4
+	pairwiseOutputs := outputs[:(conflictRepetitions-1)*3+3]
+	tripletOutputs := outputs[len(pairwiseOutputs):]
```
jonastheis, line 59: I don't get this. Could you add a comment in code why and how the outputs are "split" like that?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	txs := []*ledgerstate.Transaction{}
+	for i := 0; i < conflictRepetitions; i++ {
+		sendPairWiseConflicts(t, n.Peers(), pairwiseOutputs[i*3:i*3+3], keyPairs, &txs, i)
+		sendTripleConflicts(t, n.Peers(), tripletOutputs[i*3:i*3+3], keyPairs, &txs, i)
```
jonastheis, line 64: `&txs` doesn't look right. A slice is already a pointer.

jonastheis (changes requested):

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
+	const conflictRepetitions = 4
+	pairwiseOutputs := outputs[:(conflictRepetitions-1)*3+3]
+	tripletOutputs := outputs[len(pairwiseOutputs):]
+
+	txs := []*ledgerstate.Transaction{}
+	for i := 0; i < conflictRepetitions; i++ {
+		sendPairWiseConflicts(t, n.Peers(), pairwiseOutputs[i*3:i*3+3], keyPairs, &txs, i)
```
karimodm, line 63: This `i*3:i*3+3` is very hard to understand; refactor a bit.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+		address := peer1.Address(i)
+		addresses[i] = &address
+		keyPairs[address.String()] = peer1.KeyPair(uint64(i))
+	}
+
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
+	const conflictRepetitions = 4
+	pairwiseOutputs := outputs[:(conflictRepetitions-1)*3+3]
```
karimodm, line 58: What are you doing here? Simplify the code and comment what you are doing.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	for i := 0; i < 100; i++ {
+		address := peer1.Address(i)
+		addresses[i] = &address
+		keyPairs[address.String()] = peer1.KeyPair(uint64(i))
+	}
+
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
+	const conflictRepetitions = 4
```
karimodm, line 57: Put the const together with top variables.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	tests.AwaitInitialFaucetOutputsPrepared(t, faucet, n.Peers())
+
+	fundingAddress := peer1.Address(0)
+	tests.SendFaucetRequest(t, peer1, fundingAddress)
+	require.Eventually(t, func() bool {
+		return tests.Balance(t, peer1, fundingAddress, ledgerstate.ColorIOTA) >= uint64(tokensPerRequest)
+	}, tests.Timeout, tests.Tick)
+
+	addresses := make([]*ledgerstate.Address, 100)
+	keyPairs := map[string]*ed25519.KeyPair{}
```
karimodm, line 47: Use `make`

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/**
+sendPairWiseConflicts receives a list of outputs controlled by a peer with certain peer index.
+It send them all to addresses controlled by the next peer, but it does so several time to create pairwise conflicts.
+The conflicts are TX_B<->TX_A<->TX_C
+*/
+func sendPairWiseConflicts(t *testing.T, peers []*framework.Node, outputs ledgerstate.Outputs,
+	keyPairs map[string]*ed25519.KeyPair,
+	txs *[]*ledgerstate.Transaction, iteration int) {
```
karimodm, line 101: A pointer to an array of pointers to Transactions? Arrays are pointers already.

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -231,6 +233,59 @@ func SendFaucetRequest(t *testing.T, node *framework.Node, addr ledgerstate.Addr
...
 }
 
+// region CreateTransaction from outputs //////////////////////////////
+
+/**
+CreateTransactionFromOutputs takes the given utxos inputs and create a transaction tnat spread the total input balance
+across the targetAddresses. In order to correctly sign we have akeyPair map that maps a given address to its public key.
+Access and Consensus Mana is pledged to the node we specify.
+*/
+func CreateTransactionFromOutputs(t *testing.T, manaPledgeID identity.ID, targetAddresses []*ledgerstate.Address, keyPairs map[string]*ed25519.KeyPair, utxos ...ledgerstate.Output) *ledgerstate.Transaction {
```
karimodm, line 243: Is it really necessary to pass the map of addresses to keypair here? Why cannot this function itself get the keypair? It would hide some complexity from the calling functions.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	require.Eventually(t, func() bool {
+		return tests.Balance(t, peer1, fundingAddress, ledgerstate.ColorIOTA) >= uint64(tokensPerRequest)
+	}, tests.Timeout, tests.Tick)
+
+	addresses := make([]*ledgerstate.Address, 100)
+	keyPairs := map[string]*ed25519.KeyPair{}
+	for i := 0; i < 100; i++ {
+		address := peer1.Address(i)
+		addresses[i] = &address
+		keyPairs[address.String()] = peer1.KeyPair(uint64(i))
```
karimodm, line 51: Can you grab these keys from a deeper context? Such as `CreateTransactionFromOutputs`. It doesn't look clean to me that top-level tests functions have to deal with KeyPairs themselves. Top-level test functions should only deal with top-level actions, to make the test readeable.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
```
karimodm, line 56: Don't deal with keys here.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+			t.Logf("wallet output is %v", walletOutput)
+			output, err := walletOutput.Output.ToLedgerstateOutput()
+			require.NoError(t, err, "Failed to convert output to ledgerstate output")
+			outputs = append(outputs, output)
+		}
+	}
+	return outputs
+}
+
+func splitToAddresses(t *testing.T, node *framework.Node, output ledgerstate.Output, keyPairs map[string]*ed25519.KeyPair, addresses ...*ledgerstate.Address) ledgerstate.Outputs {
```
karimodm, line 180: Don't deal with keypairs here. Lower level functions should do the dirty job.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	outputs := getOutputsControlledBy(t, peer1, fundingAddress)
+	fundingKeyPair := map[string]*ed25519.KeyPair{fundingAddress.String(): peer1.KeyPair(0)}
+	outputs = splitToAddresses(t, peer1, outputs[0], fundingKeyPair, addresses...)
+	const conflictRepetitions = 4
+	pairwiseOutputs := outputs[:(conflictRepetitions-1)*3+3]
+	tripletOutputs := outputs[len(pairwiseOutputs):]
+
+	txs := []*ledgerstate.Transaction{}
```
karimodm, line 61: Use a `make`

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -231,6 +233,59 @@ func SendFaucetRequest(t *testing.T, node *framework.Node, addr ledgerstate.Addr
...
+		output := ledgerstate.NewSigLockedColoredOutput(ledgerstate.NewColoredBalances(outBalances), *targetAddresses[i])
+		outputs = append(outputs, output)
+	}
+
+	// create tx essence
+	txEssence := ledgerstate.NewTransactionEssence(0, time.Now(), manaPledgeID,
+		manaPledgeID, inputs, ledgerstate.NewOutputs(outputs...))
+
+	// create signatures
+	unlockBlocks := []ledgerstate.UnlockBlock{}
```
karimodm, line 275: Use `make` throughout the code for arrays and maps

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	tests.AwaitInitialFaucetOutputsPrepared(t, faucet, n.Peers())
+
+	fundingAddress := peer1.Address(0)
+	tests.SendFaucetRequest(t, peer1, fundingAddress)
+	require.Eventually(t, func() bool {
+		return tests.Balance(t, peer1, fundingAddress, ledgerstate.ColorIOTA) >= uint64(tokensPerRequest)
+	}, tests.Timeout, tests.Tick)
+
+	addresses := make([]*ledgerstate.Address, 100)
```
karimodm, line 46: Why Address pointers? Do you plan to modify them from another function?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+				require.EqualValues(t, prevGoF, metadata.GradeOfFinality,
+					"Different gofs on tx %s between peers %s and %s", tx.ID().Base58(),
+					peers[i-1].Name(), peer.Name())
+			}
+			prevGoF = metadata.GradeOfFinality
+		}
+	}
+}
+
+/**
```
karimodm, line 94: make this comment go-like

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	tx1 := tests.CreateTransactionFromOutputs(t, peers[0].ID(), targetAddresses, keyPairs, outputs...)
+	tx2 := tests.CreateTransactionFromOutputs(t, peers[1].ID(), targetAddresses, keyPairs, outputs[0])
+	tx3 := tests.CreateTransactionFromOutputs(t, peers[2].ID(), targetAddresses, keyPairs, outputs[2])
+
+	*txs = append(*txs, tx1, tx2, tx3)
+
+	PostTransactions(t, peers, peerIndex, "pairwise conflicts", tx1, tx2, tx3)
+}
+
+/**
```
karimodm, line 118: make this comment go-like

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: Do you need the concept of `iteration` here? Does this function really need to know about what step of the for loop you are calling it from?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+Creates conflicts as so
+TX_A<->TX_B TX_B<->TX_C TX_C<->TX_A
+*/
+func sendTripleConflicts(t *testing.T, peers []*framework.Node, outputs ledgerstate.Outputs,
+	keyPairs map[string]*ed25519.KeyPair,
+	txs *[]*ledgerstate.Transaction, iteration int) {
```
karimodm, line 124: Do you need the concept of `iteration` here?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+		require.Empty(t, resp.Error, "%s: There was an error in the response while posting transaction %s to peer %s",
+			attackName, tx.ID().Base58(), peers[peerIndex].Name())
+		time.Sleep(time.Second)
+	}
+}
+
+func determineTargets(peers []*framework.Node, iteration int) []*ledgerstate.Address {
+	targetIndex := (iteration + 1) % len(peers)
+	targetPeer := peers[targetIndex]
+	targetAddresses := []*ledgerstate.Address{}
```
karimodm, line 157: Turn into a `make` and don't use pointers.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+		t.Logf("%s: post tx %s on peer %s", attackName, tx.ID().Base58(), peers[peerIndex].Name())
+		require.NoError(t, err, "%s: There was an error posting transaction %s to peer %s",
+			attackName, tx.ID().Base58(), peers[peerIndex].Name())
+		require.Empty(t, resp.Error, "%s: There was an error in the response while posting transaction %s to peer %s",
+			attackName, tx.ID().Base58(), peers[peerIndex].Name())
+		time.Sleep(time.Second)
+	}
+}
+
+func determineTargets(peers []*framework.Node, iteration int) []*ledgerstate.Address {
```
karimodm, line 154: Do you need `iteration` here?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
```diff
@@ -0,0 +1,185 @@
...
+	targetIndex := (iteration + 1) % len(peers)
+	targetPeer := peers[targetIndex]
+	targetAddresses := []*ledgerstate.Address{}
+
+	for i := iteration * 3; i < iteration*3+3; i++ {
```
karimodm, line 159: Many "magic" numbers. What are you doing here?

karimodm (changes requested):

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/iotaledger/hive.go/crypto/ed25519"
```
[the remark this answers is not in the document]
GalRogozinski: Moved it

However, I noticed that the dog didn't bark, and when I do `goimports -w tester/tests/consensus/consensus_conflict_spam_test.go` or "optimize imports" command it doesn't move automatically...

I don't see a style-guide in the repo... so currently I just moved to where you think you wanted me to move it. Since `goimports` also accepts it I think I did fine.

Back when I was reviewing code all I cared is that the imports are in the order compatible with everyone's IDE settings. So that people won't reorganize imports on every commit and I could easily see changes in dependencies. 

So is there some setting I am missing. I just believe that import order is should be something purely automatic and no one should spend time commenting about it or  fixing it like I am doing now :-)

tools/integration-tests/tester/tests/testutil.go
```diff
@@ -231,6 +233,59 @@ func SendFaucetRequest(t *testing.T, node *framework.Node, addr ledgerstate.Addr
 	return resp.ID, sent
 }
 
+// region CreateTransaction from outputs //////////////////////////////
+
+/**
+CreateTransactionFromOutputs takes the given utxos inputs and create a transaction tnat spread the total input balance
+across the targetAddresses. In order to correctly sign we have akeyPair map that maps a given address to its public key.
+Access and Consensus Mana is pledged to the node we specify.
+*/
```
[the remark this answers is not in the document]
GalRogozinski: Thanks! didn't know that comment rule. You think we can make the dog bark about this?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: Not exactly. A slice consists of a pointer to an array and a header that contains the length and capacity of the array.  
Without `&` I would be copying the pointer and the header. Then the append operation I am doing to the slice would update the header copy but not the original slice.

This is according to https://go.dev/blog/slices-intro

jonastheis: That's right. Though it's not very readable imo. Maybe better to return the created txs in a slice and just append the returned slice. Something like this:
```go
tx = append(tx, sendTripleConflicts(t, n.Peers(), tripletOutputs[i*3:i*3+3], keyPairs, i)...)
```

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: 1. You meant slices are pointers. Since we are appending to the slice and passing it a long functions it must be a pointer to a slice. This is because a slice is not exactly a pointer. It is a pointer + header (that consists of length and capacity). Since we are appending to the slice we shouldn't be only chaging the different header copies.

2. Curerently `iteration` is good for logging and `determineTargets` function needs it

tools/integration-tests/tester/tests/testutil.go
GalRogozinski: how would it do it?  I create the keypairs together with the addresses. After the addresses are created I am not sure how I can get the corresponding keypairs. Unless I save the index for each address and poass it down but it is also ugly to me.

In most tests we use the wallet, but in this test I wanted to have a finer grain control on the outputs than the wallet allows. It was mainly because initially I was tryiong to do even fancier stuff that didn't work out for the test. 
Maybe think of the map and outputs as an `OutputWallet`?

I can also create an OutputWallet struct that contains all the outputs and their keys (combining 2 variables here).
Not sure if this is what you have in mind so I'll be waiting for your answer.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: see above answer

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: I changed to `make` in several places. Here I would rather not.
If I understand correctly you want to use `make` in order to create an array of certain predefined size and use indexes. This in order to decrease undelrying `malloc` operations.

But in this case, it will make the test harder to change. Because the size of txs slice may change if we add more conflict types or do other changes.

...Or maybe you wanted me to use `make` to create a 0 size slice? But then what is the advantage of using this syntax? Is it just a style thing that we have that I don't know about?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: I am kinda of always passing pointers as part of a java mindset that I still have. It is just cheaper to copy them then the actual structs.
I guess for this script it doesn't matter whether we use memory wisely and I can remove the pointers because it is maybe more readable\safer...

My question is what do we do for goshimmer? If we play it safe and not pass transactions by pointers then we bloat memory and strain the GC. If we pass pointers we are less safe since tx are supposed to be immutable. 
I would wrap them in an immutable struct that only exports getters if goshimmer would be production ready. I guess since goshimmer isn't production ready we are just doing the safe approach which is simple and bloaty?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: since the `i` in the loop won't start with 0 I think this way is more readable.
I'll remove the pointer after I get an answer about the bloat vs. safety question I asked you on another question.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: yup

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: yup

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: 1. Having this txs pointer so you can append doesn't seem very readable to me. Maybe you can append in the caller and have this function returned the txs it created?
2. Same principle here: you are dealing with iteration of a for loop in the caller inside the called function: I find it hard to read. Maybe the log messages can be factored out as part of the main loop?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: Can you factor it out? It seems this function shouldn't need to know about how you are iterating from the caller.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: Is it possible to factor it out somehow?

tools/integration-tests/tester/tests/testutil.go
karimodm: If I understand correctly, this function is meant to distribute the balances of the passed UTXOs onto target addresses. But why do you need a address->keypair map to create the new transactions?
In the caller function you actually know which peer controls which address when you create the outputs in the first place. Can this function receive that peer instead so you avoid to pass keypairs all the way through?

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: I just find `make` more go-like: not really saying this because of malloc optimizations :smile:

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
karimodm: `make` is just more in line with go style in my opinion.
And regarding the pointers: you don't really modify the Address struct itself later on no? So removing the pointer should work.

tools/integration-tests/tester/tests/testutil.go
GalRogozinski: It is not this simple unfortunately.

This is because in order to derive the correct key from an address I need it's index.
So I will need to pass down the indexes of each address. That will also be ugly.
I think it is better to pass down the map. Maybe give it another name?
I can also do that `OutputWallet` idea that will at least force to have a key associated with the output.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: 1. done
2. The problem is that `determineTargets` needs it to be passed down. Maybe I can give it a different name in the signature that is more meaningful like `startingPoint`. As for logs, I can take them out. But I will still have to pass the variable down, so not sure it will have the effect you wanted.

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: Why is it more go style in your opinion? I am actually curious because never saw anything in the docs about it.

Anyhow, I don't mind using only `make` for empty slice allocation, so I am changing it. I just wish we had a style guide so I wouldn't have to guess this stuff. Back in the IRI days I just shamelessly copied the style guide from another project that I liked.
Maybe we can do the same?
I can also learn the style via comments though if it is too much :-P

tools/integration-tests/tester/tests/consensus/consensus_conflict_spam_test.go
GalRogozinski: Really want to know your thoughts on that.

In the meanwhile I changed the addresses to be non-pointers.
By the way I tried to do the same for transactions but I got warning since they have mutexes.

karimodm: In go, all types are passed via value (int, bool, float and so on); variable of a certain type though, do not contain the entire value itself, but just a reference to it: making them effectively always pointers: this is the case for slices and maps. If you pass a slice to a function and you modify an element of it, the change would be reflected in the caller too. You can then consider that all "complex" types are passed implicitly by pointers. The only exception to this are struct themselves: if you pass a struct as a parameter, a full copy of it will be passed (non-recursively though, e.g. complex types within a struct will point to the same memory location). Hence why golang allows you to declare methods of structs using the base struct as is or using a pointer notation: the latter is used if the fields of the calling struct will be modified by the method itself as part of the operation.

In your case, declaring the inner elements of the slice as pointers doesn't really change anything: unless what you want is to change something on the inner elements themselves!

I hope I didn't make this explanation too confusing!

karimodm (approved):

karimodm:
Let's merge the new reference types first before merging this and making sure it doesn't not randomly fail #1460

jonastheis (approved):

PR #1854, opened 2021-11-22
changes to spammer tool

apenzk:
# Description of change

updated the documentation to account for the new way of using the spammer

## Type of change

- Documentation Fix

Its basically just text.

## Change checklist


plugins/spammer/webapi.go
```diff
@@ -18,7 +18,9 @@ func handleRequest(c echo.Context) error {
 	switch request.Cmd {
 	case "start":
 		if request.Rate == 0 {
+			log.Infof("Requesting invalid spamming at rate 0 mps. Setting it to 1 mps")
 			request.Rate = 1
+			request.Unit = "mps"
```
daria305, line 23: Not sure if needed here, as later mps is set as default in switch anyway. Also it still can be changed to mpm  if user provided it as a parameter. In log we can just say rate is set to 1 per time unit

daria305 (approved):
Nice update, I think line 23 is not necessary but doesn't bother too

plugins/spammer/webapi.go
apenzk: ok will remove this.

karimodm (approved):

PR #1856, opened 2021-11-23
New Scheduler buffer management & data flow

### LINKED ISSUE #1796: Scheduler Buffer Management & Data Flow
Integrate the following data flow:
dataflow

- [ ] Remove timestamp check from parser 
- [ ] Scheduler
  - [ ] Discard buffered messages on shutdown: no persisting or waiting for them to be scheduled
  - [ ] Enqueue messages to buffer (`Scheduler.Submit(id)`) after booking.
  - [ ] Mark messages ready (`Scheduler.Ready(id)`), when their parents are scheduled or confirmed.
  - [ ] Implement drop-head buffer policy
- [ ] Remove the ordered (ordering happens using the ready-flag) #1794
- [ ] Gossip scheduled message. This should not block the scheduler. #1794
- [ ] Tip management:
  - [x] Add scheduled messages as tips and remove its parents
  - [ ] Remove parents of confirmed messages
  - [x] Periodically remove tips with old timestamp
- [ ] Implement tip fishing in URTS, i.e. TSC-criteria #1766 (postponed)

Resources:
https://iota.cafe/t/buffer-management-for-the-icca-scheduler/1226

piotrm50:
# Description of change
Resolves #1796 

## Type of change

Manual testing, unit tests.

## Change checklist


packages/tangle/schedulerutils/bufferqueue.go
```diff
@@ -73,21 +69,76 @@ func (b *BufferQueue) Submit(msg Element, rep float64) error {
...
+			nodeMana := accessManaRetriever(q.NodeID())
+			if nodeMana > 0.0 {
+				if scale := float64(q.Size()) / nodeMana; scale > maxScale {
+					maxScale = scale
+					maxNodeID = q.NodeID()
+				}
+			} else if q.Size() > 0 {
+				maxScale = math.Inf(1)
+				maxNodeID = q.NodeID()
+			}
```
Wollac, line 108: We can `break` here directly and quit the loop. No need to keep iterating if `maxScale` is already at infinity.

Wollac: In this case, it would even be more go-idiomatic to use guard clauses and do:
```
if nodeMana == 0.0 {
  maxScale = math.Inf(1)
  maxNodeID = q.NodeID()
  break
}

if scale := ...
```

packages/tangle/schedulerutils/bufferqueue.go
```diff
@@ -73,21 +69,76 @@ func (b *BufferQueue) Submit(msg Element, rep float64) error {
...
+		for _, v := range longestQueue.submitted {
+			if oldestMessage == nil || oldestMessage.IssuingTime().After((*v).IssuingTime()) {
+				oldestMessage = *v
+			}
+		}
+
+		// if the submitted message is older than the oldest ready message, then drop the submitted message
+		// otherwise drop the oldest ready message
+		readyQueueFront := longestQueue.Front()
+		if oldestMessage != nil && (readyQueueFront == nil || readyQueueFront != nil && oldestMessage.IssuingTime().Before(readyQueueFront.IssuingTime())) {
```
Wollac, line 127:
```suggestion
		if oldestMessage != nil && (readyQueueFront == nil || oldestMessage.IssuingTime().Before(readyQueueFront.IssuingTime())) {
```
second `readyQueueFront` is not needed

packages/tangle/schedulerutils/bufferqueue.go
```diff
@@ -73,21 +69,76 @@ func (b *BufferQueue) Submit(msg Element, rep float64) error {
...
+		var oldestMessage Element
+
+		for _, v := range longestQueue.submitted {
+			if oldestMessage == nil || oldestMessage.IssuingTime().After((*v).IssuingTime()) {
+				oldestMessage = *v
+			}
+		}
+
+		// if the submitted message is older than the oldest ready message, then drop the submitted message
+		// otherwise drop the oldest ready message
```
Wollac, line 125:
```suggestion
		// if the oldest not-ready message is older than the oldest ready message, drop the former otherwise the latter
```
Improved the comment

plugins/remotemetrics/message.go
```diff
@@ -1,11 +1,91 @@
 package remotemetrics
 
 import (
+	"time"
+
```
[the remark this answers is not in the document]
Wollac: Let's not forget the doggy here and fix that issue

packages/tangle/scheduler.go
```diff
@@ -240,31 +297,74 @@ func (s *Scheduler) Clear() {
...
-
-	err := s.buffer.Submit(message, mana)
-	if err != nil {
-		s.Events.MessageDiscarded.Trigger(message.ID())
-	}
-	if errors.Is(err, schedulerutils.ErrInboxExceeded) {
-		s.Events.NodeBlacklisted.Trigger(nodeID)
+	s.tangle.Storage.MessageMetadata(message.ID()).Consume(func(messageMetadata *MessageMetadata) {
+		// shortly before submitting we set the queued time
+		messageMetadata.SetQueuedTime(time.Now())
```
Wollac, line 357: Should here - and for `SetDiscardedTime` - `clock.SyncedTime()` be used? Or are there some (performance) reasons against this?

packages/tangle/scheduler.go
```diff
@@ -279,37 +379,43 @@ func (s *Scheduler) schedule() *Message {
...
+				panic("MessageID could not be parsed!")
+			}
+			if s.tangle.ConfirmationOracle.IsMessageConfirmed(msgID) && clock.Since(msg.IssuingTime()) > s.confirmedMsgThreshold {
+				// if a message is confirmed, and issued some time ago, don't schedule it and take the next one from the queue
+				// do we want to mark those messages somehow for debugging?
+				s.Events.MessageSkipped.Trigger(msgID)
+				s.buffer.PopFront()
+				msg = q.Front()
+			} else {
+				foundMsg = true
```
Wollac, line 410: Just `break` instead of using `foundMsg`. Same effect but much cleaner

packages/tangle/schedulerutils/bufferqueue.go
piotrm50: We need to exit the loop after iterating through all elements, so that the internal state of `*BufferQueue` remains unchanged. It contains the `b.Current()` element, and I didn't want that to be changed after call of this function.

packages/tangle/scheduler.go
piotrm50: No, that's an error. Good catch!

Wollac (approved):
lgtm

acha-bill (approved):

PR #1860, opened 2021-11-24
Feature network GitHub action

piotrm50:
# Description of change

Add GitHub Action definition to deploy Goshimmer from any branch to feature network.

## Type of change

## Change checklist


.github/workflows/feature-network-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v1.9.0
+        with:
+          username: '${{ secrets.IOTALEDGER_HUB_DOCKER_LOGIN }}'
+          password: '${{ secrets.IOTALEDGER_HUB_DOCKER_PASSWORD }}'
+
+      - name: Publish to Docker Hub
+        uses: docker/build-push-action@v2.5.0
+        with:
+          tags: iotaledger/goshimmer:feature
```
karimodm, line 40: we should push to iotaledgertesting and change docker-compose files accordingly!

.github/workflows/feature-network-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1.9.0
+        with:
+          username: '${{ secrets.IOTALEDGER_HUB_DOCKER_LOGIN }}'
+          password: '${{ secrets.IOTALEDGER_HUB_DOCKER_PASSWORD }}'
```
karimodm, line 35: These secrets are called IOTALEDGERTESTING_HUB_DOCKER_LOGIN and IOTALEDGERTESTING_HUB_DOCKER_PASSWORD respectively.

karimodm (approved):
Everything looks great!! Thanks man!
Only thing is that the network does not seem to sync after your successful deploy: can you please check and fix before we merge?

deploy/ansible/hosts/feature.yml
```diff
@@ -35,7 +35,7 @@ goshimmers:
 
   vars:
     analysisServerHost: "analysisentry-01.feature.shimmer.iota.cafe"
-    entryNodeHost: "analysisentry-01.feature.shimmer.iota.cafe"
+    entryNodeHost: "bootstrap-01.feature.shimmer.iota.cafe"
```
georgysavva: Why is it so now?

.github/workflows/feature-network-deploy.yml
```diff
@@ -0,0 +1,84 @@
...
+  workflow_dispatch:
+    inputs:
+      snapshotUrl:
+        description: 'Custom snapshot URL:'
+        required: false
+        default: ""
+      useDefaultSnapshot:
+        description: 'Use default snapshot (0/1):'
+        required: true
+        default: "1"
```
georgysavva, line 12: Why do we need a separate input for that? If the `snapshotUrl` is an empty string "" it means we should use the default one, no?

deploy/ansible/feature-deploy.yml
```diff
@@ -52,7 +45,3 @@
         - { drandNumber: 3 }
         - { drandNumber: 4 }
         - { drandNumber: 5 }
-
-- hosts: asset_registries
-  roles:
-    - asset-registry
```
georgysavva, line 58: Why do we need a separate playbook file for feature deployment? We can reuse the existing `deploy.yml` as is. It doesn't contain the `asset_registries` part and the `goshimmer:&public_nodes` will be empty from the inventory file as it's for `pre-devnet`.

deploy/ansible/hosts/feature.yml
piotrm50: It's not like that. I was just testing to see why it doesn't do initial peering.

deploy/ansible/feature-deploy.yml
piotrm50: For Feature network we also deploy DRNG instances everytime we deploy the network, while in devnet those are created by initial setup and not touched during subsequent deployments. Maybe it could be simplified.

piotrm50: Maybe @karimodm can tell you more about it as he created this file.

.github/workflows/feature-network-deploy.yml
piotrm50: Done

deploy/ansible/feature-deploy.yml
georgysavva: @karimodm If we want drands instances to be deployed at the same time as Goshimmer nodes other environments as `pre-devnet` and `devnet` would benefit from it too. So we add it to the original `deploy.yml` file and we again don't need a separate `feature-deploy.yml`

georgysavva: The thing that was added to `feature-deploy.yml` that is not in the original `deploy.yml` is drands.
I agree with you that we don't need to deploy them every time for all our environments including the feature network.
So we can remove the recently added drands part from the `feature-deploy.yml`. That makes it identical with the original `deploy.yml` and we can get rid of the `feature-deploy.yml` at all.

PR #1865, opened 2021-11-24
Improve connection handling

georgysavva:
- Send a special packet right after Dial from the outbound peer, read and discard that packet from the other, inbound peer. This is needed to trigger the libp2p lazy negotiation process and establish the connection on both sides.
- Close and detach from libp2p host on Gossip shutdown
- Disconnect from a neighbour only upon Already Closed Error OR io.EOF
- Update to the new hive.go with the reset `rejectionFilter` in the auto peering


capossele (approved):
Hero of the day!

karimodm (approved):

PR #1868, opened 2021-11-26
Fix race condition in packetsStream with mutexes

georgysavva:
We write and read to/from the libp2p connection concurrently. Every write and ready operation consists of two sub write and read operations. The first one is for the uvarint and the second one for the data itself. Without proper locking, we have a race condition there.


capossele (approved):
Good job!

Issue #1803, opened 2021-11-12
Branch DAG: show AW, GoF and like status 


karimodm:
AW & liked branch should appear on the graph for any branch DAG, updated in real-time.

PR #1898, opened 2021-12-03
Add GoF and AW to branch info

acha-bill:
# Description of change

Add GoF and AW to branch info

Screenshot 2021-12-03 at 12 22 40

## Type of change

#1803

### DIFF
--- a/plugins/dagsvisualizer/frontend/src/index.tsx
+++ b/plugins/dagsvisualizer/frontend/src/index.tsx
@@ -5,6 +5,7 @@ import UTXOStore from 'stores/UTXOStore';
 import BranchStore from 'stores/BranchStore';
 import GlobalStore from 'stores/GlobalStore';
 import {Root} from 'components/Root';
+import React from 'react';
 
 const tangleStore = new TangleStore();
 const utxoStore = new UTXOStore();
--- a/plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
+++ b/plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
@@ -34,8 +34,9 @@ export class branchConfirmed {
 }
 
 export class branchWeightChanged {
-    ID: string;
+    ID:     string;
     weight: number;
+    gof:    string;
 }
 
 export class BranchStore {
@@ -128,6 +129,7 @@ export class BranchStore {
             return;
         }
         b.aw = branch.weight;
+        b.gof = branch.gof;
         this.branches.set(branch.ID, b)
     }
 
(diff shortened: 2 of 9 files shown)


plugins/dagsvisualizer/visualizer.go
```diff
@@ -159,9 +160,20 @@ func registerBranchEvents() {
 		})
 	})
 
+	branchWeightChangedClosure := events.NewClosure(func(e *tangle.BranchWeightChangedEvent) {
+		visualizerWorkerPool.TrySubmit(&wsMessage{
+			Type: MsgTypeBranchWeightChanged,
+			Data: &branchWeightChanged{
+				ID:     e.BranchID.Base58(),
+				Weight: e.Weight,
```
jkrvivian, line 168: When weight changes, GoF is changed, too.
Let's grab the new GoF and include it here.

plugins/dagsvisualizer/frontend/src/index.tsx
```diff
@@ -5,6 +5,7 @@ import UTXOStore from 'stores/UTXOStore';
 import BranchStore from 'stores/BranchStore';
 import GlobalStore from 'stores/GlobalStore';
 import {Root} from 'components/Root';
+import React from 'react';
```
jkrvivian, line 8: This package is not used.

plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
```diff
@@ -1,16 +1,18 @@
 import { action, makeObservable, observable, ObservableMap } from 'mobx';
-import { registerHandler, unregisterHandler, WSMsgType } from 'WS';
+import { registerHandler, unregisterHandler, WSMsgType } from '../WS';
```
jkrvivian, line 2: The baseUrl is set to `src` in tsconfig.json, `WS` should work. :thinking:

plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
```diff
@@ -85,7 +93,7 @@ export class BranchStore {
 
         if (this.paused) {
             this.branchToAddAfterResume.push(branch.ID);
-        } 
+        }
```
jkrvivian, line 96: Thanks for cleaning up extra spaces. :heart:

plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
acha-bill: For some reason `ws` was conflicting with some node_modules stuff. I think we should leave it as `../ws` to be safe.

plugins/dagsvisualizer/frontend/src/index.tsx
acha-bill: It's to fix TS error: `TS2686: 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`. 
Apparently some packages needs it.

jkrvivian (approved):

PR #1899, opened 2021-12-03
Improve volumes in docker-compose files

piotrm50:
# Description of change

Add missing prometheus.yml file for running DevNet node in Docker.

Improve volumes in docker-compose files for both local Docker network as well as DevNet node. 
Use Docker volumes instead of mapping path inside of repository. Makes it easier to clean volumes if necessary and allows users to more precisely configure them.

## Type of change

Ran both docker-compose files locally. 

## Change checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,8 +15,8 @@ services:
     # make sure to give read/write access to the folder ./mainnetdb (e.g., chmod -R 777 ./mainnetdb)
     # optionally, you can mount a config.json into the container
     volumes:
-      - "./db:/tmp/mainnetdb:rw"
-      - "./peerdb:/tmp/peerdb:rw"
+      - "goshimmer-db-volume:/tmp/mainnetdb:rw"
+      - "goshimmer-peerdb-volume:/tmp/peerdb:rw"
       - "/etc/localtime:/etc/localtime:ro"
     ports:
       # AutoPeering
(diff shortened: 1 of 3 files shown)


NicolasVautrin:
impossible to use azure container instance with actual setup.
Why not use an environment variable to set the timezone ?
environment:
- TZ= 'America/New_York'

root@goshimmer:/usr/local/bin# ./docker compose up -d
WARNING network outside: network.external.name is deprecated in favor of network.name 
host path ("/etc/localtime") not allowed as volume source, you need to reference an Azure File Share defined in the 'volumes' section

NicolasVautrin:
Here some things that prevent to deploy goShimmer on azure container instance.

root@goshimmer:/usr/local/bin# ./docker compose up 
host path ("/etc/localtime") not allowed as volume source, you need to reference an Azure File Share defined in the 'volumes' section 

Why not use environment variablew to set the timezone like TZ='france/paris' 

But the more complicated is this 
https://docs.docker.com/cloud/aci-compose-features/
Azure override the wall entrypoint when we use command in docker-compose. 

I’d tried to set the whole entry point of dockerfile but /run/dlv didn’t work ,exec is not recognize, only /run/goshimmer worked. But has you imagine my container don’t keep running without 'exec --continu'

NicolasVautrin:
Apache has something like that , to keep the container running
CMD ["apachectl", "-D",  "FOREGROUND"]

piotrm50:
> Why not use environment variablew to set the timezone like TZ='france/paris'
We want to have a generic solution that can be used out-of-the-box without customizing it beforehand. If you need, you can change it for your deployment to use environment variable :) 

 
> But the more complicated is this https://docs.docker.com/cloud/aci-compose-features/ Azure override the whole entrypoint when we use command in docker-compose.

The goal of this docker-compose is not to be one-fits-all solution, but rather a starting point to run the node in the basic environment and have something to start with if you want to run it in more specific environment.

karimodm (approved):

PR #1905, opened 2021-12-07
Scheduler skips confirmed messages

### LINKED ISSUE #1864: Remove confirmed messages from the scheduler
If a message is submitted but not yet ready, remove it immediately as those messages are stored in a map, which can be quickly accessed.

For confirmed messages that are marked as Ready, the scheduler it should skip it and take the next one. This approach would require asking the Confirmation oracle if a message has been confirmed, which can be costly. 

Another way of doing it is by removing it immediately, which would require searching the heap.

piotrm50:
# Description of change

Resolves #1864 

Confirmed messages that are older than 1 minutet are:
* removed from the scheduler upon confirmation if they are not ready yet
* skipped by the scheduler when it's their turn when they're ready. 

Confirmed messages that are younger than 1 minute (e.g. a message that confirms itself), will be scheduled because it's younger than 1 minute. 

## Type of change

Manual tests, syncing with the DevNet.

## Change checklist


packages/tangle/scheduler.go
```diff
@@ -132,6 +131,15 @@ func (s *Scheduler) Setup() {
 		if scheduled {
 			return
 		}
+		s.tangle.Storage.Message(messageID).Consume(func(message *Message) {
+			if clock.Since(message.IssuingTime()) > time.Minute {
```
capossele, line 135: Maybe better to use some parameter to define the time threshold?

packages/tangle/scheduler.go
```diff
@@ -178,6 +186,39 @@ func (s *Scheduler) NodeQueueSizes() map[identity.ID]int {
 	return nodeQueueSizes
 }
 
+// MaxBufferSize returns the max size of the buffer.
+func (s *Scheduler) MaxBufferSize() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
capossele, line 192: Is there a specific reason why we use a lock rather than a RW lock?

packages/tangle/scheduler.go
```diff
@@ -331,14 +372,29 @@ func (s *Scheduler) schedule() *Message {
...
-				rounds = r
-				schedulingNode = q
+		// while loop to skip all the confirmed messages
+		foundMsg := false
+		for msg != nil && !clock.SyncedTime().Before(msg.IssuingTime()) && !foundMsg {
+			msgID, _, err := MessageIDFromBytes(msg.IDBytes())
+			if err != nil {
+				panic("MessageID could not be parsed!")
+			}
+			if s.tangle.ConfirmationOracle.IsMessageConfirmed(msgID) && clock.Since(msg.IssuingTime()) > time.Minute {
```
capossele, line 382: same here, better to use a parameter for the time threshold

packages/tangle/scheduler.go
```diff
@@ -331,14 +372,29 @@ func (s *Scheduler) schedule() *Message {
...
+				// if a message is confirmed, and issued some time ago, don't schedule it and take the next one from the queue
+				// do we want to mark those messages somehow for debugging?
+				s.Events.MessageSkipped.Trigger(msgID)
```
capossele, line 385: This can be very dangerous as this loop must be as fast as possible and by triggering events we could risk that the consumer of the event  is not going to be fast

capossele: But I get that it's only done for debugging purposes

packages/tangle/scheduler.go
piotrm50: Sure thing!

packages/tangle/scheduler.go
piotrm50: Sure thing! Will replace it.

packages/tangle/scheduler.go
piotrm50: That's true. However I want to have a metric with number of skipped messages in Grafana, and the handler of that event is just incrementing a number. Of course we need to remember in the future not to add any complex handlers there.

piotrm50: I was more worried about the confirmation check, because the messages can be also removed upon confirmation, but doing it this way seemed easier to start with. Given that usually the queues should be relatively small, it might make sense to remove them upon confirmation. What do you think?

packages/tangle/scheduler.go
piotrm50: That's a good question. I just used whatever was already in place, but we could change it to RW lock. I'll do it :)

packages/tangle/booker.go
```diff
@@ -549,28 +558,43 @@ func messageBranchUpdatedCaller(handler interface{}, params ...interface{}) {
...
+		Manager:        markers.NewManager(tangle.Options.Store, tangle.Options.CacheTimeProvider),
 	}
 }
 
 // InheritStructureDetails returns the structure Details of a Message that are derived from the StructureDetails of its
 // strong and like parents.
 func (m *MarkersManager) InheritStructureDetails(message *Message, sequenceAlias markers.SequenceAlias) (structureDetails *markers.StructureDetails) {
-	structureDetails, _ = m.Manager.InheritStructureDetails(m.structureDetailsOfStrongAndLikeParents(message), m.tangle.Options.IncreaseMarkersIndexCallback, sequenceAlias)
+	// customize the function, so that markers are not assigned to nodes whose messages might be dropped
+	// this is a temporary solution, that's why the function signature is not change to additionaly accept messageID
```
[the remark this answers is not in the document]
piotrm50:
```suggestion
	// this is a temporary solution, that's why the function signature is not change to additionally accept messageID
```

packages/tangle/booker.go
```diff
@@ -549,28 +558,43 @@ func messageBranchUpdatedCaller(handler interface{}, params ...interface{}) {
...
+	structureDetails, _ = m.Manager.InheritStructureDetails(m.structureDetailsOfStrongAndLikeParents(message), func(sequenceID markers.SequenceID, currentHighestIndex markers.Index) bool {
+		nodeID := identity.NewID(message.IssuerPublicKey())
+		bufferUsedRatio := float64(m.tangle.Scheduler.BufferSize()) / float64(m.tangle.Scheduler.MaxBufferSize())
+		nodeQueueRatio := float64(m.tangle.Scheduler.NodeQueueSize(nodeID)) / float64(m.tangle.Scheduler.BufferSize())
+		if bufferUsedRatio > 0.01 && nodeQueueRatio > 0.1 {
+			return false
+		}
+		if discardTime, ok := m.discardedNodes[nodeID]; ok && time.Since(discardTime) < time.Minute {
+			return false
+		} else if ok && time.Since(discardTime) >= time.Minute {
```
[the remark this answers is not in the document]
piotrm50: This method tells whether a new marker should be assigned to the message. What is done here, is we check whether any message from the same issuer has been discarded within the last minute. If there was such a message, then we don't want to use this message as a marker either, because there is a good chance that it might be problematic as well.

GalRogozinski (approved):

PR #1959, opened 2022-01-14
Skip snapshot download when rebuilding images of old goshimmer version

piotrm50:
# Description of change

We don't have the valid snapshot for old versions anymore, so downloading the latest one is incorrect. 

## Type of change

## Change checklist

### DIFF
--- a/.github/workflows/rebuild-docker-image.yml
+++ b/.github/workflows/rebuild-docker-image.yml
@@ -26,4 +26,4 @@ jobs:
           username: '${{ secrets.IOTALEDGER_HUB_DOCKER_LOGIN }}'
           password: '${{ secrets.IOTALEDGER_HUB_DOCKER_PASSWORD }}'
           tags: latest,${{github.event.inputs.tagName}}
-          tag_with_ref: true
+          build-args: DOWNLOAD_SNAPSHOT=0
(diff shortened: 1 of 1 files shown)


.github/workflows/rebuild-docker-image.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           username: '${{ secrets.IOTALEDGER_HUB_DOCKER_LOGIN }}'
           password: '${{ secrets.IOTALEDGER_HUB_DOCKER_PASSWORD }}'
           tags: latest,${{github.event.inputs.tagName}}
```
karimodm, line 28: It shouldn't be tagged as latest if we are rebuilding an older image.

.github/workflows/rebuild-docker-image.yml
```diff
@@ -26,4 +26,4 @@ jobs:
...
-          tag_with_ref: true
\ No newline at end of file
+          build-args: DOWNLOAD_SNAPSHOT=0
```
karimodm, line 29: This implies that one will have to manually mount a snapshot as a volume, because none can be found in the image itself: sounds good.

karimodm (changes requested):

.github/workflows/rebuild-docker-image.yml
piotrm50: Yes. This workflow will be mostly used to rebuild the images from 0.7.5 to 0.8.3 that were not properly tagged during their release. So I think it makes sense that they should not include current snapshot.

.github/workflows/rebuild-docker-image.yml
piotrm50:
```suggestion
          tags: ${{github.event.inputs.tagName}}
```

karimodm (approved):

Issue #1893, opened 2021-12-01
Visualizer: Fix vertex limit


jkrvivian:
- [x] Adjust vertices amount when the limit changes
     - [x] when the vertex limit is lower than the default value. Currently, it will just remove vertices one by one when a new vertex joins.
- [x] Ensure removed elements are actually removed and not taking spaces from memory.

PR #1964, opened 2022-01-17
DAGs Visualizer: Fix vertices limit

daria305:
Solves #1893 

## Type of change
Adjusts vertices amount when the limit changes
Ensures removed elements are actually removed and not taking spaces from memory - added release node for svg graphics

### DIFF
--- a/plugins/dagsvisualizer/frontend/src/stores/TangleStore.tsx
+++ b/plugins/dagsvisualizer/frontend/src/stores/TangleStore.tsx
@@ -1,12 +1,7 @@
-import { action, makeObservable, observable, ObservableMap } from 'mobx';
-import {
-    connectWebSocket,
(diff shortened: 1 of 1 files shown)


plugins/dagsvisualizer/frontend/src/stores/TangleStore.tsx
```diff
@@ -61,7 +56,7 @@ export class TangleStore {
     msgOrder: Array<string> = [];
     lastMsgAddedBeforePause = '';
     selected_origin_color = '';
-    highligtedMsgs = new Map<string, string>();
+    highlightedMsgs = new Map<string, string>();
```
jkrvivian, line 59: Nice catch!! :heart:

jkrvivian (approved):

PR #1965, opened 2022-01-17
Add recieved time remote metric

### LINKED ISSUE #1885: Implement additional remotelogger metrics
Implement monitoring of new metrics related to congestion control, that would allow us to gain insight about the new congestion control. The collected metrics should allow easy analysis in the ELK stash.

List of metrics: 
https://docs.google.com/document/d/1BVNgRtSGbBrKeGwFgj_lbyN1PoTr_Ddd2GZ7l6J0nO8/edit#

GalRogozinski:
# Description of change

Adds received time remote metrics which was missing when we closed #1885

## Type of change

Crossed fingers

### DIFF
--- a/packages/remotemetrics/events.go
+++ b/packages/remotemetrics/events.go
@@ -82,12 +82,14 @@ type MessageScheduledMetrics struct {
 	GradeOfFinality          uint8     `json:"gradeOfFinality" bson:"GradeOfFinality"`
 	DeltaGradeOfFinalityTime int64     `json:"deltaGradeOfFinalityTime" bson:"deltaGradeOfFinalityTime"`
 	DeltaSolid               int64     `json:"deltaSolid,omitempty" bson:"deltaSolid"`
-	// scheduledTimestamp - IssuedTimestamp in nanoseconds
+	// ScheduledTimestamp - IssuedTimestamp in nanoseconds
 	DeltaScheduledIssued int64 `json:"deltaScheduledIssued" bson:"deltaScheduledIssued"`
 	DeltaBooked          int64 `json:"deltaBooked" bson:"deltaBooked"`
-	// scheduledTimestamp - ReceivedTimestamp in nanoseconds
+	// ScheduledTimestamp - ReceivedTimestamp in nanoseconds
 	DeltaScheduledReceived int64 `json:"deltaScheduledReceived" bson:"deltaScheduledReceived"`
-	// scheduledTimestamp - QueuedTimestamp in nanoseconds
+	// ReceivedTimestamp - IssuedTimestamp in nanoseconds
+	DeltaReceivedIssued int64 `json:"DeltaReceivedIssued" bson:"DeltaReceivedIssued"`
+	// ScheduledTimestamp - QueuedTimestamp in nanoseconds
 	SchedulingTime  int64   `json:"schedulingTime" bson:"schedulingTime"`
 	AccessMana      float64 `json:"accessMana" bson:"accessMana"`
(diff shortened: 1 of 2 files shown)


packages/remotemetrics/events.go
```diff
@@ -70,6 +70,7 @@ type MessageScheduledMetrics struct {
 	MessageID                string    `json:"messageID" bson:"messageID"`
 	TransactionID            string    `json:"transactionID,omitempty" bson:"transactionID"`
 	IssuedTimestamp          time.Time `json:"issuedTimestamp" bson:"issuedTimestamp"`
+	ReceivedTimestamp        time.Time `json:"receivedTimestamp" bson:"receivedTimestamp"`
```
piotrm50, line 73: What about difference between received and issued timestamp? We cannot easily calculate it in Elasticsearch, so this would need to be added if we want to measure the delay?

GalRogozinski: I can add comments if you want

Issued time - encoded into the message. The time set by the node that created it.
ReceivedTimestamp - the time that your node received a meesage

your question made me remember that there is a little bug that should be fixed
So I'll commit something in a second

GalRogozinski: I think processing timestamp and delta shceduled have confusing names currently

you have recommendations?

packages/remotemetrics/events.go
```diff
@@ -79,8 +82,9 @@ type MessageScheduledMetrics struct {
 	GradeOfFinality          uint8     `json:"gradeOfFinality" bson:"GradeOfFinality"`
 	DeltaGradeOfFinalityTime int64     `json:"deltaGradeOfFinalityTime" bson:"deltaGradeOfFinalityTime"`
 	DeltaSolid               int64     `json:"deltaSolid,omitempty" bson:"deltaSolid"`
-	DeltaScheduled           int64     `json:"deltaScheduled" bson:"deltaScheduled"`
-	DeltaBooked              int64     `json:"deltaBooked" bson:"deltaBooked"`
+	// scheduledTimestamp - IssuedTimestamp in nanoseconds
+	DeltaScheduled int64 `json:"deltaScheduled" bson:"deltaScheduled"`
+	DeltaBooked    int64 `json:"deltaBooked" bson:"deltaBooked"`
 	// scheduledTimestamp - ReceivedTimestamp in nanoseconds
 	ProcessingTime int64 `json:"processingTime" bson:"processingTime"`
```
GalRogozinski, line 89: I think `processingTime` and `deltaScheduled` have confusing names currently

you have recommendations?

packages/remotemetrics/events.go
piotrm50: I meant that we could have another field here, that would be the difference `ReceivedTimestamp - IssuedTimestamp`.

GalRogozinski: we already have it `ProcessingTime`

packages/remotemetrics/events.go
piotrm50: `DeltaIssuedScheduled`
`DeltaIssuedBooked`
etc...

`DeltaReceivedScheduled` - that'd be the processing time
`DeltaReceivedBooked`
etc.

that way we could encode what exactly a given field means and it would not be conflicting for deltas from issuance time and receive time.

packages/remotemetrics/events.go
GalRogozinski: I think the names are bad, see my other comment
but not sure what the new names should be @piotrm50

piotrm50 (approved):

PR #1990, opened 2022-01-26
Spam protection 

georgysavva:
# Description of change

This PR introduces a new module for spam protection.

- I am going to introduce a general entity: PeerRateLimiter, its responsibility would be to count activity and notify in case the limit is hit. We inject PeerRateLimiter into various places across Goshimmer, currently into gossip (counts messages) and tangle (counts some activity related to rejected branches).
- Another new entity would be FirewallManager it responsible for taking action in case a peer is faulty. So basically it exposes a single method OnFaultyPeer(p *peer.Peer, reason string). When this method is called it checks whether the peer is auto-peered or manual-peered in the first case it calls autopeering Blocklist API, in the second case it just logs a warning.
- We attach OnFaultyPeer to all rate limiters notification events. So for all rate limiters, when the limit is hit the OnFaultyPeer gets triggered.

This is the simplest version. I think it should be enough for now because the only action we want to do is Blocklist for autopeering and log for manual peering regardless of the rate limiter reported the peer. But it could be easily extended in the future if we have the need.

## Type of change

Blocked by: https://github.com/iotaledger/hive.go/pull/310


packages/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,110 @@
...
+	hitEvent     *events.Event
+	peersRecords *ttlcache.Cache
+	log          *logger.Logger
+}
+
+// NewPeerRateLimiter returns a new instance of the PeerRateLimiter object.
+func NewPeerRateLimiter(interval time.Duration, limit int, log *logger.Logger) (*PeerRateLimiter, error) {
+	records := ttlcache.NewCache()
+	records.SetLoaderFunction(func(_ string) (interface{}, time.Duration, error) {
+		record := &limiterRecord{counter: ratecounter.NewRateCounter(interval), limitHitReported: atomic.NewBool(false)}
```
jonastheis, line 41: Isn't it a bit of an overkill to use `ttlcache` and `ratecounter`? If I understand correctly, the cache will expire every second anyway and the record recreated (if configured for 1 second interval). Now why can't we simply count up the entry for each peer in the cache and if it hits the limit for we trigger the event. Then there's no need for `ratecounter` and `limitHitReported` anymore. Also, `ttlcache` should be thread-safe and it can simply be a mapping to an int64 to count no?

I think this should work because we specify the limit per interval when we configure. Overall, this will make it a bit leaner and easier to understand imo.

plugins/gossip/parameters.go
```diff
@@ -11,6 +13,14 @@ type ParametersDefinition struct {
 
 	// MissingMessageRequestRelayProbability defines the probability of missing message requests being relayed to other neighbors.
 	MissingMessageRequestRelayProbability float64 `default:"0.01" usage:"the probability of missing message requests being relayed to other neighbors"`
+
+	MessagesRateLimit messagesLimitParameters
+}
+
+type messagesLimitParameters struct {
```
jonastheis, line 20: I'd expect these parameter to be set in the firewall plugin and be named`firewall.rateLimiter.<parameter>` or something like this.

plugins/gossip/parameters.go
```diff
@@ -11,6 +13,14 @@ type ParametersDefinition struct {
...
+	Interval        time.Duration `default:"1s" usage:"the time interval for which we count the messages rate"`
+	Limit           int           `default:"200" usage:"the default limit of messages per interval"`
+	DuringSyncLimit int           `default:"5000" usage:"the limit of messages per interval during node sync"`
```
jonastheis, line 23: This is highly dependent on the bandwidth and power of nodes and I think very hard to specify meaningfully. How about instead of having another parameter simply counting down whenever we send a message request to a neighbor?

plugins/gossip/gossip.go
```diff
@@ -60,15 +62,42 @@ func createManager(lPeer *peer.Local, t *tangle.Tangle) *gossip.Manager {
...
+		setLimitClosure := events.NewClosure(func(ev *tangle.SyncChangedEvent) {
+			if ev.Synced {
+				mrl.SetLimit(Parameters.MessagesRateLimit.Limit)
+			} else {
+				mrl.SetLimit(Parameters.MessagesRateLimit.DuringSyncLimit)
+			}
+		})
+		deps.Tangle.TimeManager.Events.SyncChanged.Attach(setLimitClosure)
+		defer deps.Tangle.TimeManager.Events.SyncChanged.Detach(setLimitClosure)
+		mrl.Close()
```
jonastheis, line 99: Isn't this always called when `mrl` is set? Maybe a missing `defer`?

plugins/gossip/gossip.go
```diff
@@ -60,15 +62,42 @@ func createManager(lPeer *peer.Local, t *tangle.Tangle) *gossip.Manager {
...
+	if mrl := deps.GossipMgr.MessagesRateLimiter(); mrl != nil {
+		setLimitClosure := events.NewClosure(func(ev *tangle.SyncChangedEvent) {
+			if ev.Synced {
+				mrl.SetLimit(Parameters.MessagesRateLimit.Limit)
+			} else {
+				mrl.SetLimit(Parameters.MessagesRateLimit.DuringSyncLimit)
+			}
+		})
+		deps.Tangle.TimeManager.Events.SyncChanged.Attach(setLimitClosure)
+		defer deps.Tangle.TimeManager.Events.SyncChanged.Detach(setLimitClosure)
```
jonastheis, line 98: If we get rid of the separate `DuringSyncLimit` this is not needed anymore. Maybe even the possibility to change the limit during runtime is not needed anymore. But since the latter it's coded, we could just leave it.

jonastheis (changes requested):
Looks good overall, maybe some simplification possible

tools/docker-network/docker-compose.yml
```diff
@@ -71,6 +71,33 @@ services:
...
+      restart_policy:
+        condition: on-failure
+        max_attempts: 5
+    secrets:
+      - goshimmer.config.json
+      - goshimmer.message.snapshot.bin
+    networks:
+      - shimmer
+    depends_on:
+      - peer_master
```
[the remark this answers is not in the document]
georgysavva: It was for debugging purpuses. I forgot to delete it. Thanks!

packages/gossip/manager.go
```diff
@@ -74,14 +75,20 @@ type Manager struct {
...
 	messageRequestWorkerPool *workerpool.NonBlockingQueuedWorkerPool
 }
 
+// ManagerOption configures the Manager instance.
+type ManagerOption func(m *Manager)
+
 // NewManager creates a new Manager.
-func NewManager(libp2pHost host.Host, local *peer.Local, f LoadMessageFunc, log *logger.Logger) *Manager {
+func NewManager(libp2pHost host.Host, local *peer.Local, f LoadMessageFunc, log *logger.Logger, opts ...ManagerOption,
+) (*Manager, error) {
```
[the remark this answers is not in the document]
georgysavva: Yeah, in the latest version there is no error. I'll delete it

packages/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Close closes PeerRateLimiter instance, it can't be used after that.
+func (prl *PeerRateLimiter) Close() {
+	if err := prl.peersRecords.Close(); err != nil {
+		prl.log.Errorw("Failed to close peers records cache", "err", err)
+	}
+}
+
+func (prl *PeerRateLimiter) doCount(p *peer.Peer) error {
+	prl.log.Debug("Count peer activity", "peerId", p.ID())
```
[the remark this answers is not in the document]
georgysavva: Agree.

packages/ratelimiter/rate_limiter.go
georgysavva: The cache doesn't expire if you touch it. So every time you count activity for the peer, its cache record TTL gets reset.
We could use an int counter for every peer and timestamp of the last reset. But it would be a different rate-limiting algorithm, less accurate.

plugins/gossip/parameters.go
georgysavva: I wanted to decouple firewall logic and rate limiters logic. We might have multiple different rate limiters with specific configuration settings that are injected in various goshimmer places (gossip, tangle, etc.). 
Other than that, not only a rate limiter is allowed to trigger the firewall, potentially it could be exposed via web API and triggered from the outside.
Even more, I tried to put the RateLimiter and Firewall in the single package in the begging and I got cycle dependency because Firewall depends on Gossip and Gossip depends on RateLimiter.

plugins/gossip/parameters.go
georgysavva: You mean another rate counter for sent messages requests?

jonastheis: No instead of having a separate limit while syncing I'd do the following per neighbor:
- count up if we receive a message
- count down if request a message (this should only happen during syncing/solidification) 

In that way we do not need to specify `DuringSyncLimit` a priori.

packages/ratelimiter/rate_limiter.go
jonastheis: https://github.com/ReneKroon/ttlcache says:
> Note (issue #25): by default, due to historic reasons, the TTL will be reset on each cache hit and you need to explicitly configure the cache to use a TTL that will not get extended.

`SkipTTLExtensionOnHit(true)` should expire within the limit even when touching it.

georgysavva: Yes, we can do this. What you are proposing is indeed simpler, but it would be a different rate-limiting algorithm, your version is called Fixed Window and the one using the rate counter is Sliding Window.
The fixed window is less accurate since it has a notion of hard reset of the timer, it's possible to generate twice more activity than Is set by the rate limiter. 
For example, we have a rate limit of 10 messages per second. With Fixed Window it's possible to spam 20 messages per second: 10 messages sent within 0.5 seconds before the time reset and 10 messages sent within 0.5 seconds after the time reset results in a total of 20 messages per second.

plugins/gossip/parameters.go
georgysavva: My original idea was almost the same as yours (I just thought to increment the limit instead of count down on messages request)
It seems a little bit fragile to me and might rate limit some healthy peers during the sync. The questions I had:
1. When we count down/increment the limit we basically give "credit" to a peer. Why do we give it for the rate counter "interval" amount of time (1s) Why not less or more? What if takes more time for a peer to send the request message back to us.
2. With the rate counting algorithm you proposed (Fixed Window), we lose the credit for the peer when we reset the timer. Is it okay still?

packages/ratelimiter/rate_limiter.go
jonastheis: Let's leave it with the sliding window approach but keep the other one in mind in case this is too inefficient.

plugins/gossip/parameters.go
jonastheis: 1. Good question, it's somewhat arbitrary I guess. We'd just need to find a good estimate on when the neighbor answers as to make sure that it does not cause rate limiting healthy peers. We can probably assume that a message request is answered within 1 second. RTT should be fairly fast between direct neighbors. 
2. Let's leave a sliding window approach for now.

So seeing the benefits of simpler concept, simplified code, and one less parameter to specify I'd just go for it (with 1 second for now). If we get false positives during syncing we can still change the policy or increase the limit for a longer period of time.

georgysavva: Okay, let's try and see. This is indeed a more elegant solution.

jonastheis (approved):
Looks good! 
My only concern would be that especially syncing and normal node operation are impacted performance wise. Because of locking overhead for every message that we send and request. But we can take a look at that later if it turns out to be a bottleneck.

PR #1997, opened 2022-02-02
Marker simplification (first iteration)

jonastheis:
# Description of change

This PR introduces a simplified version of markers without concept of `sequenceAlias` inside of the markers. Branches are still mapped onto markers (one way) utilizing the structural information that markers yield over the tangle. This simplifies concepts and code quite a bit and makes it less error prone if a branch is merged back to master. A new sequence is only created if the distance to the last past marker is above a certain threshold.

The little sketch shows how markers and sequences are created with a setting of `PastMarkerGap=3`

Markers Test drawio


jonastheis:
Awesome!

PR #1999, opened 2022-02-04
[dags] fix node.links is not iterable

acha-bill:
Fixes error when selecting an orphan message on the tangle DAG.


daria305 (approved):
good catch

PR #2037, opened 2022-02-11
DAGs Vis: color dags on confirmation and hide aggr branches

daria305:
# Description of change
Improvements for DAGs Visualizer:
 - different colors for confirmed branches and transactions
 - checkbox allowing to not render aggregated branches


plugins/dagsvisualizer/frontend/src/stores/BranchStore.tsx
```diff
@@ -105,6 +111,7 @@ export class BranchStore {
 
         b.isConfirmed = true;
         this.branches.set(confirmedBranch.ID, b);
+        updateConfirmedBranch(b, this.graph.cy);
```
jkrvivian, line 114: Pass  `this.graph` instead, and extract `cy` in `updateConfirmedBranch`.
We want to hide the graph library details in stores.

jkrvivian (approved):

PR #2042, opened 2022-02-14
Add issuerID to remote metrics

piotrm50:
# Description of change

Add issuerID to remote metrics

## Type of change

## Change checklist


plugins/remotelog/server/.env.default
```diff
@@ -1,3 +1,3 @@
-ELK_VERSION=7.5.2
+ELK_VERSION=7.13.2
```
GalRogozinski, line 1: There's already 8.0.0, no?

I usually recommend doing those changes in a seperate PR (assuming ansible creates a `.env` file from this template). This is because it is considered a bad practice to hide version upgradings inside PRs.

If someone tries to query elastic and get different behavior looking at `git log` won't tell him a thing...
Also if for some reason we understand that upgrading ELK is bad `git revert` won't work.

I guess in this project we less care about this stuff and I am not the repo maintainer, so I will simply approve and you can choose what you want to do with this comment :-)

GalRogozinski (approved):
Great!

I have an optional comment

plugins/remotelog/server/.env.default
piotrm50: This is only for local deployment only. The version deployed by ansible is `7.13.2` already so it's more of a cosmetic change.

PR #2045, opened 2022-02-14
clean test cache

GalRogozinski:
# Description of change

Sometimes when you don't run the tests using the bash scripts, clearing the cache is helpful.
When the tests run automatically, the command doesn't add any time overhead so I see no reason why not to include this.

## Type of change


jkrvivian (approved):

karimodm:
I am not sure what the `go clean` does specifically, but the `-count=1` argument to the test should already prevent the cache from being used no?

GalRogozinski:
@karimodm 

The internet says you are right.
But I know that the results were cached when I ran it and this is how I solved it

daria305 (approved):

PR #2053, opened 2022-02-17
Try to adjust rate limit better

georgysavva:
@piotrm50 noticed that node triggers rate limit during sync. 
It means that our logic about how we extend the rate limit is wrong. 
I try to extend the limit twice more hoping it could solve the problem and make rate limit more forgiving during sync.


karimodm:
Not ready for merge. Current debugging of Devnet might need to change some values here.

jonastheis (approved):

karimodm (approved):

PR #2069, opened 2022-02-22
Set DEBUG=1 to debug the integration tester

karimodm:
# Description of change

When running integration tests, if the `DEBUG` is set to 1, the `go test` is ran under a headless `dlv` listening on localhost on port `40000`.
It is possible to connect to the debugger via the IDE using the following VSCode launch.json:

```
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Connect to server",
			"type": "go",
			"request": "attach",
			"mode": "remote",
			"port": 50000,
			"host": "127.0.0.1"
		}
	]
}
```

### DIFF
--- a/tools/integration-tests/assets/entrypoint.sh
+++ b/tools/integration-tests/assets/entrypoint.sh
@@ -5,5 +5,12 @@ cp -rp /tmp/assets/* /assets
 chmod 777 /assets/*
 echo "assets:"
 ls /assets
-echo "running tests..."
-go test ./tests/"${TEST_NAME}" -v -timeout 30m -count=1
+if [ "x$DEBUG" = "x1" ]; then
+  echo "install debugging tools..."
+  go install github.com/go-delve/delve/cmd/dlv@master
+  echo "debugging tests..."
+  dlv test --headless --listen ':40000' --accept-multiclient --api-version 2 ./tests/${TEST_NAME}
+else
+  echo "running tests..."
+  go test ./tests/"${TEST_NAME}" -v -timeout 30m -count=1
+fi
--- a/tools/integration-tests/tester/docker-compose.yml
+++ b/tools/integration-tests/tester/docker-compose.yml
@@ -10,7 +10,7 @@ services:
       - DEBUG=${DEBUG}
       - TEST_NAME=${TEST_NAME}
     ports:
-      - 127.0.0.1:40000:40000
+      - "127.0.0.1:50000:40000"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
       - ../../..:/tmp/goshimmer:rw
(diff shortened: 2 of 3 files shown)


tools/integration-tests/assets/entrypoint.sh
```diff
@@ -5,5 +5,12 @@ cp -rp /tmp/assets/* /assets
 chmod 777 /assets/*
 echo "assets:"
 ls /assets
-echo "running tests..."
-go test ./tests/"${TEST_NAME}" -v -timeout 30m -count=1
+if [ "x$DEBUG" = "x1" ]; then
```
GalRogozinski, line 8: Just curious why are you using `x`?
I saw that `"$DEBUG" = "1"` works perfectly well!
I guess you wanted to make sure it is a string?

I don't care much for the extra character
I just must know you reason to have peace of mind :-P

tools/integration-tests/tester/docker-compose.yml
```diff
@@ -3,11 +3,14 @@ version: "3.5"
...
     container_name: tester
-    image: golang:1.16.2
+    image: golang:1.16
     working_dir: /tmp/goshimmer/tools/integration-tests/tester
     command: /tmp/assets/entrypoint.sh
     environment:
+      - DEBUG=${DEBUG}
       - TEST_NAME=${TEST_NAME}
+    ports:
+      - 127.0.0.1:40000:40000
```
GalRogozinski, line 13:
```suggestion
      - "127.0.0.1:50000:50000"
```
1. 40000 is the default node debug port. But we are not debugging the node here so better use some other random number.
2. My IDE shouted at me for not using quotes because of some edge cases. I don't like being shouted at.

This is the only comment that caused me to press "request changes"

tools/integration-tests/tester/docker-compose.yml
```diff
@@ -3,11 +3,14 @@ version: "3.5"
 services:
   tester:
     container_name: tester
-    image: golang:1.16.2
+    image: golang:1.16
```
GalRogozinski, line 6: 👍 

Just want to say that one day a bug may pop up from a silent update of a minor version of some shitty lib (probably not go), and it would be hard to track where it came from.

GalRogozinski (changes requested):
Works

just the port mapping thingie should be changed

tools/integration-tests/tester/docker-compose.yml
karimodm: Good point.

tools/integration-tests/assets/entrypoint.sh
karimodm: It is a bit common practice in GNU tools. Using "$DEBUG" also works.

GalRogozinski (approved):

PR #2070, opened 2022-02-22
Remove aggregated branches

jonastheis:
# Description of change

To simplify the concepts of the BranchDAG, we remove the AggregatedBranches from the code base and just carry `BranchIDs` in every object (lots of duplicate branches might be stored but due to merge to master it should be limited and go back to master branch quickly). Later on we can introduce a storage optimization aka `CompressedBranchIDs` to handle deduplication on the storage layer itself. In that way the interface of `Message`, `Transaction` and `Output` does not change regardless of how the branch IDs are actually stored. 

## Type of change


packages/tangle/approvalweightmanager.go
```diff
@@ -60,24 +59,10 @@ func (a *ApprovalWeightManager) processBookedMessage(messageID MessageID) {
...
-
-	weight = math.MaxFloat64
-	for conflictBranchID := range conflictBranchIDs {
-		a.tangle.Storage.BranchWeight(conflictBranchID).Consume(func(branchWeight *BranchWeight) {
-			if branchWeight.Weight() <= weight {
-				weight = branchWeight.Weight()
-			}
-		})
-	}
+	a.tangle.Storage.BranchWeight(branchID).Consume(func(branchWeight *BranchWeight) {
```
karimodm, line 62: I wrote this so I shouldn't review it.

packages/tangle/ledgerstate.go
```diff
@@ -64,7 +64,7 @@ func (l *LedgerState) TransactionValid(transaction *ledgerstate.Transaction, mes
 
 // TransactionConflicting returns whether the given transaction is part of a conflict.
 func (l *LedgerState) TransactionConflicting(transactionID ledgerstate.TransactionID) bool {
-	return l.BranchID(transactionID) == ledgerstate.NewBranchID(transactionID)
+	return l.BranchIDs(transactionID).Contains(ledgerstate.NewBranchID(transactionID))
```
karimodm, line 67:
```
// IsConflicting returns true if the Transaction is conflicting with another Transaction (has its own Branch).
func (t *TransactionMetadata) IsConflicting() bool {
	branchIDs := t.BranchIDs()
	return len(branchIDs) == 1 && branchIDs.Contains(NewBranchID(t.ID()))
}
```

should we do the same `len(branchIDs) == 1` here?

karimodm (changes requested):
Great work man, thanks!

karimodm (approved):

packages/tangle/message.go
```diff
@@ -1090,52 +1092,58 @@ func (m *MessageMetadata) StructureDetails() *markers.StructureDetails {
...
+func (m *MessageMetadata) SetAddedBranchIDs(addedBranchIDs ledgerstate.BranchIDs) (modified bool) {
 	m.addedBranchIDsMutex.Lock()
 	defer m.addedBranchIDsMutex.Unlock()
 
-	if m.addedBranchIDs == aggregatedAddedBranchIDs {
-		return
-	}
-
-	m.addedBranchIDs = aggregatedAddedBranchIDs
+	m.addedBranchIDs = addedBranchIDs.Clone()
```
hmoog, line 1100: again - check for equality and return modified false.

packages/tangle/message.go
```diff
@@ -1090,52 +1092,58 @@ func (m *MessageMetadata) StructureDetails() *markers.StructureDetails {
...
+func (m *MessageMetadata) SetSubtractedBranchIDs(subtractedBranchIDs ledgerstate.BranchIDs) (modified bool) {
 	m.subtractedBranchIDsMutex.Lock()
 	defer m.subtractedBranchIDsMutex.Unlock()
 
-	if m.subtractedBranchIDs == aggregatedSubtractedBranchIDs {
-		return
-	}
-
-	m.subtractedBranchIDs = aggregatedSubtractedBranchIDs
+	m.subtractedBranchIDs = subtractedBranchIDs.Clone()
```
hmoog, line 1134: check for equality and return false

packages/txstream/tangleledger/tangleledger.go
```diff
@@ -71,13 +71,18 @@ func (t *TangleLedger) GetUnspentOutputs(addr ledgerstate.Address, f func(output
...
-			found = true
-			t.tangleInstance.LedgerState.Transaction(txid).Consume(f)
+		for txBranchID := range txmeta.BranchIDs() {
+			if found = found && t.tangleInstance.ConfirmationOracle.IsBranchConfirmed(txBranchID); !found {
+				break
+			}
 		}
 	})
+
+	if found {
```
hmoog, line 83: shouldnt it only execute the callback if ALL branches have high gof AND the tx has high gof?

hmoog (changes requested):
I have left some comments

PR #2078, opened 2022-02-24
Tools API to send data message with custom parents

### LINKED ISSUE #2074: Tools API to post an entire message
## Description

We want to be able to control the selection of parents when constructing a message. This helps create more complicated scenarios.
Hence, message construction is delegated from the `MessageFactory` to the client.

- Create a `POST` endpoint `/tools/message` that accepts the message bytes.
- The handler parses the message and does some checks. e.g If the `nonce` is 0, perform PoW on the node and update the message.
- Then it passes the message to the normal gossip flow by triggering `MessageConstructedEvent`.
- The handler should wait for the message to be stored before returning success. Else, return an error.

acha-bill:
# Description of change

API to send a data message with custom parents.

## Type of change

Added integration test

## Change checklist

Fixes #2074

### DIFF
--- a/plugins/webapi/tools/message/send_message.go
+++ b/plugins/webapi/tools/message/send_message.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/iotaledger/goshimmer/packages/jsonmodels"
 	"github.com/iotaledger/goshimmer/packages/tangle"
+	"github.com/iotaledger/goshimmer/packages/tangle/payload"
 )
 
 const maxIssuedAwaitTime = 5 * time.Second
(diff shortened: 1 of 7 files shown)


plugins/webapi/tools/message/send_message.go
```diff
@@ -0,0 +1,96 @@
...
+	if len(request.Data) == 0 {
+		return c.JSON(http.StatusBadRequest, jsonmodels.DataResponse{Error: "no data provided"})
+	}
+
+	msg, _, err := tangle.MessageFromBytes(request.Data)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, jsonmodels.DataResponse{Error: fmt.Sprintf("error: %s", err.Error())})
+	}
+
+	if msg.Version() != tangle.MessageVersion {
```
daria305, line 41: Why should user even care if he puts a correct version in the API request if the message is prepared by node anyway?

tools/integration-tests/tester/tests/diagnostics/diagnostics_test.go
```diff
@@ -96,3 +102,42 @@ func TestDiagnosticApis(t *testing.T) {
...
+	api := node.GoShimmerAPI
+
+	sent := tests.SendDataMessages(t, n.Peers()[:1], 5)
+	parentMessageIDs := tangle.NewParentMessageIDs()
+
+	for k := range sent {
+		parentID, _ := tangle.NewMessageID(k)
+		parentMessageIDs = parentMessageIDs.Add(tangle.StrongParentType, parentID)
+	}
+	msg, err := tangle.NewMessage(
```
daria305, line 124: Message is prepared anyway by the node we send the request to, it is a little misleading that we need to create this dummy message and put the bytes. What about passing only parents and payload to api.SendMessage() and this wrapping and changing to bytes will be done by client lib

plugins/webapi/tools/message/send_message.go
```diff
@@ -0,0 +1,96 @@
...
+	var parents tangle.MessageIDsSlice
+	msg.ForEachParent(func(parent tangle.Parent) {
+		parents = append(parents, parent.ID)
+	})
+	issuingTime := deps.Tangle.MessageFactory.GetIssuingTime(parents)
+	sequenceNumber, err := deps.Tangle.MessageFactory.NextSequenceNumber()
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, jsonmodels.DataResponse{Error: err.Error()})
+	}
+	nonce, err := deps.Tangle.MessageFactory.DoPOW(references, issuingTime, issuerPublicKey, sequenceNumber, msg.Payload())
```
daria305, line 60: Out of curiosity, have you considered the version with PoW done by the requester, not by the node? Probably this sequence number would be a problem

daria305 (changes requested):
Looks good, I left some comments you might consider to add

plugins/webapi/tools/message/send_message.go
acha-bill: yeah. We shouldn't care about this.

plugins/webapi/tools/message/send_message.go
acha-bill: Not put any thought into this. :)

tools/integration-tests/tester/tests/diagnostics/diagnostics_test.go
acha-bill: Great idea actually. We should have the API only accept parents and payload. :)

plugins/webapi/tools/message/send_message.go
```diff
@@ -0,0 +1,100 @@
...
+	signature, err := deps.Tangle.MessageFactory.Sign(references, issuingTime, issuerPublicKey, sequenceNumber, msgPayload, nonce)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, jsonmodels.DataResponse{Error: err.Error()})
+	}
+	msg, err := tangle.NewMessage(references, issuingTime, issuerPublicKey, sequenceNumber, msgPayload, nonce, signature)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, jsonmodels.DataResponse{Error: err.Error()})
+	}
+
+	go deps.Tangle.MessageFactory.Events.MessageConstructed.Trigger(msg)
```
jkrvivian, line 75: What about wrapping those in a function in the messagefactory?
Maybe name the function to `IssuePayloadWithParents`, then there's no need to export doPoW, sign and such.

daria305 (approved):

Issue #2022, opened 2022-02-10
Abstract transaction related APIs, spammers and implement evil wallet


jkrvivian:
Evil wallet lib should provide easy-to-use facilities to do, along others:

* [x] Move spammer tools to GoShimmer (as a library with building blocks, no ready tools)
* [x] refactor spammer tools
* [x] evil wallet design spamming
* [x] Wrap up transaction-related APIs
  * [x] wait for message confirmation
  * [x] issue a double-spend without dealing with keys and outputs
  * [x] issue simple double-spend
  * [x] issue nested double-spend: double-spend a double-spent output
* [x] Make output handling automated - a user does not need to specifically provide output.

karimodm:
Also refer to https://github.com/iotaledger/goshimmer-tools/issues/38

daria305:
Also related PR: #2124
With that all basic functionality supporting deep spam and faucet requesting is done and issue can be closed

PR #2123, opened 2022-03-14
Evil Wallet: automate deep spamming

daria305:
Solves issue #2022 
Last fixes for evil wallet faucet requests. Evil wallet should now automatically handle inputs assigning and allow for requesting funds in 3 batch sizes: 100, 10000, n x 10000.
At this stage, it should be enough to replace code in integration tests and go further with spammer and tooling integration.


client/evilwallet/wallets.go
```diff
@@ -73,28 +74,60 @@ func (w *Wallets) GetUnspentOutput(wallet *Wallet) *Output {
...
+		}
+	}
+	return wallet, nil
 }
 
-func (w *Wallets) RemoveWallet() *Wallet {
+// removeWallet removes wallet, for fresh wallet it will be the first wallet in a queue.
+func (w *Wallets) removeWallet(wType WalletType) {
+	switch wType {
+	case fresh:
```
jkrvivian, line 106: What about `reuse` type?

client/evilwallet/wallets.go
```diff
@@ -73,28 +74,60 @@ func (w *Wallets) GetUnspentOutput(wallet *Wallet) *Output {
...
+		w.removeWallet(fresh)
+		if !w.IsFaucetWalletAvailable() {
+			return nil, errors.New("no faucet wallets available, need to request more funds")
+		}
+		// take next wallet
+		wallet = w.wallets[w.faucetWallets[0]]
+		if wallet.IsEmpty() {
+			return nil, errors.New("wallet is empty, need to request more funds")
+		}
+	}
```
jkrvivian, line 99: `IsFaucetWalletAvailable` should be checked before retrieving the wallet. Then the process can be in a loop, return a valid wallet until we get one, or return an error instead.

client/evilwallet/evilwallet.go
```diff
@@ -346,12 +344,23 @@ func (e *EvilWallet) registerOutputAliases(outputs ledgerstate.Outputs, addrAlia
...
+			if err != nil {
+				return nil, err
+			}
+		}
 		// No output found for given alias, use internal fresh output if wallets are non-empty.
 		if !ok {
+			err = e.getIssuerWallet(buildOptions)
+			if err != nil {
+				return nil, err
+			}
```
jkrvivian, line 362: Put these in `else`.

client/evilwallet/wallets.go
```diff
@@ -14,11 +14,14 @@ import (
...
 type WalletType int8
+type WalletStatus int8
 
 const (
-	// fresh is used for automatic Faucet Requests, outputs are returned one by one
 	other WalletType = iota
+	// fresh is used for automatic Faucet Requests, outputs are returned one by one
 	fresh
+	// reuse stores resulting outputs of double spends or transactions issued by the evilWallet
+	reuse
```
jkrvivian, line 24: When is a wallet labeled as `reuse`?

client/evilwallet/wallets.go
daria305: not sure If I understand correctly, but `IsFaucetWalletAvailable` is already checked in line 85 and the error is returned, as the check simply says that there are no wallets at all, so any further action makes no sense. The empty check is checking if all outputs were used, if yes --> take the next wallet ( that should not be empty - line 96 is just in case but should never happen )

client/evilwallet/wallets.go
daria305: The reuse wallet should contain outputs that are the leafs in the custom conflict tree - outputs that were not used during the custom conflict creation. The idea is to store them in a separate wallet for further reuse in spamming - we may provide custom conflicts struct as a spam secenario and instruct that spammer can reuse those outputs for deep spamming.
However I put this just to remember the idea, it's not implemented, we might discuss if it's the right direction, otherwise we can remove this category.

client/evilwallet/wallets.go
daria305: The `reuse` wallets are not implemented for now, I explined in the further comment.

client/evilwallet/wallets.go
jkrvivian: Ah ok! then maybe having a `GetWallet` function to retrieve one that is not empty would be cleaner?

client/evilwallet/wallets.go
```diff
@@ -80,20 +93,45 @@ func (w *Wallets) GetUnspentOutput(wallet *Wallet) *Output {
...
+// FreshWallet returns the first non-empty wallet from the faucetWallets queue. If current wallet is empty,
+// it is removed and the next enqueued one is returned.
+func (w *Wallets) FreshWallet() (*Wallet, error) {
+	wallet, err := w.GetNextWallet(fresh)
+	if err != nil {
+		w.removeWallet(fresh)
+		wallet, err = w.GetNextWallet(fresh)
+		if err != nil {
+			return nil, err
+		}
```
jkrvivian, line 110: I think these logics should be in GetNextWallet, it should iterate the whole wallet list, remove the empty one and return one if it's usable.

daria305: Why do we want to iterate over this list int the first place? As this list is basically a queue, new wallets are added at the end, currently used wallet is at index 0, if is used up we get next from queue. We never use wallets at index different than 0.

I see that I did not add switch for wallet type
Also I will add a  rule for `SetWalletReady` to not add empty wallets just in case.

jkrvivian: I mean something like this in `GetWallet`:
```go
for _, wallet := range w.wallets {
    if wallet.IsEmpty() {
	w.removeWallet(fresh)
    }
    return w, nil
}

return nil, errors.New("wallet type not supported for ordered usage, use GetWallet by ID instead")
```

jkrvivian (approved):

PR #2132, opened 2022-03-21
Fix Reference Creation

jonastheis:
The transaction issued in this message broke the devnet:  http://public-node-01.devnet.shimmer.iota.cafe:8081/explorer/message/CK6x17YPtryqfJmXGzPTRxfepnpbcaM4DAM6kpBov8et

**Problem:**  It reintroduced an old conflict branch. So naturally, every node wanted to set a like reference to the branch that was "winning". However, that branch was created in a message 3 days ago and so the like reference could not be set because it's too old (max parent age is 30min).

**Bug:** Now, every time a node selects tips, this message ends up in the strong parents (because there are only so few tips), and currently we throw the error message below if any of the strong parents could not be referenced. That, in turn, means that every time a node tried to select tips it failed. -> no ne messages can be created -> all nodes run out of sync

**Solution:** In this case the message should just be orphaned, i.e., if we can't express the opinion on it we can't select it as a tip and it should be removed and its parent re-added to the tip pool.

### DIFF
--- a/packages/tangle/messagefactory.go
+++ b/packages/tangle/messagefactory.go
@@ -324,9 +324,11 @@ var ZeroWorker = WorkerFunc(func([]byte) (uint64, error) { return 0, nil })
 // ReferencesFunc is a function type that returns like references a given set of parents of a Message.
 type ReferencesFunc func(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error)
 
-// PrepareReferences is an implementation of LikeReferencesFunc.
+// PrepareReferences is an implementation of ReferencesFunc.
 func PrepareReferences(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error) {
 	references = NewParentMessageIDs()
+	// TODO: these messages should be removed from the tips and their parents re-added.
+	referenceNotPossible := NewMessageIDs()
 
 	for strongParent := range strongParents {
 		if strongParent == EmptyMessageID {
@@ -344,8 +346,12 @@ func PrepareReferences(strongParents MessageIDs, issuingTime time.Time, tangle *
 		opinionCanBeExpressed := true
 		for strongParentBranchID := range strongParentBranchIDs {
 			referenceParentType, referenceMessageID, err := referenceFromStrongParent(tangle, strongParentBranchID, issuingTime)
+			// Explicitly ignore error since we can't create a like/dislike reference to the message.
+			// We need to simply ignore it and remove it from the strong parents.
 			if err != nil {
-				return nil, errors.Errorf("failed to determine valid reference from Branch with %s: %w", strongParentBranchID, err)
+				referenceNotPossible.Add(strongParent)
+				opinionCanBeExpressed = false
+				break
 			}
 
 			if referenceParentType == UndefinedParentType {
--- a/packages/tangle/messagefactory.go
+++ b/packages/tangle/messagefactory.go
@@ -125,7 +125,7 @@ func (f *MessageFactory) issuePayload(p payload.Payload, references ParentMessag
 		issuingTime = f.getIssuingTime(strongParents)
 		if len(references) == 0 {
 			var referenceNotPossible MessageIDs
-			references, err, referenceNotPossible = f.referencesFunc(strongParents, issuingTime, f.tangle)
+			references, referenceNotPossible, err = f.referencesFunc(strongParents, issuingTime, f.tangle)
 			for m := range referenceNotPossible {
 				f.Events.Error.Trigger(errors.Errorf("References for %s could not be determined", m))
 				f.Events.MessageReferenceImpossible.Trigger(m)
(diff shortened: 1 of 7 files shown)


packages/tangle/messagefactory.go
```diff
@@ -344,8 +346,12 @@ func PrepareReferences(strongParents MessageIDs, issuingTime time.Time, tangle *
 		opinionCanBeExpressed := true
 		for strongParentBranchID := range strongParentBranchIDs {
 			referenceParentType, referenceMessageID, err := referenceFromStrongParent(tangle, strongParentBranchID, issuingTime)
+			// Explicitly ignore error since we can't create a like/dislike reference to the message.
+			// We need to simply ignore it and remove it from the strong parents.
 			if err != nil {
-				return nil, errors.Errorf("failed to determine valid reference from Branch with %s: %w", strongParentBranchID, err)
+				referenceNotPossible.Add(strongParent)
+				opinionCanBeExpressed = false
```
piotrm50, line 353: This should be logged somehow.

packages/tangle/messagefactory.go
```diff
@@ -324,9 +324,11 @@ var ZeroWorker = WorkerFunc(func([]byte) (uint64, error) { return 0, nil })
 // ReferencesFunc is a function type that returns like references a given set of parents of a Message.
 type ReferencesFunc func(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error)
 
-// PrepareReferences is an implementation of LikeReferencesFunc.
+// PrepareReferences is an implementation of ReferencesFunc.
 func PrepareReferences(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error) {
 	references = NewParentMessageIDs()
+	// TODO: these messages should be removed from the tips and their parents re-added.
```
piotrm50, line 330: Don't we want to implement this too before merging?

packages/tangle/messagefactory.go
```diff
@@ -344,8 +346,12 @@ func PrepareReferences(strongParents MessageIDs, issuingTime time.Time, tangle *
 		opinionCanBeExpressed := true
 		for strongParentBranchID := range strongParentBranchIDs {
 			referenceParentType, referenceMessageID, err := referenceFromStrongParent(tangle, strongParentBranchID, issuingTime)
+			// Explicitly ignore error since we can't create a like/dislike reference to the message.
+			// We need to simply ignore it and remove it from the strong parents.
 			if err != nil {
-				return nil, errors.Errorf("failed to determine valid reference from Branch with %s: %w", strongParentBranchID, err)
+				referenceNotPossible.Add(strongParent)
```
karimodm, line 352: A code bit is missing no?

Later if `if !opinionCanBeExpressed {` we should introduce this strong parent as a weak reference if we like all its payload branches (we already do this bit) and in any case drop the message from the tippool and add its strong parents if they are not beyond `maxParentAge`.

packages/tangle/messagefactory_test.go
```diff
@@ -358,7 +363,11 @@ func checkReferences(t *testing.T, tangle *Tangle, parents MessageIDs, expectedR
 	}
 	require.NoError(t, err)
 
-	assert.Equal(t, parents, actualReferences[StrongParentType])
+	if expectedStrongParents, ok := expectedReferences[StrongParentType]; ok {
```
karimodm, line 366: Isn't this already covered by the loop over all parents type on line 372?

karimodm (changes requested):

packages/tangle/messagefactory.go
jonastheis: Done

packages/tangle/messagefactory.go
jonastheis: Triggering an error where the function is called

packages/tangle/messagefactory_test.go
jonastheis: I made the way we call the `checkreferences` function explicit. So we can handle strong parents the same as the others.

packages/tangle/messagefactory.go
jonastheis: Handled the return now in a new event and attaching to it in the `TipManager`

packages/tangle/messagefactory.go
```diff
@@ -322,11 +331,12 @@ var ZeroWorker = WorkerFunc(func([]byte) (uint64, error) { return 0, nil })
 // region PrepareLikeReferences ///////////////////////////////////////////////////////////////////////////////////////////////////
 
 // ReferencesFunc is a function type that returns like references a given set of parents of a Message.
-type ReferencesFunc func(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error)
+type ReferencesFunc func(strongParents MessageIDs, issuingTime time.Time, tangle *Tangle) (references ParentMessageIDs, err error, referenceNotPossible MessageIDs)
```
piotrm50, line 334: Put error at the end of return list.

piotrm50 (approved):

karimodm (approved):

Issue #2104, opened 2022-03-07
Create ready scenarios


daria305:
Add evil wallet ready scenarios, for spamming and single usage, possible examples:
  * [x] Create conflict sets with >100 conflict members
  * [x] Create conflicts where a conflict is member of different conflict sets
  * [x] Create nested conflicts
All options above are easy to create with evilwallet and evilspammer functionality.
Deep spam is done with SendCustomConflicts by providing ConflictScice, e.g.
```go
EvilBatch{
		[]ScenarioAlias{
			{Inputs: []string{"1"}, Outputs: []string{"2", "3"}},
		},
		[]ScenarioAlias{
			{Inputs: []string{"2"}, Outputs: []string{"4"}},
			{Inputs: []string{"2"}, Outputs: []string{"5"}},
		},
		[]ScenarioAlias{
			{Inputs: []string{"6", "5"}, Outputs: []string{"7"}},
		},
	}
```
Spamer uses EvilScenarios to define the spam logic. It is possible to deep spam by enabling reusage of outputs generated during the spam, or providing deep conflict batch.

PR #2140, opened 2022-03-30
Implement evil scenario

jkrvivian:
# Description of change

related to #2104 

* Fix WithInput option
* Fix taking wrong slice in  `areOutputsProvidedWithoutAliases`
* Add ScenarioAlias to define custom spamming scenario easily
* Get output information from internal map and web api

## Type of change

- Enhancement
- Bug Fix

## Change checklist

### DIFF
--- a/client/evilwallet/customscenarios.go
+++ b/client/evilwallet/customscenarios.go
@@ -5,13 +5,21 @@ import (
 )
 
 func SingleTransactionBatch() EvilBatch {
-	return EvilBatch{{[]Option{WithInputs("1"), WithOutput(&OutputOption{aliasName: "2"})}}}
+	return EvilBatch{{ScenarioAlias{inputs: []string{"1"}, outputs: []string{"2"}}}}
 }
 
 func DoubleSpendBatch(spentNum int) EvilBatch {
-	conflictSlice := make(ConflictSlice, 0)
+	conflictSlice := make(EvilBatch, 0)
+	aliasCounter := 1
+	// each double spend uses a different unspent output
+	inputStartNum := spentNum * 2
+
 	for i := 1; i <= spentNum; i++ {
-		conflictSlice = append(conflictSlice, []Option{WithInputs("0"), WithOutput(&OutputOption{aliasName: strconv.Itoa(i)})})
+		conflictSlice = append(conflictSlice, []ScenarioAlias{
+			{inputs: []string{strconv.Itoa(inputStartNum + i)}, outputs: []string{strconv.Itoa(aliasCounter)}},
+			{inputs: []string{strconv.Itoa(inputStartNum + i)}, outputs: []string{strconv.Itoa(aliasCounter + 1)}},
+		})
+		aliasCounter += 2
 	}
-	return EvilBatch{conflictSlice}
+	return conflictSlice
 }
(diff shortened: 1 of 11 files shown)


client/evilwallet/evilwallet.go
```diff
@@ -686,43 +687,96 @@ func (e *EvilWallet) updateOutputIDs(txID ledgerstate.TransactionID, outputs led
 }
 
 func (e *EvilWallet) PrepareTransaction(scenario *EvilScenario) (tx *ledgerstate.Transaction, err error) {
-	wallet, err := e.wallets.FreshWallet()
+	inWallet, err := e.wallets.GetNextWallet(Fresh)
```
daria305, line 690: maybe we should change FreshWalet to lowercase, as I did the same mistake that you corrected  on my new branch with ReuseWallet 😄

client/evilwallet/output_manager.go
```diff
@@ -174,15 +174,31 @@ func (o *OutputManager) UpdateOutputsFromTxs(txIDs []string) error {
 
 // GetOutput returns the Output of the given outputID.
 func (o *OutputManager) GetOutput(outputID ledgerstate.OutputID) (output *Output) {
```
daria305, line 176: What I don't like about this function is that we make an API call each time, so during the spam we send API request 100*number of inputs per second, and we usually have all needed informations stored in *evilwallet.Output and can simply recreate an input like 
```
evilInput := wallet.GetUnspentOutput()
outBalance := getIotaColorAmount(evilInput.Balance)
out := ledgerstate.NewSigLockedColoredOutput(evilInput.Balance, evilInput.Address)
input := out.SetID(evilInput.OutputID)
```

daria305 (approved):

client/evilwallet/output_manager.go
jkrvivian: We only request api if the provided unspent output is not in the wallets (line 177), I think this could be useful in some cases, users can provide valid input even though it doesn't belong to one of the wallet we managed.

client/evilwallet/customscenarios.go
```diff
@@ -5,34 +5,40 @@ import (
 )
 
 func SingleTransactionBatch() EvilBatch {
-	return EvilBatch{{[]Option{WithInputs("1"), WithOutput(&OutputOption{aliasName: "2"})}}}
+	return EvilBatch{{ScenarioAlias{Inputs: []string{"1"}, Outputs: []string{"2"}}}}
 }
 
 func DoubleSpendBatch(spentNum int) EvilBatch {
```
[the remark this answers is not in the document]
jkrvivian: ah ok, I misunderstood your design here :stuck_out_tongue_closed_eyes:

client/evilwallet/output_manager.go
daria305: ok, I did not notice that it is not used always, then it's alright

client/evilwallet/evilwallet.go
jkrvivian: I'm confused, which do you prefer, `freshWallet()` or `GetNextWallet` with type as a parameter?

daria305: GetNextWallet, I just forgot about it

Issue #2126, opened 2022-03-15
Refactor evil wallet, spammer, tool


jkrvivian:
* [x] Remove testing files in evilwallet

pqmertz:
Can be done with two nodes, cli-wallet and copied wallet.dat file from one to the other.   request-funds setup a cron job to send a token every minute.   Bam conflicts and pending funds.

daria305:
Clever, but we're building the tooling that will also allows for creation of deep structures in branch and UTXO DAG. See  screenshots from: #2145

PR #2162, opened 2022-04-19
Parametrize ds filter

daria305:
Part of  #2126
- Add parameter to disable double-spend filter
- Disable the double-spend filter by default

### DIFF
--- a/plugins/webapi/ledgerstate/plugin.go
+++ b/plugins/webapi/ledgerstate/plugin.go
@@ -3,6 +3,7 @@ package ledgerstate
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


plugins/webapi/ledgerstate/plugin.go
```diff
@@ -63,24 +67,53 @@ func init() {
 
 // Filter returns the double spend filter singleton.
 func Filter() *DoubleSpendFilter {
+	fmt.Println("Enable Filter")
```
jkrvivian, line 70: Remove this debug log. :stuck_out_tongue:

PR #2167, opened 2022-04-21
Update TSC to compare tangle time with time of confirmed message.

piotrm50:
# Description of change

Solves #2160 

## Type of change

Choose a type of change, and delete any options that are not relevant.

- Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

Unit tests, manual testing. 

## Change checklist

Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [x] My code follows the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2193, opened 2022-05-05
New time concepts

georgysavva:
This PR adds new time types: CTT, RCTT, FTT, and RFTT to the TimeManager.
It also changes the logic from `synced` to `bootstrapped`.


packages/tangle/timemanager.go
```diff
@@ -136,7 +144,7 @@ func (t *TimeManager) synced() bool {
 func (t *TimeManager) updateSyncedState() {
```
karimodm, line 144: We don't need the `Synced` and `updateSyncedState` in their current form anymore right? We just want to have a concept of bootstrapped and a concept of synced that just checks a threshold between FTT and RCTT.

packages/tangle/timemanager.go
```diff
@@ -53,8 +56,9 @@ func NewTimeManager(tangle *Tangle) *TimeManager {
 
 	// initialize with Genesis
 	t.lastConfirmedMessage = LastConfirmedMessage{
-		MessageID: EmptyMessageID,
-		Time:      time.Unix(DefaultGenesisTime, 0),
+		MessageID:     EmptyMessageID,
+		Time:          time.Unix(DefaultGenesisTime, 0),
```
karimodm, line 60: Let's add some comments on these fields, as it may be confusing now with two times.

packages/tangle/tipmanager.go
```diff
@@ -375,7 +375,7 @@ func (t *TipManager) isPastConeTimestampCorrect(messageID MessageID) (timestampV
 	}
 
 	// if last confirmed message if older than minSupportedTimestamp, then all tips are invalid
-	if t.tangle.TimeManager.LastConfirmedMessage().Time.Before(minSupportedTimestamp) {
+	if t.tangle.TimeManager.CurrentTangleTime().Before(minSupportedTimestamp) {
```
karimodm, line 378: We do not want to invalidate tips if we stop confirmations. I think this time should be RCTT.

plugins/networkdelay/plugin.go
```diff
@@ -125,7 +125,7 @@ func sendToRemoteLog(networkDelayObject *Payload, receiveTime int64) {
 		ReceiveTime: receiveTime,
 		Delta:       receiveTime - networkDelayObject.SentTime,
 		Clock:       clockEnabled,
-		Sync:        deps.Tangle.Synced(),
+		Sync:        deps.Tangle.Bootstrapped(),
```
karimodm, line 128: Should we change this filed to`Bootstrapped` too?

plugins/networkdelay/plugin.go
```diff
@@ -139,7 +139,7 @@ func sendPoWInfo(payload *Payload, powDelta time.Duration) {
 		ReceiveTime: 0,
 		Delta:       powDelta.Nanoseconds(),
 		Clock:       clockEnabled,
-		Sync:        deps.Tangle.Synced(),
+		Sync:        deps.Tangle.Bootstrapped(),
```
karimodm, line 142: same here

plugins/remotemetrics/scheduler.go
```diff
@@ -18,7 +18,7 @@ func obtainSchedulerStats(timestamp time.Time) {
 	record := remotemetrics.SchedulerMetrics{
 		Type:                         "schedulerSample",
 		NodeID:                       myID,
-		Synced:                       deps.Tangle.Synced(),
+		Synced:                       deps.Tangle.Bootstrapped(),
```
karimodm, line 21: Change field to Bootstrapped?

plugins/webapi/info/plugin.go
```diff
@@ -100,11 +100,15 @@ func getInfo(c echo.Context) error {
 	sort.Strings(disabledPlugins)
 
 	// get TangleTime
-	lcm := deps.Tangle.TimeManager.LastConfirmedMessage()
+	tm := deps.Tangle.TimeManager
+	lcm := tm.LastConfirmedMessage()
 	tangleTime := jsonmodels.TangleTime{
-		Synced:    deps.Tangle.TimeManager.Synced(),
-		Time:      lcm.Time.UnixNano(),
+		Synced:    deps.Tangle.TimeManager.Bootstrapped(),
```
karimodm, line 106: same as above

karimodm (changes requested):

packages/tangle/timemanager.go
georgysavva: Done

packages/tangle/tipmanager.go
georgysavva: Done

plugins/webapi/info/plugin.go
georgysavva: Done

packages/tangle/timemanager.go
georgysavva: I thought we leave the synced condition as is and just update it once. When the node becomes the first time synced, it means it bootstrapped, and we don't update synced status anymore.

packages/tangle/tipmanager.go
karimodm: I am sorry, but I was actually wrong; here we have to evaluate the difference between CTT (which won't advance if confirmations stop) and the timestamp of the oldest confirmed message of the tip's past cone. If we would compare the latter with RCTT or wall clock we will end up invalidating all tips.

karimodm (changes requested):

packages/tangle/timemanager.go
karimodm: The bootstrapped flag is the only one that affects algorithms: until the flag turns true we won't emit messages (blocks in the new terminology). This flag won't transition to false once set to true.
The synced flag instead has no effect in the code whatsoever, and it is purely a user-facing value. This flag can transition back and forth between true and false (as it is now).

packages/tangle/tipmanager.go
piotrm50: This is updated in another PR, so we can resolve it there and not change it here.

packages/tangle/timemanager.go
```diff
@@ -107,17 +110,22 @@ func (t *TimeManager) LastConfirmedMessage() LastConfirmedMessage {
 	return t.lastConfirmedMessage
 }
 
-// Time returns the TangleTime, i.e., the issuing time of the last confirmed message.
-func (t *TimeManager) Time() time.Time {
+// CurrentTangleTime returns the CTT, i.e., the issuing time of the last confirmed message.
+func (t *TimeManager) CurrentTangleTime() time.Time {
```
piotrm50, line 114: Shouldn't it be ConfirmedTangleTime instead as we have FinalizedTangleTime below?

packages/tangle/timemanager.go
```diff
@@ -152,16 +160,44 @@ func (t *TimeManager) updateTime(messageID MessageID) {
...
 		t.updateSyncedState()
 	})
 }
 
+// RCTT return relative current tangle time.
+func (t *TimeManager) RCTT() time.Time {
+	newRCTT := t.calculateNewRCTT()
+	t.lastRCTTMutex.Lock()
+	defer t.lastRCTTMutex.Unlock()
+	if newRCTT.After(t.lastRCTT) {
```
piotrm50, line 177: When confirmations stop for some time, then RCTT will go forward, but then when confirmations resume, RCTT will stop until confirmations catch up with the last RCTT value (or at most the difference between new CTT and the last RCTT value) - is this a desired property?

packages/tangle/timemanager.go
```diff
@@ -181,6 +217,7 @@ func (t *TimeManager) mainLoop() {
 type LastConfirmedMessage struct {
 	MessageID MessageID `serix:"0"`
 	Time      time.Time `serix:"1"`
+	ConfirmedTime time.Time `serix:"2"`
```
piotrm50, line 220: property naming is a little confusing - which one stores time of a message and which one the time when the message is confirmed? Would be nice to at least have some comment here.

Also update the comment on the whole struct.

plugins/dashboard/plugin.go
```diff
@@ -225,8 +225,12 @@ type nodestatus struct {
 }
 
 type tangleTime struct {
-	Synced    bool   `json:"synced"`
-	Time      int64  `json:"time"`
+	Synced bool  `json:"synced"`
+	CTT    int64 `json:"time"`
```
piotrm50, line 229: maybe change the json field name to be in line with the others

plugins/metrics/sync.go
```diff
@@ -7,7 +7,7 @@ import (
 var isTangleTimeSynced atomic.Bool
 
 func measureSynced() {
-	tts := deps.Tangle.TimeManager.Synced()
```
piotrm50, line 10: we could add the synced field to the metrics here as well so it's visible in Grafana dashboard. But this can be done in a separate PR which also updates the dashboard.

plugins/remotemetrics/message.go
```diff
@@ -19,10 +19,6 @@ func onMessageDiscarded(messageID tangle.MessageID) {
 }
 
 func sendMessageSchedulerRecord(messageID tangle.MessageID, recordType string) {
-	if !deps.Tangle.Synced() {
```
piotrm50, line 22: Can we use bootstrapped here? So the nodes that are bootstrapping won't spam our metrics.

plugins/remotemetrics/branch.go
```diff
@@ -48,7 +48,7 @@ func onBranchConfirmed(branchID ledgerstate.BranchID) {
 	// update branch metric counts even if node is not synced.
 	oldestAttachmentTime, oldestAttachmentMessageID, err := updateMetricCounts(branchID, transactionID)
 
-	if err != nil || !deps.Tangle.Synced() {
+	if err != nil {
```
piotrm50, line 51: Can we use bootstrapped here?

plugins/remotemetrics/sync.go
```diff
@@ -11,20 +11,23 @@ var isTangleTimeSynced atomic.Bool
 
 func checkSynced() {
 	oldTangleTimeSynced := isTangleTimeSynced.Load()
-	tts := deps.Tangle.TimeManager.Synced()
```
piotrm50, line 14: Actually we'd be more interested in nodes running out of sync as in RCTT being different than CTT.

plugins/webapi/info/plugin.go
piotrm50, line 106: Update the property name to 'Boostrapped' to have consistent naming. I think it would be nice to have another field 'synced' that is based on the difference  RCTT and CTT for API stuff and monitoring metrics.

plugins/webapi/tools/message/send_message.go
```diff
@@ -17,10 +17,6 @@ const maxIssuedAwaitTime = 5 * time.Second
 
 // SendMessage is the handler for tools/message endpoint.
 func SendMessage(c echo.Context) error {
-	if !deps.Tangle.Synced() {
```
piotrm50, line 20: The node needs to be bootstrapped to be able to issue a block. Is this checked in some other place?

piotrm50 (changes requested):

packages/tangle/tipmanager.go
georgysavva: @karimodm do you agree with @piotrm50? Can I resolve and leave it as is?

packages/tangle/timemanager.go
georgysavva: Done

packages/tangle/timemanager.go
georgysavva: @daria305 can you participate?

packages/tangle/timemanager.go
georgysavva: Done

plugins/dashboard/plugin.go
georgysavva: I didn't want to break anything when changing the JSON field name, because I don't know where else it is used (maybe some other repositories or by community members)

plugins/remotemetrics/message.go
georgysavva: Done

plugins/remotemetrics/branch.go
georgysavva: Done

plugins/remotemetrics/sync.go
georgysavva: Can you explain a bit better what you want to change here?

plugins/webapi/info/plugin.go
georgysavva: Done

plugins/webapi/tools/message/send_message.go
georgysavva: Fixed

plugins/dashboard/plugin.go
georgysavva: Done

packages/tangle/timemanager.go
georgysavva: @daria305 can you participate?

packages/remotemetrics/events.go
```diff
@@ -37,8 +37,14 @@ type SyncStatusChangedEvent struct {
 	CurrentStatus bool `json:"currentStatus" bson:"currentStatus"`
 	// PreviousStatus contains previous sync status
 	PreviousStatus bool `json:"previousStatus" bson:"previousStatus"`
-	// LastConfirmedMessageTime contains time of the last confirmed message
-	LastConfirmedMessageTime time.Time `json:"lastConfirmedMessageTime" bson:"lastConfirmedMessageTime"`
+	// CTT contains time of the last confirmed message
```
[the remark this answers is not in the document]
georgysavva: We still trigger this event when the node falls out of sync. It happens if FTT - wall time is less than the threshold.

plugins/metrics/sync.go
georgysavva: Done

plugins/remotemetrics/sync.go
georgysavva: Done

packages/tangle/timemanager.go
karimodm: RCTT should always be ahead of CTT. So we do not want to have this check here.

packages/tangle/tipmanager.go
```diff
@@ -375,7 +375,7 @@ func (t *TipManager) isPastConeTimestampCorrect(messageID MessageID) (timestampV
 	}
 
 	// if last confirmed message if older than minSupportedTimestamp, then all tips are invalid
```
karimodm, line 377: Can you change the comment here? This condition should be now very rare.

karimodm (changes requested):

packages/tangle/tipmanager.go
karimodm: Yeah, I think we should merge this PR into the other.

packages/tangle/tipmanager.go
karimodm: This check has been actually removed in @piotrm50 PR.

georgysavva: Done

packages/tangle/timemanager.go
georgysavva: Done

plugins/remotemetrics/message.go
piotrm50: Is it? I meant still having this check but like the following:
```
	if !deps.Tangle.Bootstrapped() 
```
This way nodes won't spam our metrics before they bootstrap.

piotrm50: Already fixed it

piotrm50 (approved):

georgysavva:
@jonastheis @karimodm are you guys ok with the PR? Can we merge it into https://github.com/iotaledger/goshimmer/pull/2167 already

karimodm (approved):

jonastheis (approved):

PR #2200, opened 2022-05-06
Fix: snapshot path for public nodes

karimodm:


jonastheis (commented):
Not sure this is valid anymore with https://github.com/iotaledger/goshimmer/pull/2201

karimodm:
It is, it aligns node's start configuration.

karimodm:
Removed snapshot path from all nodes' compose templates, use Dockerfile-defined defaults instead.

PR #2228, opened 2022-05-30
Feat/update block structure

acha-bill:
- Refactors notarisation plugin (move it to message layer)
- Refactor epoch commitment (moves it to tangle)
- Updates block structure


packages/notarization/commitments.go
```diff
@@ -9,30 +9,19 @@ import (
...
-}
-
 // Commitment is a compressed form of all the information (messages and confirmed value payloads) of an epoch.
 type Commitment struct {
 	EI                EI
 	tangleRoot        *smt.SparseMerkleTree
 	stateMutationRoot *smt.SparseMerkleTree
 	stateRoot         *smt.SparseMerkleTree
-	prevECR           []byte
+	prevECR           [32]byte
```
daria305, line 20: It should be prevEC thatis equal to Hash(prevEC+ECR+EI)
Also it would be good to introduce type that will hide [32]byte

georgysavva (approved):

PR #2234, opened 2022-06-01
Delete unused modules

jonastheis:
This PR cleans up the code base from unused code. We are not using the API `/tools` nor dRNG. In our upcoming refactors any additional code that needs to be maintained is a liability. Therefore, we need to make the whole code base as lean as possible.


hmoog (approved):
LGTM(BQ) :p

PR #2265, opened 2022-06-09
Snapshot mana epochs

georgysavva:


plugins/messagelayer/mana_plugin.go
```diff
@@ -799,11 +799,45 @@ func QueryAllowed() (allowed bool) {
...
+			}
+			return nil
+		})
+
+	}
+	processOutputs(*snapshot.Outputs, snapshot.OutputsMetadata, true /* isCreated */)
+	for i := snapshot.DiffEpochIndex; i < snapshot.FullEpochIndex; i++ {
+		diff, exists := snapshot.EpochDiffs.Get(i)
+		if !exists {
+			continue
```
karimodm, line 835: This should be a fatal error. There is something wrong with the snapshot if we miss a diff in between fullindex and diffindex.

packages/notarization/epoch.go
```diff
@@ -0,0 +1,26 @@
+package notarization
```
karimodm, line 1: I've created a models.go file, these should go in there.

packages/snapshot/snapshot.go
```diff
@@ -18,16 +15,15 @@ import (
 	"github.com/iotaledger/goshimmer/packages/ledger/utxo"
 	"github.com/iotaledger/goshimmer/packages/ledger/vm/devnetvm"
 	"github.com/iotaledger/goshimmer/packages/mana"
+	"github.com/iotaledger/goshimmer/packages/notarization"
 )
 
 type Snapshot struct {
 	LedgerSnapshot *ledger.Snapshot
-	ManaSnapshot   *mana.Snapshot
```
karimodm, line 25: Can we just have the inner struct as this now has only one field?

packages/mana/consensusbasevector.go
```diff
@@ -92,29 +92,29 @@ func txInfoFromPledgeEvent(ev *PledgedEvent) *TxInfo {
...
-				ManaType:      c.Type(),
-				TransactionID: record.TxID,
-			})
-		}
-
-		c.vector[nodeID] = &ConsensusBaseMana{
-			BaseMana1: value,
-		}
-	}
+func (c *ConsensusBaseManaVector) LoadSnapshot(snapshot map[identity.ID]float64) {
```
karimodm, line 103: Here we just store the base vector right?
Also let's maybe change the function name.

karimodm (changes requested):

tools/genesis-snapshot/snapshotcreator/snapshotcreator.go
```diff
@@ -56,7 +56,6 @@ func CreateSnapshot(genesisTokenAmount uint64, genesisSeedBytes []byte, nodesToP
 
 	return &snapshot.Snapshot{
 		LedgerSnapshot: ledger.NewSnapshot(outputs, outputsMetadata),
```
daria305, line 58: `manaSnapshot := mana.NewSnapshot()` above can also be removed as it is no longer used

packages/snapshot/snapshot.go
georgysavva: I would leave it as is for now. Maybe we will add another field in the future, then we would need to change it back.

packages/mana/consensusbasevector.go
georgysavva: Yes, we populate the vector with the map data.
Sure, I changed the name.

packages/mana/accessbasevector.go
```diff
@@ -38,25 +38,25 @@ func (a *AccessBaseManaVector) Has(nodeID identity.ID) bool {
...
-		}
-		// trigger events
-		Events.Pledged.Trigger(&PledgedEvent{
-			NodeID:   nodeID,
-			Amount:   record.AccessMana.Value,
-			Time:     record.AccessMana.Timestamp,
-			ManaType: a.Type(),
-		})
-	}
+func (a *AccessBaseManaVector) LoadSnapshot(snapshot map[identity.ID]float64) {
```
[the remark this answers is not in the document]
georgysavva: Yes, we populate the vector with the map data.
Sure, I changed the name.

PR #2267, opened 2022-06-09
AccessMana is now Mana1

karimodm:
* package/mana/access* all gone
* consensusbase renamed to manabase implementing base interface for both access and consensus mana
* consensusbasevector renamed to manabasevector implementing basevector interface for both access and consensus mana
* both mana vectors are now maintained using 2 instances of ManaBaseVector (former ConsensusBaseVector)
* update methods removed from both interfaces and implementations
* SetCoefficients gone, along with corresponding mana plugin parameters
* EffectiveValue methods removed, only BaseValue remains
* manarefresher is gone
    * DelegationAddress stuff managed by the plugin also gone
        * Reclaim functionality removed
    * corresponding WebAPI endpoints and response fields removed
    * cli-wallet delegation & reclaim removed
    * occurrences removed from Ansible and docker-compose files
* GetMana and GetManaMap methods do not receive and optional time anymore, but they still return a timestamp which is always set to time.Now(). This timestamp should be tied to the epoch timestamp for the committed mana vector.


packages/mana/basevector.go
```diff
@@ -48,9 +44,9 @@ type BaseManaVector interface {
 func NewBaseManaVector(vectorType Type) (BaseManaVector, error) {
```
piotrm50, line 44: Do we need this function now? Both cases in the switch perform the same action anyway

packages/mana/events_test.go
```diff
@@ -92,8 +92,8 @@ func newRevokeEvent() *RevokedEvent {
 func newUpdateEvent() *UpdatedEvent {
 	return &UpdatedEvent{
 		NodeID:   randomNodeID(),
-		OldMana:  &ConsensusBaseMana{},
-		NewMana:  &ConsensusBaseMana{},
+		OldMana:  &ManaBase{},
+		NewMana:  &ManaBase{},
```
piotrm50, line 96: This is safer as it already initializes the models.

```suggestion
		OldMana:  NewManaBase(0.0),
		NewMana:  NewManaBase(0.0),
```

plugins/messagelayer/parameters.go
```diff
@@ -29,12 +29,6 @@ type ParametersDefinition struct {
 
 // ManaParametersDefinition contains the definition of the parameters used by the mana plugin.
 type ManaParametersDefinition struct {
-	// EmaCoefficient1 defines the coefficient used for Effective Base Mana 1 (moving average) calculation.
```
piotrm50, line 32: Did we want to remove the coefficient for mana1 as well? Or was it never actually used?

plugins/metrics/mana.go
```diff
@@ -202,6 +201,4 @@ func measureMana() {
 		consensusAvg = consensusSum / float64(len(neighborConsensusMap))
 	}
 	averageNeighborsConsensus.Store(consensusAvg)
-
-	delegationAmount.Store(manarefresher.TotalDelegatedFunds())
```
piotrm50, line 206: This is the only place where we set value of this variable. We should remove it completely from the metrics.

piotrm50 (changes requested):
Added comments for some minor changes.

plugins/messagelayer/parameters.go
karimodm: It was never actually used!

packages/mana/basevector.go
karimodm: True! I've turned it into a basic constructor.

jonastheis (approved):
LGTM but I think you'll need to adjust the (mana) integration tests as well

karimodm:
Yeah... will fix all the tests in the base branch.

karimodm:
Yeah... will fix all the tests in the base branch.

piotrm50 (approved):

PR #2291, opened 2022-06-16
Epoch commitments: Mana Root & fixes 

karimodm:
* Implement Mana Root
* Implement epoch-based diff storage
* Fix several bugs to properly produce commitments
* Code cleanup


packages/notarization/storage.go
```diff
@@ -94,36 +78,131 @@ func (s *EpochCommitmentStorage) CachedECRecord(ei epoch.Index, computeIfAbsentC
 	return s.ecRecordStorage.Load(ei.Bytes())
 }
 
+func (s *EpochCommitmentStorage) SetFullEpochIndex(ei epoch.Index) error {
```
daria305, line 81: We have here four different indexes and the below functions do the same and just use different key strings. Maybe we can refactor to one general function that receives ei and key string

packages/epoch/types.go
```diff
@@ -38,13 +39,25 @@ func EIFromBytes(bytes []byte) (ei Index, consumedBytes int, err error) {
...
+func (m MerkleRoot) Base58() string {
+	return base58.Encode(m[:])
+}
+
+func (m MerkleRoot) Bytes() []byte {
+	return m[:]
+}
+
 // Epoch is a time range used to define a bucket of messages.
 type Epoch struct {
```
daria305, line 62: We don't actually use it anywhere.

daria305 (approved):

daria305 (approved):

Issue #2270, opened 2022-06-13
EC chains display in Dashboard or Grafana


karimodm:
- Simple tracking of the chain that each node sees
	- committment per epoch Index according to node. Subscribe to the EpochCommitted event, extend the event to include commitments
	- mana weight (supporters) for each committment chain, per epoch. Subscribe to MessageConfirmed/MessageBooked event, using the message object go to the ApprovalManager and query the weight by the NodeID. Store weight per node per epoch. 
- API
	- UTXOs in the epoch. Subscribe to the TrasactionConfirmed Event.
	- Messages in the epoch. Subscribe to the MessagedConfirmed Event.
	- Transactions in the epoch. Subscribe to the TransactionConfirmed Event.
- Grafana
	- Last comitted epoch. Subscribe to a notarization event.
	- Pending branch count. Stored in the notarization manager.
	- Last committable epoch?

PR #2293, opened 2022-06-17
Epoch commitment metrics

georgysavva:
Fixes: https://github.com/iotaledger/goshimmer/issues/2270


packages/jsonmodels/info.go
```diff
@@ -32,6 +37,16 @@ type InfoResponse struct {
...
+
+	// Epoch related metrics.
+	EpochCommitments   map[epoch.Index]metrics.EpochInfo
+	EpochVotersWeight  map[epoch.Index]map[identity.ID]float64
+	EpochUTXOs         map[epoch.Index][]string
+	EpochMessages      map[epoch.Index][]string
+	EpochTransactions  map[epoch.Index][]string
+	LastCommittedEpoch metrics.EpochInfo
+	PendingBranchCount map[epoch.Index]uint64
+
```
karimodm, line 49: We need storage on all these

georgysavva: Yep. I will change that.

plugins/metrics/epoch_commitments.go
```diff
@@ -0,0 +1,235 @@
...
+			objectstorage.StoreOnCreation(true),
+		),
+	}
+	return metrics, nil
+}
+
+func (m *EpochCommitmentsMetrics) saveCommittedEpoch(ecr *epoch.ECRecord) {
+	m.epochCommitmentsMutex.Lock()
+	defer m.epochCommitmentsMutex.Unlock()
+	m.epochCommitments[ecr.EI()] = *ecr
```
karimodm, line 96: Can we keep it as a pointer to avoid memory duplication?

plugins/metrics/epoch_commitments.go
```diff
@@ -0,0 +1,235 @@
...
+		subDuplicate := make(map[identity.ID]float64, len(v))
+		for subK, subV := range v {
+			subDuplicate[subK] = subV
+		}
+		duplicate[k] = subDuplicate
+	}
+	return duplicate
+}
+
+func (m *EpochCommitmentsMetrics) saveEpochUTXOs(tx *devnetvm.Transaction) {
```
karimodm, line 145: I think it would be better to actually obtain the same information the notarization manager has. In fact, you cannot just attach to the TransactionConfirmation, you also have to take into account TransactionInclusionTimeUpdate events and such: and the information could still diverge.
In short: we want to obtain the exact data the notarization manager used to generate the commitment.

plugins/metrics/epoch_commitments.go
```diff
@@ -0,0 +1,235 @@
...
+	duplicate := make(map[epoch.Index][]string, 0)
+	m.epochUTXOs.ForEach(func(_ []byte, obj *objectstorage.CachedObject[*EpochContent]) bool {
+		ec := obj.Get()
+		duplicate[ec.M.EI] = ec.M.Content
+		return true
+	})
+	return duplicate
+}
+
+func (m *EpochCommitmentsMetrics) saveEpochMessage(msg *tangle.Message) {
```
karimodm, line 174: Same here. We want the exact data that the notarization manager used to create a certain commitment for an epoch.

karimodm (changes requested):

plugins/metrics/epoch_commitments.go
georgysavva: Done

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	committedEpochsMutex.Lock()
+	if len(committableEpochs) < maxEpochContentsToKeep {
+		committedEpochsMutex.Unlock()
+		return
+	}
+	committableEpochs = committableEpochs[len(committableEpochs)-maxEpochContentsToKeep:]
+	committedEpochsMutex.Unlock()
+}
+
+func GetCommittedEpochs() (ecRecords map[epoch.Index]*epoch.ECRecord) {
```
karimodm, line 194: Rename to GetCommittableEpochs like the committableEpochs var?

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+
+var (
+	// Plugin is the plugin instance of the messagelayer plugin.
+	Plugin *node.Plugin
+	deps   = new(dependencies)
+
+	epochContentsMutex     sync.RWMutex
+	epochContents          = make(map[epoch.Index]*epochContentStorages, 0)
+	committedEpochsMutex   sync.RWMutex
+	committableEpochs      = make([]*epoch.ECRecord, 0)
```
karimodm, line 39: Lock and var should have the same name, both Committable*?

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func GetEpochVotersWeight(ei epoch.Index) (weights map[epoch.ECR]map[identity.ID]float64) {
+	epochVotersWeightMutex.RLock()
+	defer epochVotersWeightMutex.RUnlock()
+	if _, ok := epochVotersWeight[ei]; !ok {
+		return
+	}
+
+	duplicate := make(map[epoch.ECR]map[identity.ID]float64, len(epochVotersWeight[ei]))
```
karimodm, line 284: You already have `weights` declared as your return variable. use make on that one

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	for k, v := range epochVotersWeight[ei] {
```
karimodm, line 285: Change the name of these variables to make it more readable.

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+		subDuplicate := make(map[identity.ID]float64, len(v))
+		for subK, subV := range v {
```
karimodm, line 287: Same as above.

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+			subDuplicate[subK] = subV
+		}
+		duplicate[k] = subDuplicate
+	}
+	return duplicate
+}
+
+func getEpochContentStorage(ei epoch.Index) (*epochContentStorages, error) {
+	if ei != epoch.Index(math.MaxInt64) && ei < minEpochIndex {
```
karimodm, line 296: It shouldn't be necessary to check for the first condition, the epoch.Index you receive is already an uint64 under the hood, and therefore its capacity is already bounded.

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+		return nil, errors.New("Epoch storage is no longer exists")
```
karimodm, line 297:
```suggestion
		return nil, errors.New("Epoch storage no longer exists")
```

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	messageIDs     kvstore.KVStore
+	transactionIDs kvstore.KVStore
+}
+
+type dependencies struct {
+	dig.In
+
+	Tangle          *tangle.Tangle
+	NotarizationMgr *notarization.Manager
+	Storage         kvstore.KVStore
```
karimodm, line 64: This Storage is injected but never used! You should use it for permament storage instead of the MemDB in `newEpochContentStorage`

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+		epochContentsMutex.Lock()
+		epochContents[ei] = stores
+		epochContentsMutex.Unlock()
+	}
+
+	return stores, nil
+}
+
+func newEpochContentStorage() *epochContentStorages {
+	db, _ := database.NewMemDB()
```
karimodm, line 315: You should use the injected `Storage` here and use a prefix to store stuff: `unique prefix for this plugin + the epoch index`. Like:
`	spentDiffStore, err := s.baseStore.WithRealm(append([]byte{database.PrefixEpochsStorage, spentOutputsPrefix}, ei.Bytes()...))`

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	// keep data in temporary storage
+	return &epochContentStorages{
+		spentOutputs:   db.NewStore(),
+		createdOutputs: db.NewStore(),
+		transactionIDs: db.NewStore(),
+		messageIDs:     db.NewStore(),
+	}
+}
+
+func getEpochTransactionIDs(ei epoch.Index) ([]utxo.TransactionID, error) {
```
karimodm, line 326: What is the difference with the exported method `GetEpochTransactions`?

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+		if _, err := txID.Decode(key); err != nil {
+			panic("TransactionID could not be parsed!")
+		}
+		transactionIDs = append(transactionIDs, txID)
+		return true
+	})
+	return transactionIDs, nil
+}
+
+func getEpochMessageIDs(ei epoch.Index) ([]tangle.MessageID, error) {
```
karimodm, line 344: You have an exported method with the same code.

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	if _, ok := epochVotersWeight[epochIndex]; !ok {
+		epochVotersWeight[epochIndex] = make(map[epoch.ECR]map[identity.ID]float64)
+	}
+	if _, ok := epochVotersWeight[epochIndex][message.M.ECR]; !ok {
+		epochVotersWeight[epochIndex][message.M.ECR] = make(map[identity.ID]float64)
+	}
+	epochVotersWeight[epochIndex][message.M.ECR][voter] = activeWeights[voter]
+}
+
+func outputIDFromBytes(outputBytes []byte) (utxo.OutputID, error) {
```
karimodm, line 468: This can be useful in general. Let's put it in utxo/types.go next to the other OutputID From functions

plugins/metrics/epochcommittment.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/iotaledger/hive.go/generics/event"
+
+	"github.com/iotaledger/goshimmer/packages/epoch"
+	"github.com/iotaledger/goshimmer/packages/notarization"
+)
+
+var (
+	lastCommittedEpoch      = epoch.NewECRecord(epoch.Index(0))
+	lastCommittedEpochMutex sync.RWMutex
```
karimodm, line 14: Can this be put in the epochstorage plugin as well? To have all the epochs stuff self-contained.

plugins/metrics/plugin.go
```diff
@@ -38,10 +39,11 @@ var (
...
 
-	Tangle    *tangle.Tangle
-	GossipMgr *gossip.Manager     `optional:"true"`
-	Selection *selection.Protocol `optional:"true"`
-	Local     *peer.Local
+	Tangle          *tangle.Tangle
+	GossipMgr       *gossip.Manager     `optional:"true"`
+	Selection       *selection.Protocol `optional:"true"`
+	Local           *peer.Local
+	NotarizationMgr *notarization.Manager
```
karimodm, line 46: Can this be put in the epochstorage plugin as well?

plugins/epochstorage/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+	for _, c := range created {
+		if err := epochContentStorage.createdOutputs.Delete(c.ID().Bytes()); err != nil {
+			return errors.New("Fail to remove created output from epoch store")
+		}
+	}
+
+	return nil
+}
+
+func saveEpochVotersWeight(message *tangle.Message) {
```
karimodm, line 452: If a voter already voted for another ECR of the same epoch we should delete the old entry! To determine which vote is the newest you will have to save the message IssuingTimestamp

karimodm (changes requested):
I am very impressed by the codebase, good quality! I just have a bunch of style comments and a couple more tricky ones!

plugins/epochstorage/plugin.go
jkrvivian: It's exactly the same, I'll remove it :sweat_smile:

plugins/metrics/epochcommittment.go
jkrvivian: The idea is to enable metrics plugin (grafana) without epochstorage plugin, so it's not necessary to store all the epoch contents unless we intended to.

karimodm (approved):
lgtm!

PR #2310, opened 2022-06-23
Fixes to epoch committments

karimodm:
# Description of change

* Do not use epoch delay when initializing aMana from snapshot
* Panic when errors are encountered during snapshot loading
* Drop committmentTrees upon committing an epoch: as far as I can tell this should not leak memory, as all the references to the underlying MemDB and Storages are now unrecheable (and therefore collected by the GC). What still leaks though is the map itself: as it never get shrank, even if entries are deleted.


packages/notarization/commitments.go
```diff
@@ -339,13 +339,18 @@ func (f *EpochCommitmentFactory) newEpochRoots(ei epoch.Index) (commitmentRoots
...
 		EI:                ei,
 		stateRoot:         epoch.NewMerkleRoot(stateRoot),
 		manaRoot:          epoch.NewMerkleRoot(manaRoot),
 		tangleRoot:        epoch.NewMerkleRoot(commitmentTrees.tangleTree.Root()),
 		stateMutationRoot: epoch.NewMerkleRoot(commitmentTrees.stateMutationTree.Root()),
-	}, nil
+	}
+
+	// We are never going to use this epoch's commitment trees again.
+	delete(f.commitmentTrees, ei)
```
jkrvivian, line 351: After this deletion, the proof of inclusion is not available anymore, should we keep the tree until pruning or epoch notarized, or even make it configurable?

karimodm: We had a few discussions about this, and we concluded that it is not the job of the core plugin to generate proofs, but rather have a specialized plugin that deals with such things by storing data for the required amount of time.
If you notice I've separated all the proofs related stuff in a separate file, and I would actually propose to drop it for the time being: still it will be nice to have some tests to actually create and verify some proofs of inclusion: mostly to test that the trees are fine rather than having a proofing / verifying framework.
What do you think @daria305 @jkrvivian ?

jkrvivian: I agree, then we can have a discussion about testing :tada:

daria305: We could add the proof-related things in form of a plugin, in some of next sprints, after we finish switching to heavier chains. But for now I agree we can limit to some simpe tests with the usage of proof functions.

plugins/messagelayer/mana_plugin.go
```diff
@@ -759,40 +759,49 @@ func QueryAllowed() (allowed bool) {
...
-				accessManaByNode[outputMetadata.AccessManaPledgeID()] -= manaValue
+				baseVector[outputMetadata.ConsensusManaPledgeID()] -= float64(balance)
 			}
 		}
 
 		return
 	}
 
-	processOutputs(snapshot.OutputsWithMetadata, true /* areCreated */)
+	processOutputs(snapshot.OutputsWithMetadata, consensusManaByNode, true /* areCreated */)
```
daria305, line 780: What are those comments for `/* areCreated */` , isn't this usually showed by the IDE?

daria305 (approved):

PR #2312, opened 2022-06-27
Trigger mana vector update by notarization manager

### LINKED ISSUE #2282: Attach mana plugin to EpochCommitted to recompute vector

daria305:
Fixes #2282 

Will be merged after: #2311
 - New event to detect committable epoch regardless if we committed or not to it (if the node is not emitting messages for instance).
 - The mana plugin will need to fetch UTXO Diff from the notarization manager corresponding to the latest committable epoch - configurable parameter (mana commit delay).
 - Update the vector applying the diff.

## Type of change

- added unit test


packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+	func() {
+		m.Lock()
+		defer m.Unlock()
+		// first, revoke mana from previous owners
+		for _, output := range spent {
+			idToRevoke := m.getIDBasedOnManaType(output)
+			outputIOTAs, _ := output.Output().(devnetvm.Output).Balances().Get(devnetvm.ColorIOTA)
+			oldMana := m.getOldManaAndRevoke(idToRevoke, float64(outputIOTAs))
+			// save events for later triggering
+			revokeEvents = append(revokeEvents, &RevokedEvent{idToRevoke, float64(outputIOTAs), output.OutputMetadata().CreationTime(), m.Type(), output.ID().TransactionID, output.ID()})
```
karimodm, line 181: Can you spell the fields out of these events one per line? It would be more redeable.

packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+	var pledgeEvents []*PledgedEvent
+	var updateEvents []*UpdatedEvent
+	// only lock mana vector while we are working with it
+	func() {
+		m.Lock()
+		defer m.Unlock()
+		// first, revoke mana from previous owners
+		for _, output := range spent {
+			idToRevoke := m.getIDBasedOnManaType(output)
+			outputIOTAs, _ := output.Output().(devnetvm.Output).Balances().Get(devnetvm.ColorIOTA)
```
karimodm, line 178: What if the color doesn't exist in the output?

packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+			updateEvents = append(updateEvents, &UpdatedEvent{idToRevoke, &oldMana, m.M.Vector[idToRevoke], m.Type()})
+
+		}
+		// second, pledge mana to new nodes
+		for _, output := range created {
+			idToPledge := m.getIDBasedOnManaType(output)
+			outputIOTAs, _ := output.Output().(devnetvm.Output).Balances().Get(devnetvm.ColorIOTA)
+			oldMana := m.getOldManaAndPledge(idToPledge, float64(outputIOTAs))
+			pledgeEvents = append(pledgeEvents, &PledgedEvent{idToPledge, float64(outputIOTAs), output.OutputMetadata().CreationTime(), m.Type(), output.Output().ID().TransactionID})
+			updateEvents = append(updateEvents, &UpdatedEvent{idToPledge, &oldMana, m.M.Vector[idToPledge], m.Type()})
```
karimodm, line 191: Same as above comments.

packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+		id = output.OutputMetadata().AccessManaPledgeID()
+	}
+	return id
+}
+
+func (m *ManaBaseVector) getOldManaAndRevoke(oldPledgeNodeID identity.ID, amount float64) ManaBase {
+	if _, exist := m.M.Vector[oldPledgeNodeID]; !exist {
+		// first time we see this node
+		m.M.Vector[oldPledgeNodeID] = &ManaBase{}
+		m.M.Vector[oldPledgeNodeID].Init()
```
karimodm, line 212: Use `NewManaBase`, it is more idiomatic. Also it won't be necessary to call `Init`

packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+	}
+	// save old mana
+	oldMana := *m.M.Vector[oldPledgeNodeID]
+	// revoke BM1
+	err := m.M.Vector[oldPledgeNodeID].revoke(amount)
+	if errors.Is(err, ErrBaseManaNegative) {
```
karimodm, line 218: Just checking: if it is the first time we saw this node this error will trigger right? I think it is ok that it produces a panic, it should never happen that you revoke some Mana you don't even know the existence of.

packages/mana/manabasevector.go
```diff
@@ -175,6 +164,75 @@ func (m *ManaBaseVector) Book(txInfo *TxInfo) {
...
+	var id identity.ID
+	if m.Type() == ConsensusMana {
+		id = output.OutputMetadata().ConsensusManaPledgeID()
+	} else {
+		id = output.OutputMetadata().AccessManaPledgeID()
+	}
+	return id
+}
+
+func (m *ManaBaseVector) getOldManaAndRevoke(oldPledgeNodeID identity.ID, amount float64) ManaBase {
```
karimodm, line 208: Can you add the oldMana as a named return parameter for this function and all the others (unless very trivial)? It makes it clearer what a function is supposed to return.

packages/mana/manabasevector_test.go
```diff
@@ -71,12 +76,49 @@ var (
...
+	addr := seed.NewSeed().Address(0).Address()
+	out := devnetvm.NewSigLockedSingleOutput(amount, addr)
+	meta := ledger.NewOutputMetadata(out.ID())
+	meta.SetConsensusManaPledgeID(createdPledgeID)
+	meta.SetCreationTime(now)
+	meta.SetID(out.ID())
+	outWithMeta := ledger.NewOutputWithMetadata(out.ID(), out, meta)
+	outWithMeta.SetID(out.ID())
+	outWithMeta.SetOutput(out)
+	outWithMeta.SetOutputMetadata(meta)
```
karimodm, line 118: These Set calls are not needed, NewOutputWithMetadata does them for you.

packages/notarization/manager.go
```diff
@@ -153,6 +145,16 @@ func (m *Manager) LoadSnapshot(snapshot *ledger.Snapshot) {
 		panic("could not set last confirmed epoch index")
 	}
 
+	// We set it to the last committed ei. It will be updated upon first confirmed message will arrive.
+	if err := m.epochCommitmentFactory.storage.SetLatestCommittableEpochIndex(snapshot.DiffEpochIndex); err != nil {
+		panic("could not set latest committable epoch index")
+	}
+
+	// We set it to the next epoch after snapshotted one. It will be updated upon first confirmed message will arrive.
+	if err := m.epochCommitmentFactory.storage.SetCurrentEpochIndex(snapshot.DiffEpochIndex + 1); err != nil {
```
karimodm, line 154: Why do we have to store this? We can always obtain the current Epoch index based on our clock. There is a `CurrentIndex` method in `epoch/types.go`

packages/notarization/manager.go
```diff
@@ -153,6 +145,16 @@ func (m *Manager) LoadSnapshot(snapshot *ledger.Snapshot) {
...
+		panic("could not set latest committable epoch index")
```
karimodm, line 155:
```suggestion
		panic("could not set current epoch index")
```

packages/notarization/manager.go
```diff
@@ -381,11 +384,25 @@ func (m *Manager) latestCommittableEpoch() (lastCommittedEpoch, latestCommittabl
 	return lastCommittedEpoch, latestCommittableEpoch, nil
 }
 
-// isCommittable returns if the epoch is committable, if all conflicts are resolved and the epoch is old enough.
+// IsCommittable returns if the epoch is committable, if all conflicts are resolved and the epoch is old enough.
```
karimodm, line 387:
```suggestion
// isCommittable returns if the epoch is committable, if all conflicts are resolved and the epoch is old enough.
```

packages/notarization/manager.go
```diff
@@ -381,11 +384,25 @@ func (m *Manager) latestCommittableEpoch() (lastCommittedEpoch, latestCommittabl
...
 func (m *Manager) isCommittable(ei epoch.Index) bool {
 	t := ei.EndTime()
-	diff := time.Since(t)
-	return m.pendingConflictsCounters[ei] == 0 && diff >= m.options.MinCommittableEpochAge
+	currentATT := m.tangle.TimeManager.ATT()
+	diff := currentATT.Sub(t)
+	if diff < m.options.MinCommittableEpochAge {
+		return false
+	}
+	latestEi, err := m.epochCommitmentFactory.storage.LatestCommittableEpochIndex()
```
karimodm, line 395: Should this be `LatestCommittedEpochIndex`?

packages/notarization/manager.go
```diff
@@ -469,12 +486,53 @@ func (m *Manager) outputsToOutputIDs(outputs devnetvm.Outputs) (createdIDs utxo.
...
 }
 
+// CheckIfEpochChanged check if next epoch started and trigger the event.
+func (m *Manager) CheckIfEpochChanged(issuingTime time.Time) {
+	currentTime := m.tangle.TimeManager.ATT()
+	if issuingTime.After(currentTime) {
+		currentTime = issuingTime
+	}
+	ei := epoch.IndexFromTime(currentTime)
+	currentEpochIndex, err := m.epochCommitmentFactory.storage.CurrentEpochIndex()
```
karimodm, line 496: Isn't this equivalent to checking the last epoch we committed to? So LastCommittedEpoch?

packages/notarization/manager.go
```diff
@@ -469,12 +486,53 @@ func (m *Manager) outputsToOutputIDs(outputs devnetvm.Outputs) (createdIDs utxo.
...
+		}
+		if !m.isCommittable(ei) {
+			return
+		}
+		if err = m.epochCommitmentFactory.storage.SetLatestCommittableEpochIndex(ei); err != nil {
+			m.log.Error(errors.Wrap(err, "could not set latest committable epoch index"))
+			return
+		}
+		m.triggerManaVectorUpdate(ei)
+		// todo move commitments trees to the latest committable epoch
```
karimodm, line 515: Is this still a TODO?

packages/notarization/storage.go
```diff
@@ -35,9 +35,9 @@ type EpochCommitmentStorage struct {
 	shutdownOnce sync.Once
 }
 
-type epochDiffStorage struct {
-	spent   *objectstorage.ObjectStorage[*ledger.OutputWithMetadata]
-	created *objectstorage.ObjectStorage[*ledger.OutputWithMetadata]
+type EpochDiffStorage struct {
+	Spent   *objectstorage.ObjectStorage[*ledger.OutputWithMetadata]
+	Created *objectstorage.ObjectStorage[*ledger.OutputWithMetadata]
```
karimodm, line 40: Do we use these outside the package?

packages/notarization/storage.go
```diff
@@ -78,22 +78,6 @@ func (s *EpochCommitmentStorage) CachedECRecord(ei epoch.Index, computeIfAbsentC
...
-}
-
-func (s *EpochCommitmentStorage) SetDiffEpochIndex(ei epoch.Index) error {
-	return s.setIndexFlag("diffEpochIndex", ei)
-}
-
-func (s *EpochCommitmentStorage) DiffEpochIndex() (ei epoch.Index, err error) {
-	return s.getIndexFlag("diffEpochIndex")
-}
-
```
karimodm, line 96: These two were used to know at what index we have our full LedgerState and up to what index we have the diffs: I although forgot to set these upon `commitLedgerState`! That's why you removed them?

karimodm (changes requested):
Thanks for the code! Looks great

packages/notarization/manager.go
daria305: To know that we moved to the next epoch, we need to see that index has increased

packages/notarization/manager.go
daria305: No, I forgot to remove it. At the beginning I had an idea to use this new epoch event to trigger update commitment roots, this way we will not do it on new message if commitment does not exist. I discarded it as what we have right now works fine

packages/notarization/manager.go
```diff
@@ -384,8 +387,22 @@ func (m *Manager) latestCommittableEpoch() (lastCommittedEpoch, latestCommittabl
...
+	diff := currentATT.Sub(t)
+	if diff < m.options.MinCommittableEpochAge {
+		return false
+	}
+	latestEi, err := m.epochCommitmentFactory.storage.LatestCommittableEpochIndex()
+	if err != nil {
+		return false
+	}
+	// epoch is not committable if there are any not resolved conflicts in this and past epochs
+	for index := latestEi; index <= ei; index++ {
```
jkrvivian, line 400: `latestEi` -> `latestEI`

jkrvivian (changes requested):

packages/mana/manabasevector.go
daria305: Not sure if it will always be the first time we see the nodeID as those are inputs. Previously we had panic but we booked tx by tx, so it should not happen that the balance will be negative. However with BookEpoch it should be checked only after all diffs are processed. I should change it

packages/notarization/storage.go
daria305: Not anymore, forgot to change it back after discarded idea

packages/notarization/manager.go
daria305: no, what if the node was not sending messages, then the last committed epoch stayed unchanged, as we update committments when we call `GetLatestEC()`. currentEpochIndex is advancing even if we do not create committments

packages/notarization/storage.go
daria305: I thought that they are not really used after loading the snapshot, so I removed full index and diff index. I assumed that when creating the snapshot we use lastCommittedEpochIndex and for diff we can simply retrieve it from parameter. We can add diff flag if it will be needed

karimodm (approved):

PR #2328, opened 2022-07-13
Keep epoch contents in permanent storage in epochstorage plugin

jkrvivian:
# Description of change

* Keep epoch contents in permanent storage
* Add `/ec` endpoint to fetch current ec

## Type of change

- Enhancement

## Change checklist

### DIFF
--- a/plugins/epochstorage/plugin.go
+++ b/plugins/epochstorage/plugin.go
@@ -32,6 +32,8 @@ var (
 	Plugin *node.Plugin
 	deps   = new(dependencies)
 
+	baseStore kvstore.KVStore
+
 	epochContentsMutex     sync.RWMutex
 	epochContents          = make(map[epoch.Index]*epochContentStorages, 0)
 	committableEpochsMutex sync.RWMutex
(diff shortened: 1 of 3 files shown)


plugins/epochstorage/plugin.go
```diff
@@ -296,99 +263,54 @@ func GetEpochVotersWeight(ei epoch.Index) (weights map[epoch.ECR]map[identity.ID
...
 
 func insertTransactionToEpoch(ei epoch.Index, txID utxo.TransactionID) error {
-	epochContentStorage, err := getEpochContentStorage(ei)
-	if err != nil {
-		return err
-	}
+	prefix := append([]byte{database.PrefixEpochsStorage, prefixTransactionIDs}, ei.Bytes()...)
 
-	if err := epochContentStorage.transactionIDs.Set(txID.Bytes(), txID.Bytes()); err != nil {
+	if err := baseStore.Set(append(prefix, txID.Bytes()...), txID.Bytes()); err != nil {
```
karimodm, line 287: I wouldn't append the prefix so directly. You can use `baseStore.WithRealm` to obtain a copy of the same underlying KVStore with the prefix set for you; so each object you operate on will always have that realm (prefix).

karimodm (changes requested):
Yeah, that works!!! I just left a comment to use the `WithRealm` instead of prepending the key of an object with its prefix every time.

karimodm (approved):

PR #2367, opened 2022-07-28
Warpsync: epoch syncing

karimodm:
# Description of change
Issue #2380
Implementation of the `Warpsync` plugin that will allow to retrieve entire epoch's content for a specific epoch commitment chain: for faster syncing and for switching to a heavier EC chain. 

## Type of change


packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+
+	epochBlocks := make(map[epoch.Index][]*blockReceived)
+	tangleRoots := make(map[epoch.Index]*smt.SparseMerkleTree)
+	epochSyncEnds := make(map[epoch.Index]*epochSyncEnd)
+
+	// We do not request the start nor the ending epoch, as we know the beginning and the end of the chain.
+	for ei := endEpoch - 1; ei > startEpoch; ei-- {
+		db, _ := database.NewMemDB()
+		tangleRoots[ei] = smt.NewSparseMerkleTree(db.NewStore(), db.NewStore(), lo.PanicOnErr(blake2b.New256(nil)))
+		m.requestEpochBlocks(ei, epoch.ComputeEC(ecChain[ei]))
```
jonastheis, line 66: Shouldn't this use the `concurrency` option? Also doing all in parallel seems like there might be **lots** of data coming in and stored in memory which could lead to OOM if there's lots of epochs to sync

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+			}
+
+			epochBlocksLeft[ei] = epochSyncStart.blocksCount
+			m.log.Debugw("read epoch block count", "EI", ei, "blockCount", epochBlocksLeft[ei])
+
+			epochsToReceive--
+			m.log.Debugf("epochs left %d", epochsToReceive)
+			if epochsToReceive == 0 {
+				break startReadLoop
+			}
```
jonastheis, line 101: Seems a bit odd to count the epochs instead of matching them explicitly. But even more so, we won't start processing anything unless we have received all `EpochStarter` packages. Imo it should try to process epochs independently of each other and operate more in a sliding window approach

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+
+			epochReceivedBlocks[ei][block.ID()] = types.Void
+			epochBlocksLeft[ei]--
+			m.log.Debugf("%d blocks left for epoch %d", epochBlocksLeft[ei], ei)
+			if epochBlocksLeft[ei] == 0 {
+				epochsToReceive--
+				m.log.Debugf("%d epochs left", epochsToReceive)
+				if epochsToReceive == 0 {
+					break
+				}
```
jonastheis, line 162: unnecessary since you check it at the beginning of the loop

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+				m.log.Warnf("received block on wrong EC chain")
+				continue
+			}
+			if _, exists := epochSyncEnds[ei]; exists {
+				m.log.Warnf("received block for already-ended epoch %d", ei)
+				continue
+			}
+			if epochBlocks[ei] == nil {
+				epochBlocks[ei] = make([]*blockReceived, 0)
+				epochReceivedBlocks[ei] = make(map[tangle.BlockID]types.Empty)
```
jonastheis, line 138: Why not just store `*blockReceived` in the map?

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+			}
+
+			block := epochSyncBlock.block
+			if _, exists := epochReceivedBlocks[ei][block.ID()]; exists {
+				continue
+			}
+
+			m.log.Debugw("read block", "peer", epochSyncBlock.peer.ID(), "EI", ei, "blockID", block.ID())
+
+			tangleRoots[ei].Update(block.IDBytes(), block.IDBytes())
```
jonastheis, line 148: iiuc: as it is currently where we request the epochs from all neighbors there's no way to tell whether a node sends us a crap block that doesn't belong in the commitment and thus fucks up the whole epoch. What if we request epoch by epoch from different neighbors?

jonastheis (changes requested):

packages/core/notarization/manager.go
```diff
@@ -100,6 +102,10 @@ func NewManager(epochCommitmentFactory *EpochCommitmentFactory, t *tangle.Tangle
 		new.OnAcceptanceTimeUpdated(event.ATT)
 	}))
 
+	new.tangle.Storage.Events.BlockStored.Attach(event.NewClosure(func(event *tangle.BlockStoredEvent) {
```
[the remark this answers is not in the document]
karimodm: I cannot really attach to the BlockMissing event as I don't have the block's data to determine what epochs should I try to warpsync.

packages/node/warpsync/manager.go
```diff
@@ -0,0 +1,173 @@
...
+	wpPacket := packet.(*wp.Packet)
+	switch packetBody := wpPacket.GetBody().(type) {
+	case *wp.Packet_EpochBlocksRequest:
+		if added := event.Loop.TrySubmit(func() { m.processEpochBlocksRequestPacket(packetBody, nbr) }); !added {
+			return fmt.Errorf("WorkerPool full: packet block discarded")
+		}
+	case *wp.Packet_EpochBlocksStart:
+		if added := event.Loop.TrySubmit(func() { m.processEpochBlocksStartPacket(packetBody, nbr) }); !added {
+			return fmt.Errorf("WorkerPool full: packet block discarded")
+		}
```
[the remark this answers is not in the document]
karimodm: Alright! For this to work I had to turn the submitTask to a generic to match the type of the callback with the packetBody, to avoid interface casting inside the packet consumer.

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,384 @@
...
+	for batchNum := 0; batchNum <= len(blockIDs)/m.blockBatchSize; batchNum++ {
+		blocksBytes := make([][]byte, 0)
+		for i := batchNum * m.blockBatchSize; i < len(blockIDs) && i < (batchNum+1)*m.blockBatchSize; i++ {
+			m.tangle.Storage.Block(blockIDs[i]).Consume(func(block *tangle.Block) {
+				blockBytes, err := block.Bytes()
+				if err != nil {
+					m.log.Errorf("failed to serialize block %s: %s", block.ID().String(), err)
+					return
+				}
+				blocksBytes = append(blocksBytes, blockBytes)
```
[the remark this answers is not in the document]
karimodm: How? I am creating a 2d array [][], I want to add the current block's []byte to the slice

packages/core/notarization/manager.go
```diff
@@ -701,4 +732,12 @@ type ManaVectorUpdateEvent struct {
 	EpochDiffSpent   []*ledger.OutputWithMetadata
 }
 
+// SyncRangeEvent is a container that acts as a dictionary for the SyncRange event related parameters.
+type SyncRangeEvent struct {
+	StartEI   epoch.Index
+	EndEI     epoch.Index
```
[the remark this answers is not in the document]
karimodm: Introduced a locking mechanism + check on end of last successfully warpsynced epoch.

packages/node/warpsync/validation.go
```diff
@@ -0,0 +1,190 @@
...
+				}
+
+				for peerID, epochCommitment := range neighborCommitmentsForEpoch {
+					if !validPeers.Has(peerID) {
+						continue
+					}
+					proposedECRecord := epochCommitment.ecRecord
+					if ecRecordChain[epochToValidate+1].PrevEC() != proposedECRecord.ComputeEC() {
+						m.log.Infof("ignoring commitment outside of the target chain", "peer", peerID)
+						validPeers.Delete(peerID)
```
jonastheis, line 100: What happens if a neighbor didn't send us any ECRecords? Is it still seen as a valid peer for the epoch sync?

packages/node/warpsync/validation.go
```diff
@@ -0,0 +1,190 @@
...
+						continue
+					}
+
+					// We store the valid committment for this chain.
+					ecRecordChain[epochToValidate] = proposedECRecord
+					ecChain[epochToValidate] = proposedECRecord.ComputeEC()
```
jonastheis, line 106: This is executed multiple times even though it doesn't change anymore later no?

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,327 @@
...
+
+	var wg sync.WaitGroup
+	epochProcessingChan := make(chan epoch.Index)
+	epochProcessingStopChan := make(chan struct{})
+	discardedPeers := set.NewAdvancedSet[identity.ID]()
+
+	// Spawn concurreny amount of goroutines.
+	for worker := 0; worker < m.concurrency; worker++ {
+		wg.Add(1)
+		go func() {
```
jonastheis, line 64: this should live inside it's own utility function to make it more readable

packages/node/warpsync/syncing.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+	}
+	close(epochProcessingStopChan)
+
+	wg.Wait()
+	close(epochProcessingChan)
+
+	if err != nil {
+		return errors.Errorf("sync failed for range %d-%d", start, end)
+	}
```
jonastheis, line 138: Instead of assigning the error in the callback I'd suggest the following: hand over a ctx that you can cancel to the goroutines and make them write into a result channel that contains an error and the completed EI (or maybe we don't even need that and counting is enough). Then you can cancel all goroutines on demand if there's an error and you wait on the result channel  instead of the epochProcessingChan

karimodm: Seems to be addressed now with the resultChan, cancellable errCtx and errorChan.

jonastheis (approved):

plugins/warpsync/parameters.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// ParametersDefinition contains the definition of configuration parameters used by the gossip plugin.
+type ParametersDefinition struct {
+	// Concurrency defines the amount of epochs to attempt to sync at the same time.
+	Concurrency int `default:"10" usage:"the amount of epochs to attempt to sync at the same time"`
+	// BlockBatchSize defines the amount of blocks to send in a single epoch blocks response"`
+	BlockBatchSize int `default:"100" usage:"the amount of blocks to send in a single epoch blocks response"`
+	// SyncRangeTimeOut defines the time after which a sync range is considered as failed.
+	SyncRangeTimeOut time.Duration `default:"30s" usage:"the time after which a sync range is considered as failed"`
```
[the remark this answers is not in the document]
karimodm: Increased it to 5m

Issue #2287, opened 2022-06-16
Store solid entrypoints


karimodm:
The solid entrypoints of an epoch can be determined on message arrival: the message's references that cross to an older epoch.

PR #2373, opened 2022-08-05
Implement solid entry points

jkrvivian:
# Description of change
#2287 

## Type of change

- Breaking change

## Change checklist


packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+// NewManager creates and returns a new snapshot manager.
+func NewManager(store kvstore.KVStore, t *tangleold.Tangle, nmgr *notarization.Manager) (new *Manager) {
+	new = &Manager{
+		tangle:          t,
+		notarizationMgr: nmgr,
+	}
+
+	new.baseStore = store
+
+	new.tangle.ConfirmationOracle.Events().BlockAccepted.Attach(event.NewClosure(func(e *tangleold.BlockAcceptedEvent) {
```
karimodm, line 38: Could we attach to these from the plugin? This is usually how events are wired up.

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/iotaledger/hive.go/core/kvstore"
+)
+
+const (
+	prefixSolidEntryPoint byte = iota
+)
+
+// Manager is the snapshot manager.
+type Manager struct {
+	tangle          *tangleold.Tangle
```
karimodm, line 22: We only use the tangle for its `CommittmentFunc`, don't we have it already through the notarizationMgr dependency?

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+func NewManager(store kvstore.KVStore, t *tangleold.Tangle, nmgr *notarization.Manager) (new *Manager) {
+	new = &Manager{
+		tangle:          t,
+		notarizationMgr: nmgr,
+	}
+
+	new.baseStore = store
+
+	new.tangle.ConfirmationOracle.Events().BlockAccepted.Attach(event.NewClosure(func(e *tangleold.BlockAcceptedEvent) {
+		e.Block.ForEachParent(func(parent tangleold.Parent) {
```
karimodm, line 39: We want to consider strong parents only.

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	m.notarizationMgr.WriteLockLedger()
+	defer m.notarizationMgr.WriteUnlockLedger()
+
+	// lock the entire solid entry points storage until the snapshot is created.
+	m.sepsLock.Lock()
+	defer m.sepsLock.Unlock()
+
+	headerPord := func() (header *ledger.SnapshotHeader, err error) {
+		header = &ledger.SnapshotHeader{
+			FullEpochIndex: lastConfirmedEpoch,
```
karimodm, line 71: Currently our snapshottable ledgerstate it is not fixed at our latestConfirmedEpoch, take a look at the `newEpochRoots`:
```
	// We advance the LedgerState to the next epoch.
	f.commitLedgerState(ei - epoch.Index(f.snapshotDepth))
```

This bit is invoked whenever we are able to advance our latest committable epoch.

karimodm: Until we have a confirmation gadget you have to make sure that FullEpochIndex here is the same one we have the Notarization manager's LedgerState on!

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+// CreateSnapshot creates a snapshot file from node with a given name.
+func (m *Manager) CreateSnapshot(snapshotFileName string) (header *ledger.SnapshotHeader, err error) {
+	ecRecord, lastConfirmedEpoch, err := m.tangle.Options.CommitmentFunc()
+	if err != nil {
+		return nil, err
+	}
+
+	// lock the entire ledger in notarization manager until the snapshot is created.
+	m.notarizationMgr.WriteLockLedger()
```
karimodm, line 62: We should lock before CommittmentFunc()

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, b := range seps.Seps {
+		m.insertSolidEntryPoint(b)
+	}
+}
+
+func (m *Manager) insertSolidEntryPoint(id tangleold.BlockID) error {
+	m.sepsLock.Lock()
+	defer m.sepsLock.Unlock()
+
+	sepsStore, err := m.baseStore.WithRealm(append([]byte{database.PrefixSnapshot, prefixSolidEntryPoint}, id.EpochIndex.Bytes()...))
```
karimodm, line 102: I've recently realized we have a more idiomatic way of doing this: `byteutils.ConcatBytes`

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+const (
+	prefixSolidEntryPoint byte = iota
+)
+
+// Manager is the snapshot manager.
+type Manager struct {
+	tangle          *tangleold.Tangle
+	notarizationMgr *notarization.Manager
+
+	baseStore kvstore.KVStore
```
karimodm, line 25: I don't think we should have a persistent storage for solid entrypoints: they are volatile and limited in nature, until we commit to an epoch. Let's use a ShrinkingMap instead.

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+	var blkID tangleold.BlockID
+	_, err = blkID.FromBytes(idBytes)
+	if err != nil {
+		return err
+	}
+
+	// cannot remove sep from confirmed epoch
+	if blkID.EpochIndex <= lastConfirmedEpoch {
+		return
```
karimodm, line 136: We should return an error.

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	return
+}
+
+func (m *Manager) SnapshotSolidEntryPoints(lastConfirmedEpoch, latestCommitableEpoch epoch.Index, prodChan chan *SolidEntryPoints) {
+	go func() {
+		for i := lastConfirmedEpoch; i <= latestCommitableEpoch; i++ {
+			sepsPrefix := append([]byte{database.PrefixSnapshot, prefixSolidEntryPoint}, i.Bytes()...)
+			seps := make([]tangleold.BlockID, 0)
+
+			m.baseStore.IterateKeys(sepsPrefix, func(key kvstore.Key) bool {
```
karimodm, line 153: We have to remove the storage in favor of a map, but anyways: this IterateKeys will return the `sepsPrefix` as part of the key in the callback! So you are going to have weird BlockIDs deserialized, as you are consuming the prefix as part of the blockID. For this kind of operation it is better to do the IterateKeys over a storage you specialized with `WithRealm`

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// lock the entire ledger in notarization manager until the snapshot is created.
+	m.notarizationMgr.WriteLockLedger()
+	defer m.notarizationMgr.WriteUnlockLedger()
+
+	// lock the entire solid entry points storage until the snapshot is created.
+	m.sepsLock.Lock()
+	defer m.sepsLock.Unlock()
+
+	headerPord := func() (header *ledger.SnapshotHeader, err error) {
```
karimodm, line 69:
```suggestion
	headerProd := func() (header *ledger.SnapshotHeader, err error) {
```

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	epochDiffsProd := NewEpochDiffsProducer(lastConfirmedEpoch, ecRecord.EI(), m.notarizationMgr)
+
+	header, err = CreateSnapshot(snapshotFileName, headerPord, sepsProd, outputWithMetadataProd, epochDiffsProd)
+
+	return
+}
+
+// LoadSolidEntryPoints add solid entry points to storage.
+func (m *Manager) LoadSolidEntryPoints(seps *SolidEntryPoints) {
+	if seps == nil {
```
karimodm, line 89: You should Lock

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	prefixSolidEntryPoint byte = iota
+)
+
+// Manager is the snapshot manager.
+type Manager struct {
+	tangle          *tangleold.Tangle
+	notarizationMgr *notarization.Manager
+
+	baseStore kvstore.KVStore
+	sepsLock  sync.RWMutex
```
karimodm, line 26: Can we embed the Lock?

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if err := sepsStore.Set(id.Bytes(), id.Bytes()); err != nil {
+		return errors.New("Fail to insert block to epoch store")
+	}
+
+	return nil
+}
+
+func (m *Manager) removeSolidEntryPoint(b *tangleold.Block, lastConfirmedEpoch epoch.Index) (err error) {
```
karimodm, line 114: Do we really need the lastConfirmedEpoch if we are deleting stuff anyway?

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+		panic(err)
+	}
+
+	idBytes, err := sepsStore.Get(b.ID().Bytes())
+	if err != nil || idBytes == nil {
+		return errors.New("solid entry point doesn't exist in storage or fail to fetch it")
+	}
+
+	var blkID tangleold.BlockID
+	_, err = blkID.FromBytes(idBytes)
```
karimodm, line 129: Make it oneline with the if

packages/core/snapshot/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+
+	// cannot remove sep from confirmed epoch
+	if blkID.EpochIndex <= lastConfirmedEpoch {
+		return
+	}
+
+	err = sepsStore.Delete(b.ID().Bytes())
```
karimodm, line 139: Make it oneline with the if

packages/core/snapshot/read.go
```diff
@@ -79,6 +88,21 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
 	return header, nil
 }
 
+func readSolidEntryPoints(scanner *bufio.Scanner) (seps *SolidEntryPoints, err error) {
+	scanner.Scan()
+	data := scanner.Bytes()
+
+	if len(data) > 0 {
+		seps = &SolidEntryPoints{}
+		_, err = serix.DefaultAPI.Decode(context.Background(), data, &seps, serix.WithValidation())
```
karimodm, line 97: Make it oneline with the if

packages/core/snapshot/read.go
```diff
@@ -38,14 +39,22 @@ func streamSnapshotDataFrom(
 	header.LatestECRecord = ecRecord
 	headerConsumer(header)
 
+	// read solid entry points
+	for i := header.FullEpochIndex; i <= header.DiffEpochIndex; i++ {
+		seps, err := readSolidEntryPoints(scanner)
```
karimodm, line 44: It would be better not to use delimiters but length fields instead for the whole snapshot.

packages/core/snapshot/write.go
```diff
@@ -109,6 +122,20 @@ func streamSnapshotDataTo(
 	return header, nil
 }
 
+// NewSolidEntryPointsProducer returns a SolidEntryPointsProducerFunc that provide solid entry points from the snapshot manager.
+func NewSolidEntryPointsProducer(lastConfirmedEpoch, latestCommitableEpoch epoch.Index, smgr *Manager) SolidEntryPointsProducerFunc {
+	prodChan := make(chan *SolidEntryPoints)
+	smgr.SnapshotSolidEntryPoints(lastConfirmedEpoch, latestCommitableEpoch, prodChan)
+
+	return func() *SolidEntryPoints {
+		obj, ok := <-prodChan
```
karimodm, line 131: I think we have a problem with this Producer: you are checking if there is data in the channel, if not you are returning nil. But since you are writing to the channel from a goroutine, there is not guarantee that there is data in it at any point in time when you are reading it; in which case you will only obtain partial data!
I've checked and the other "producers" that you have wrote as part of the streameable snapshot PR suffer from the same issue!
Can we simplify these producers? Do they really need a channel after all? They can maybe just be synchronous normal function calls without a goroutine.

plugins/blocklayer/plugin.go
```diff
@@ -166,8 +166,9 @@ func configure(plugin *node.Plugin) {
 		}
 
 		headerConsumer := func(*ledger.SnapshotHeader) {}
+		sepsConsumer := func(*snapshot.SolidEntryPoints) {}
```
karimodm, line 169: Maybe change the variable name to highlight that this is a nil / empty consumer

plugins/blocklayer/snapshot_plugin.go
```diff
@@ -0,0 +1,83 @@
...
+	SnapshotPlugin = node.NewPlugin(SnapshotPluginName, snapshotnDeps, node.Enabled, configureSnapshotPlugin, runSnapshotPlugin)
+
+	SnapshotPlugin.Events.Init.Hook(event.NewClosure(func(event *node.InitEvent) {
+		if err := event.Container.Provide(newSnapshotManager); err != nil {
+			SnapshotPlugin.Panic(err)
+		}
+	}))
+}
+
+func configureSnapshotPlugin(plugin *node.Plugin) {
```
karimodm, line 56: Events should be wired here.

karimodm (changes requested):

packages/core/snapshot/manager.go
jkrvivian: Lock is handled in insertion.

packages/core/snapshot/read.go
jkrvivian: Will fix that all in another PR.

packages/core/snapshot/write.go
jkrvivian: Hmm... I see your point, however at the time we actually get data from `prodChan`, it should be filled with data already.

packages/core/snapshot/manager.go
jkrvivian: There's a read lock in last confirmed and committable EI functions, so we can't put write lock before those. :thinking:

jkrvivian: I found a solution. :stuck_out_tongue:

karimodm (approved):

karimodm:
Merge #2381 on top before merging.

Issue #2372, opened 2022-08-04
Get rid of delimiter in snapshot


jkrvivian:
There's always a chance to break the data due to the serialized bytes are the same as the delimiter, instead, we should write/read the length of data to/from the snapshot.

PR #2381, opened 2022-08-12
Remove delimiter in snapshot

jkrvivian:
# Description of change

#2372 

## Type of change

 Breaking change

## Change checklist


packages/core/ledger/ledger.go
```diff
@@ -99,27 +99,20 @@ func (l *Ledger) LoadOutputWithMetadatas(outputsWithMetadatas []*OutputWithMetad
...
-		epochdiff, exists := epochDiffs[ei]
-		if !exists {
-			panic("epoch diff not found for epoch")
-		}
-
-		for _, spent := range epochdiff.Spent() {
-			l.Storage.outputStorage.Delete(spent.ID().Bytes())
-			l.Storage.outputMetadataStorage.Delete(spent.ID().Bytes())
-		}
+func (l *Ledger) LoadEpochDiffs(ei epoch.Index, epochDiff *EpochDiff) error {
```
karimodm, line 102: We are not using `ei` anymore right?

packages/core/notarization/manager.go
```diff
@@ -159,34 +159,32 @@ func (m *Manager) LoadOutputsWithMetadata(outputsWithMetadatas []*ledger.OutputW
 }
 
 // LoadEpochDiffs updates the state tree from a given snapshot.
-func (m *Manager) LoadEpochDiffs(header *ledger.SnapshotHeader, epochDiffs map[epoch.Index]*ledger.EpochDiff) {
+func (m *Manager) LoadEpochDiffs(ei epoch.Index, epochDiff *ledger.EpochDiff) {
```
karimodm, line 162: We can drop the `ei` parameter and rename the method to `LoadEpochDiff` (singular), to highlight that this now loads a single epoch's diff.

packages/core/ledger/ledger.go
karimodm, line 102:
```suggestion
func (l *Ledger) LoadEpochDiff(ei epoch.Index, epochDiff *EpochDiff) error {
```

packages/core/notarization/manager.go
```diff
@@ -214,15 +212,17 @@ func (m *Manager) LoadECandEIs(header *ledger.SnapshotHeader) {
...
-func (m *Manager) SnapshotEpochDiffs(fullEpochIndex, latestCommittableEpoch epoch.Index) (map[epoch.Index]*ledger.EpochDiff, error) {
-	// No need to lock because this is called in the context of a StartSnapshot.
-	epochDiffsMap := make(map[epoch.Index]*ledger.EpochDiff)
-	for ei := fullEpochIndex + 1; ei <= latestCommittableEpoch; ei++ {
-		spent, created := m.epochCommitmentFactory.loadDiffUTXOs(ei)
-		epochDiffsMap[ei] = ledger.NewEpochDiff(spent, created)
-	}
+func (m *Manager) SnapshotEpochDiffs(fullEpochIndex, latestCommitableEpoch epoch.Index, prodChan chan *ledger.EpochDiff, stopChan chan struct{}) {
+	go func() {
+		for i := fullEpochIndex; i <= latestCommitableEpoch; i++ {
```
karimodm, line 217:
```suggestion
		for ei := fullEpochIndex; ei <= latestCommitableEpoch; ei++ {
```

packages/core/snapshot/read.go
```diff
@@ -50,19 +40,22 @@ func streamSnapshotDataFrom(
...
 		}
 		i += len(outputs)
 		outputConsumer(outputs)
 	}
 
-	epochDiffs, err := readEpochDiffs(scanner)
-	if err != nil {
-		return errors.Errorf("failed to parse epochDiffs from bytes: %w", err)
+	// read epochDiffs
+	for i := header.FullEpochIndex + 1; i <= header.DiffEpochIndex; i++ {
```
karimodm, line 52:
```suggestion
	for ei := header.FullEpochIndex + 1; ei <= header.DiffEpochIndex; ei++ {
```

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	}
+	seps.EI = epoch.Index(index)
+
+	// read numbers of solid entry point
+	var sepsLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &sepsLen); err != nil {
+		return nil, fmt.Errorf("unable to read seps len: %w", err)
+	}
+
+	spentLenInt := int(sepsLen)
```
karimodm, line 115:
```suggestion
	sepsLenInt := int(sepsLen)
```

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	for i := 0; i < spentLenInt; {
```
karimodm, line 116: I don't understand why we need this for loop: don't we have a single solidEntryPoint set of blocksIDs per epoch index we read from the snapshot file? We should know ahead of time how many EI + SEPS combos we should read.

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	seps.Seps = blkIDs
+
 	return seps, nil
 }
 
 // readOutputWithMetadata consumes a slice of OutputWithMetadata from the given reader.
-func readOutputWithMetadata(scanner *bufio.Scanner) (outputMetadatas []*ledger.OutputWithMetadata, err error) {
-	scanner.Scan()
-	data := scanner.Bytes()
+func readOutputWithMetadata(reader io.ReadSeeker) (outputMetadatas []*ledger.OutputWithMetadata, err error) {
```
karimodm, line 142:
```suggestion
func readOutputsWithMetadata(reader io.ReadSeeker) (outputMetadatas []*ledger.OutputWithMetadata, err error) {
```

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	var spentLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &spentLen); err != nil {
+		return nil, fmt.Errorf("unable to read epochDiffs spent len: %w", err)
+	}
+	spentLenInt := int(spentLen)
+	for i := 0; i < spentLenInt; {
+		s, err := readOutputWithMetadata(reader)
 		if err != nil {
-			return nil, errors.Errorf("failed to parse epochDiffs from bytes: %w", err)
+			return nil, fmt.Errorf("unable to read epochDiffs spent: %w", err)
```
karimodm, line 181: use `errors.Errorf`

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
-	if len(data) > 0 {
-		_, err = serix.DefaultAPI.Decode(context.Background(), data, &epochDiffs, serix.WithValidation())
+func readEpochDiffs(reader io.ReadSeeker) (epochDiffs *ledger.EpochDiff, err error) {
+	spent := make([]*ledger.OutputWithMetadata, 0)
+	created := make([]*ledger.OutputWithMetadata, 0)
+
+	// read spent
+	var spentLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &spentLen); err != nil {
+		return nil, fmt.Errorf("unable to read epochDiffs spent len: %w", err)
```
karimodm, line 175: use `errors.Errorf`

packages/core/snapshot/read.go
karimodm, line 115: No need for an extra variable: cast it inline as part of the for condition

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	}
+	spentLenInt := int(spentLen)
```
karimodm, line 177: No need for an extra variable: cast it inline as part of the for condition

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
-			for _, createdOutput := range epochdiff.Created() {
-				createdOutput.SetID(createdOutput.M.OutputID)
-				createdOutput.Output().SetID(createdOutput.M.OutputID)
-			}
+	// read created
+	var createdLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &createdLen); err != nil {
+		return nil, fmt.Errorf("unable to read epochDiffs created len: %w", err)
+	}
+	createdLenInt := int(createdLen)
```
karimodm, line 192: No need for an extra variable: cast it inline as part of the for condition

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
+	created := make([]*ledger.OutputWithMetadata, 0)
+
+	// read spent
+	var spentLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &spentLen); err != nil {
+		return nil, fmt.Errorf("unable to read epochDiffs spent len: %w", err)
+	}
+	spentLenInt := int(spentLen)
+	for i := 0; i < spentLenInt; {
+		s, err := readOutputWithMetadata(reader)
```
karimodm, line 179: This method reads multiple outputs, but here you use it to consume only one.

packages/core/snapshot/read.go
karimodm: Is this method actually to consume multiple outputs? You seem to use it to consume single outputs.

packages/core/snapshot/read.go
```diff
@@ -85,68 +78,129 @@ func readSnapshotHeader(reader io.ReadSeeker) (*ledger.SnapshotHeader, error) {
...
-				createdOutput.Output().SetID(createdOutput.M.OutputID)
-			}
+	// read created
+	var createdLen int64
+	if err := binary.Read(reader, binary.LittleEndian, &createdLen); err != nil {
+		return nil, fmt.Errorf("unable to read epochDiffs created len: %w", err)
+	}
+	createdLenInt := int(createdLen)
+	for i := 0; i < createdLenInt; {
+		c, err := readOutputWithMetadata(reader)
```
karimodm, line 194: Why the for loop if this method returns multiple outputs?

packages/core/snapshot/write.go
```diff
@@ -158,26 +126,124 @@ func NewLedgerUTXOStatesProducer(nmgr *notarization.Manager) UTXOStatesProducerF
...
+		return err
+	}
+
+	s := diffs.Spent()
+	var end int
+	for i := 0; i < spentLen; {
+		if i+chunkSize > spentLen {
+			end = spentLen
+		} else {
+			end = i + chunkSize
```
karimodm, line 160: If you are let's say writing 300 outputs, with a chunksize of 100, this is not what you want!
Maybe something of this sort?
```
	for batchNum := 0; batchNum <= len(blockIDs)/m.blockBatchSize; batchNum++ {
		blocks := make([]*tangleold.Block, 0)
		for i := batchNum * m.blockBatchSize; i < len(blockIDs) && i < (batchNum+1)*m.blockBatchSize; i++ {
```

packages/core/snapshot/write.go
```diff
@@ -158,26 +126,124 @@ func NewLedgerUTXOStatesProducer(nmgr *notarization.Manager) UTXOStatesProducerF
...
+		}
+
+		i = end
+	}
+
+	return nil
+}
+
+func writeOutputsWithMetadata(writeSeeker io.WriteSeeker, outputsChunks []*ledger.OutputWithMetadata) error {
+	if len(outputsChunks) == 0 {
```
karimodm, line 229: Are these outputsChunks? This should be just one chunk of outputs no?

karimodm (changes requested):
I found the chunking a bit confusing to follow... it seems that you are trying to dump chunks of outputs but without prepending them with the chunk length, but just total length; and then using functions that would read several outputs to obtain just one.

packages/core/snapshot/write.go
jkrvivian: the example with 300 should work:
chunk 1: [0:100] (index 0 ~ index 99)
chunk 2: [100:200] (index 100 ~ index 199)
chunk 3: [200:300] (index 200~index 299)

`end` is assigned to `i` in each round

packages/core/snapshot/read.go
jkrvivian: `readOutputWithMetadata` reads the bytes length of a outputWithMetadata slice, deserialize them and return the slice.

packages/core/snapshot/read.go
jkrvivian: Here is to read blockIDs in chunks, we stop when we get the number of total seps (`sepsLenInt`) in the epoch. Doing this could avoid crashing the reader with a big slice of bytes at once, same goes for other data types.

The reading/writing process is:
1. write/read the total number of data types
2. split data into chunks (size = 100)
    a. write/read the bytes length of chunk
    b. write/read the data with fixed length, the reader will consume deserialized data 
    c. repeat a., b., until we hit the total number
3. move on to the next data type if any

packages/core/snapshot/write.go
jkrvivian: No, it writes a slice of OutputWithMetadata at once.

packages/core/snapshot/read.go
jkrvivian: No, it reads a slice of OutputsWithMetadata, I renamed it to readOutputsWithMetadata`s`

packages/core/snapshot/read.go
jkrvivian: OutputWithMetadatas are read/write in chunks, so every time it's read, it returns a slice.

karimodm (approved):

PR #2391, opened 2022-08-19
Implement simplified faucet

jkrvivian:
# Description of change

## Type of change

- Breaking change

## Change checklist


plugins/faucet/plugin.go
```diff
@@ -150,21 +122,13 @@ func run(plugin *node.Plugin) {
...
-		}
-		plugin.LogInfo("Deriving faucet state from the ledger... done")
-
-		defer fundingWorkerPool.Stop()
-		defer preparingWorkerPool.Stop()
-
+		_faucet.DeriveStateFromTangle()
 		initDone.Store(true)
 
+		go _faucet.Start(ctx, requestChan)
```
piotrm50, line 128: Is a new goroutine necessary here? In the next line we block on the ctx channel, and the goroutine waits for the same channel. Maybe it's not necessary to wait for the context here but simlpy make the `_faucet.Start` blocking?

plugins/faucet/plugin.go
```diff
@@ -150,21 +122,13 @@ func run(plugin *node.Plugin) {
...
-		if err := _faucet.DeriveStateFromTangle(ctx); err != nil {
-			plugin.LogErrorf("failed to derive state: %s", err)
-			return
-		}
-		plugin.LogInfo("Deriving faucet state from the ledger... done")
-
-		defer fundingWorkerPool.Stop()
-		defer preparingWorkerPool.Stop()
-
+		_faucet.DeriveStateFromTangle()
```
piotrm50, line 125: If we use `client/wallet` package then it automatically does that upon creation.

plugins/faucet/faucet.go
```diff
@@ -0,0 +1,344 @@
...
+			Plugin.LogInfof("send funds to %s: blockID: %s", p.Address().Base58(), blk.ID().Base58())
+
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// DeriveStateFromTangle derives unspent outputs of faucet from snapshot.
+func (f *Faucet) DeriveStateFromTangle() {
```
piotrm50, line 74: This shouldn't be necessary if we use `client/wallet`, because it initializes the state automatically.

piotrm50 (changes requested):
We should use `client/wallet` package to manage funds and addresses instead of having to manage it in the faucet plugin. It'd make the code more simple and readable. 
If we configure the wallet to reuse the same address, then it will always read the correct state from the tangle. 
It also would be nice to only handle requests that are not older than X minutes, because otherwise in case of faucet node restart, it would try to fullfill all the requests again.

plugins/faucet/faucet.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// remainder stays on index 0
+type Faucet struct {
+	*wallet.Wallet
+}
+
+// NewFaucet creates a new Faucet instance.
+func NewFaucet(faucetSeed *seed.Seed, walletStates string) *Faucet {
+	seed, lastAddressIndex, spentAddresses, assetRegistry, err := importWalletStateFile(walletStates)
```
piotrm50, line 32: You don't need to dump the wallet file. Wallet should always be initialized from seed.

plugins/faucet/parameters.go
```diff
@@ -11,6 +11,9 @@ type ParametersDefinition struct {
 	// Seed defines the base58 encoded seed the faucet uses.
 	Seed string `usage:"the base58 encoded seed of the faucet, must be defined if this faucet is enabled"`
 
+	// FaucetWalletFile defines the file name of faucet wallet states.
+	FaucetWalletFile string `default:"faucet.wallet" usage:"the states that is preserved for faucet wallet"`
```
piotrm50, line 15: We don't need to dump the wallet to a file as we always pass the seed.

plugins/faucet/plugin.go
```diff
@@ -38,16 +36,12 @@ const (
...
-	preparingWorkerQueueSize = MaxFaucetOutputsCount + 1
-	targetPoWDifficulty      int
+	Plugin              *node.Plugin
+	_faucet             *Faucet
+	powVerifier         = pow.New()
+	requestChanSize     = 100
+	requestChan         = make(chan *faucet.Payload, requestChanSize)
+	targetPoWDifficulty int
 	// blacklist makes sure that an address might only request tokens once.
 	blacklist         *orderedmap.OrderedMap[string, bool]
```
piotrm50, line 46: Remove the whole blacklist logic along with those fields as we don't need it and want to simplify as much as possilbe

piotrm50 (changes requested):

plugins/faucet/plugin.go
```diff
@@ -38,26 +35,20 @@ const (
...
+	targetPoWDifficulty int
+
 	// signals that the faucet has initialized itself and can start funding requests.
 	initDone     atomic.Bool
 	bootstrapped chan bool
 
-	waitForManaWindow = 5 * time.Second
-	deps              = new(dependencies)
+	waitForManaWindow   = 5 * time.Second
+	faucetRequestWindow = 3 * time.Minute
```
piotrm50, line 50: After all we might not need this because the faucet does not pledge tokens when it's not bootstrapped, which ultimately covers this check.

piotrm50 (changes requested):
Remove the faucet request window check, fix the tests and I think it's good to go :)

plugins/faucet/plugin.go
```diff
@@ -150,22 +105,12 @@ func run(plugin *node.Plugin) {
 		}
```
jonastheis, line 105: Can't comment on the actual line: do we really still need `waitForMana`? After the node is bootstrapped, can't we simply check if it has mana and if not we log an error and return?

plugins/faucet/connector.go
```diff
@@ -0,0 +1,126 @@
...
+		return errors.Errorf("%v: tx %s", err, tx.ID().String())
+	}
+	return nil
+}
+
+func (f *FaucetConnector) RequestFaucetFunds(address address.Address, powTarget int) (err error) {
+	return nil
+}
+
+func (f *FaucetConnector) GetAllowedPledgeIDs() (pledgeIDMap map[mana.Type][]string, err error) {
```
jonastheis, line 88: Not sure how this is internally used but it seems that there shouldn't be any allowed pledge IDs be configured on the faucet node and then the mana plugin will always just return the faucet nodeID in the allowed list. The pledges are still created correctly. So maybe this function can be simplified by just returning the current nodeID ?

jonastheis (changes requested):
It looks already pretty good and makes everything so much simpler. But I think some things can still be improved

jonastheis (approved):

piotrm50 (approved):

PR #2396, opened 2022-08-22
Mana vector fixes

daria305:
- Use `manaDelay` parameter only from mana plugin. Update mana vector event has now only EI.
- Adjust activity record during snapshotting: we keep only activity window for `lastConfirmed-activeEpochThreshold`


packages/core/tangleold/cmanaweightprovider.go
```diff
@@ -155,20 +157,25 @@ func (c *CManaWeightProvider) WeightsOfRelevantVoters() (weights map[identity.ID
...
 // SnapshotEpochActivity returns the activity log for snapshotting.
-func (c *CManaWeightProvider) SnapshotEpochActivity() (epochActivity epoch.SnapshotEpochActivity) {
+func (c *CManaWeightProvider) SnapshotEpochActivity(epochDiffIndex epoch.Index) (epochActivity epoch.SnapshotEpochActivity) {
 	epochActivity = epoch.NewSnapshotEpochActivity()
 
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
 	for ei, al := range c.activityLog {
 		al.SetEpochs.ForEach(func(nodeID identity.ID) error {
```
karimodm, line 174: Instead of having ActivityLog exposing an AdvanceSet, can we create a ForEach on the activity record itself while unexporting the AdvancedSet itself?

packages/core/tangleold/cmanaweightprovider.go
```diff
@@ -238,15 +248,15 @@ func (c *CManaWeightProvider) activityBoundaries() (lowerBoundEpoch, upperBoundE
 }
 
 // clean removes all activity logs for epochs lower than provided bound.
-func (c *CManaWeightProvider) clean(lowerBound epoch.Index) {
+func (c *CManaWeightProvider) clean(cutoffEpoch epoch.Index) {
 	for ei := range c.activityLog {
-		if ei < lowerBound {
+		if ei < cutoffEpoch {
 			delete(c.activityLog, ei)
```
karimodm, line 254: c.activityLog should be turned into a ShrinkingMap so we don't leak memory in the map.

packages/core/epoch/types.go
```diff
@@ -246,38 +249,42 @@ func (al *NodesActivityLog) Bytes() []byte {
...
+
+// endregi`azon ///////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 // region ActivityLog //////////////////////////////////////////////////////////////////////////////////////////////////
 
 // ActivityLog is a time-based log of node activity. It stores information when a node is active and provides
 // functionality to query for certain timeframes.
 type ActivityLog struct {
-	SetEpochs *set.AdvancedSet[identity.ID] `serix:"0,lengthPrefixType=uint32"`
+	setEpochs *set.AdvancedSet[identity.ID] `serix:"0,lengthPrefixType=uint32"`
```
karimodm, line 263: Can we embed this set directly by removing its name as it is the only field? So functions such as ForEach, Add and so on will be exposed on ActivityLog itself, without the need to rewrite them.

karimodm (changes requested):

packages/core/tangleold/cmanaweightprovider.go
daria305: Now I use ForEach, but it turned out that I cannot unexport `SetEpochs` as the serialization stops working

packages/core/epoch/types.go
```diff
@@ -304,6 +351,22 @@ func (a *ActivityLog) Clone() *ActivityLog {
 	return clone
 }
 
+// ForEach iterates through the activity set and calls the callback for every element.
+func (a *ActivityLog) ForEach(callback func(nodeID identity.ID) (err error)) (err error) {
+	a.SetEpochs.ForEach(func(nodeID identity.ID) (err error) {
```
karimodm, line 363: You could just do `	return a.SetEpochs.ForEach(callback)`

packages/core/epoch/types.go
```diff
@@ -304,6 +351,22 @@ func (a *ActivityLog) Clone() *ActivityLog {
...
+		if err = callback(nodeID); err != nil {
+			return err
+		}
+		return nil
+	})
+
+	return err
+}
+
+func (a *ActivityLog) Size() int {
```
karimodm, line 373: missing comment

karimodm (approved):

PR #2405, opened 2022-08-27
Feat: Finalize CausalOrder

hmoog:
# Description of change

This PR finalizes the CausalOrder prior to moving it to hive.go.

## Type of change

## Change checklist


packages/core/causalorder/testframework.go
```diff
@@ -2,35 +2,118 @@ package causalorder
...
+					t.test.Logf("%s ordered", entity.id.alias)
+				}
+
+				t.orderedEntitiesMutex.Lock()
+				t.orderedEntities[entity.id.alias] = true
+				t.orderedEntitiesMutex.Unlock()
+
+				return nil
+			}, func(entity *MockedOrderedEntity, reason error) {
+				entity.invalid = true
```
jonastheis, line 62: Can this be called from different goroutines? If yes, then we might need a setter/getter for the `MockedOrderedEntity` properties

packages/core/causalorder/testframework.go
```diff
@@ -55,67 +138,110 @@ func (t *TestFramework) EntityIDs(aliases ...string) (entityIDs []MockEntityID)
...
+// AssertOrdered asserts that the given Entities are ordered.
+func (t *TestFramework) AssertOrdered(aliases ...string) {
+	t.orderedEntitiesMutex.RLock()
+	defer t.orderedEntitiesMutex.RUnlock()
+
+	require.Equal(t.test, len(aliases), len(t.orderedEntities))
+
+	for _, alias := range aliases {
+		require.True(t.test, t.orderedEntities[alias])
+	}
```
jonastheis, line 156: Could we also get the entities here and check whether the ordered flag is set?

packages/core/causalorder/testframework.go
```diff
@@ -55,67 +138,110 @@ func (t *TestFramework) EntityIDs(aliases ...string) (entityIDs []MockEntityID)
...
+// AssertEvicted asserts that the given Entities are evicted.
+func (t *TestFramework) AssertEvicted(aliases ...string) {
+	t.evictedEntitiesMutex.RLock()
+	defer t.evictedEntitiesMutex.RUnlock()
+
+	require.Equal(t.test, len(aliases), len(t.evictedEntities))
+
+	for _, alias := range aliases {
+		require.True(t.test, t.evictedEntities[alias])
+	}
```
jonastheis, line 168: Would be nice to not evict from the entity provider (within the testframework) and retrieve the entity and check whether the corresponding flag is set. ie when not solid and evicted -> orphaned/invalid right? so then this check would validate in more depth and also whether the corresponding callback was executed correctly

jonastheis (approved):

PR #2412, opened 2022-09-01
Fix: several bugs in Notarization and Networking

karimodm:
# Description of change

Fixes:
- epochProcessBlockCommand: index out of range [0] with length 0.
- p2p concurrent map write.
- objectstoreage creation leak from mana_plugin.
- tentative fix: readloop still seems to consume too much CPU (maybe still in an endless loop?).
- implemented shrinkingnmaps in notarization manager.


jonastheis (approved):
nice bunch of fixes

PR #2417, opened 2022-09-05
Fix notarization and networking bugs

karimodm:
# Description of change

This PR fixes a few bugs:
- Notarization's storeDiffUTXOs output included multiple times in epoch.
- "send on closed channel" when handling a closed stream.

## Type of change

### DIFF
--- a/packages/node/p2p/stream.go
+++ b/packages/node/p2p/stream.go
@@ -283,6 +283,8 @@ func (m *Manager) removeAcceptMatcher(am *AcceptMatcher, protocolID protocol.ID)
 	existingAm.StreamChMutex.Lock()
 	defer existingAm.StreamChMutex.Unlock()
 
+	delete(existingAm.StreamCh, protocolID)
+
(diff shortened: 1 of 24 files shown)


packages/node/p2p/stream.go
```diff
@@ -210,8 +210,14 @@ func (m *Manager) handleStream(stream network.Stream) {
...
 		am.StreamChMutex.RLock()
+		defer am.StreamChMutex.RUnlock()
 		streamCh := am.StreamCh[protocolID]
-		am.StreamChMutex.RUnlock()
+
+		select {
+		case <-am.Ctx.Done():
+			return
+		default:
+		}
```
jonastheis, line 220: Are we 100% sure this can't lead to a deadlock? because it is waiting on the channel while the `RLock` is still being held

karimodm: I was indeed adressing that =)

jonastheis (approved):

PR #2420, opened 2022-09-06
Prunable scheduler

piotrm50:
# Description of change

Copied existing scheduler logic and made it compatible with the current pruning approach.

## Type of change

- Documentation Fix

## Change checklist


packages/protocol/engine/congestioncontrol/congestioncontrol.go
```diff
@@ -1,14 +1,31 @@
...
 type CongestionControl struct {
+	Events *Events
+
+	optsScheduler []options.Option[scheduler.Scheduler]
+
+	*scheduler.Scheduler
 }
 
-func New(opts ...options.Option[CongestionControl]) (congestionControl *CongestionControl) {
+func New(isBlockAccepted func(models.BlockID) bool, blockAcceptedEvent *event.Linkable[*acceptance.Block, acceptance.Events, *acceptance.Events], tangle *tangle.Tangle, accessManaMapRetrieverFunc func() map[identity.ID]float64, totalAccessManaRetrieveFunc func() float64, rate time.Duration, opts ...options.Option[CongestionControl]) (congestionControl *CongestionControl) {
```
jonastheis: That's a lot to set up. But I suppose this way it's independent of acceptance gadget etc and easier testable? the rate could be an option with default value no? otherwise we'd need to make it a mandatory parameter also to the engine and protocol. I was also always wondering why it is given in `time.Duration`. This seems confusing as we usually talk about BPS. Could we change this as well? Should only be a simple conversion

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
```diff
@@ -0,0 +1,309 @@
...
+	if !ok {
+		return nil
+	}
+	return element.Value.(*IssuerQueue)
+}
+
+// Submit submits a block. Return blocks dropped from the scheduler to make room for the submitted block.
+// The submitted block can also be returned as dropped if the issuing node does not have enough access mana.
+func (b *BufferQueue) Submit(blk *Block, accessManaRetriever func(identity.ID) float64) (elements []*Block, err error) {
+	nodeID := identity.NewID(blk.IssuerPublicKey())
```
jonastheis: `block.IssuerID()`

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
```diff
@@ -0,0 +1,309 @@
...
+			panic("scheduler buffer size exceeded and the longest scheduler queue is empty.")
+		}
+	}
+	return droppedBlocks
+}
+
+// Unsubmit removes a block from the submitted blocks.
+// If that block is already marked as ready, Unsubmit has no effect.
+func (b *BufferQueue) Unsubmit(block *Block) bool {
+	issuerID := identity.NewID(block.IssuerPublicKey())
```
jonastheis: replace this everywhere with `block.IssuerID()`

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
```diff
@@ -0,0 +1,309 @@
...
+		q = b.Next()
+		if q == start {
+			break
+		}
+	}
+	return
+}
+
+// InsertIssuer creates a queue for the given node and adds it to the list of active nodes.
+func (b *BufferQueue) InsertIssuer(nodeID identity.ID) {
```
jonastheis: Maybe you can also do a rename from `nodeID` to `issuerID`  everywhere to be more aligned with the new terminology

packages/protocol/engine/engine.go
```diff
@@ -28,21 +29,30 @@ type Engine struct {
...
+	optsSchedulerRate         time.Duration
 	optsTangle                []options.Option[tangle.Tangle]
 	optsConsensus             []options.Option[consensus.Consensus]
+	optsCongestionControl     []options.Option[congestioncontrol.CongestionControl]
 }
 
 func New(snapshotTime time.Time, ledger *ledger.Ledger, evictionManager *eviction.Manager[models.BlockID], validatorSet *validator.Set, opts ...options.Option[Engine]) (engine *Engine) {
 	return options.Apply(&Engine{
 		optsBootstrappedThreshold: 10 * time.Second,
+		optsSchedulerRate:         5 * time.Millisecond,
```
jonastheis: Why not have this in the CongestionControl as a default parameter?

packages/protocol/engine/congestioncontrol/icca/scheduler/testframework.go
```diff
@@ -0,0 +1,278 @@
...
+		if !exists {
+			delete(t.issuersMana, identity.ID())
+		}
+	}
+}
+
+func (t *TestFramework) Issuer(alias string) (issuerIdentity *identity.Identity) {
+	issuerIdentity, exists := t.issuersByAlias[alias]
+	if !exists {
+		panic("identity aliast not registered")
```
jonastheis:
```suggestion
		panic("identity alias not registered")
```

packages/protocol/engine/congestioncontrol/icca/scheduler/issuerqueue.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/iotaledger/goshimmer/packages/protocol/engine/tangle/models"
+)
+
+// region IssuerQueue /////////////////////////////////////////////////////////////////////////////////////////////
+
+// IssuerQueue keeps the submitted blocks of a issuer.
+type IssuerQueue struct {
+	issuerID  identity.ID
+	submitted map[models.BlockID]*Block
```
jonastheis: This map could potentially become very large as these objects are long-lived no? Might be better to switch it already for a shrinkingmap

packages/protocol/engine/congestioncontrol/icca/scheduler/issuerqueue.go
```diff
@@ -0,0 +1,148 @@
...
+	q.size.Dec()
+	return blk
+}
+
+// endregion ///////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// region ElementHeap /////////////////////////////////////////////////////////////////////////////////////////////
+
+// ElementHeap holds a heap of blocks with respect to their IssuingTime.
+type ElementHeap []*Block
```
jonastheis, line 119: It's basically the same heap we need to use for TSC but with a different type. Can we somehow avoid the duplicate code?

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -0,0 +1,610 @@
...
+	Tangle *tangle.Tangle
+
+	EvictionManager *eviction.LockableManager[models.BlockID]
+
+	blocks        *memstorage.EpochStorage[models.BlockID, *Block]
+	ticker        *time.Ticker
+	bufferMutex   sync.RWMutex
+	buffer        *BufferQueue
+	deficitsMutex sync.RWMutex
+	deficits      map[identity.ID]*big.Rat
```
jonastheis: ShrinkingMap?

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	s.markAsDropped(droppedBlocks)
+
+	return nil
+}
+
+func (s *Scheduler) markAsDropped(droppedBlocks []*Block) {
+	for _, droppedBlock := range droppedBlocks {
+		s.Tangle.SetOrphaned(droppedBlock.Block.Block.Block, true)
```
jonastheis: Should a block become orphaned immediately if we drop it? Shouldn't we better wait for the TSC threshold to kick in?

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -0,0 +1,610 @@
...
+
+}
+
+func (s *Scheduler) setupEvents() {
+	// pass booked blocks to the scheduler
+	s.Tangle.VirtualVoting.Events.BlockTracked.Attach(event.NewClosure(s.AddBlock))
+
+	s.Events.BlockScheduled.Hook(event.NewClosure(s.UpdateChildren))
+
+	s.blockAcceptedEvent.Attach(event.NewClosure(s.HandleAcceptedBlock))
```
jonastheis: Handing around this event seems to be a bit awkward. Does it help if we attach the inter-component events on the CongestionControl component (and hand the acceptance gadget to it)?

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -0,0 +1,610 @@
...
+
+	if err := s.Submit(block); err != nil {
+		if !errors.Is(err, ErrInsufficientMana) {
+			s.Events.Error.Trigger(errors.Wrap(err, "failed to submit to scheduler"))
+		}
+	}
+	s.tryReady(block)
+}
+
+func (s *Scheduler) HandleAcceptedBlock(acceptedBlock *acceptance.Block) {
```
jonastheis: Here we still have a dependency to `acceptance`. Should better receive a `virtualvoting.Block`

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -0,0 +1,610 @@
...
+			break loop
+		}
+	}
+}
+
+func (s *Scheduler) schedule() *Block {
+	s.bufferMutex.Lock()
+	defer s.bufferMutex.Unlock()
+	s.EvictionManager.RLock()
+	defer s.EvictionManager.RUnlock()
```
jonastheis, line 441: these two locks seem to be highly correlated. it's a bit of a mess to follow the code and the locking tbh

jonastheis (changes requested):

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
piotrm50: :+1:

packages/protocol/engine/congestioncontrol/icca/scheduler/issuerqueue.go
piotrm50: We can do that when we merge both PRs. Now we'd have to copy it individually and solve potential conflicts.

packages/protocol/engine/congestioncontrol/icca/scheduler/issuerqueue.go
piotrm50: done

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: That's true, but I don't think we only can have one. Eviction lock is used whenever we execute a publicly facing method that accesses block storage or `schedule` method so that eviction is not performed in the middle. 
Buffer mutex is used by any method that accesses or modifies the block, so it's true that they often go together, but not always.
Also eviction is mostly read locked while buffer is often write locked.

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: done

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: This dependency is not bad, as we can mock this block. The main motivation for not passing acceptance.Gadget to scheduler was that it would be hard to mock in tests

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: That could work - then we'd need to also manually attach it in the test framework, but that's definitely less awkward than passing an event instance around :D

piotrm50: done

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: Done, attached to the event in congestion control and unpacked `virtualvoting.Block`

packages/protocol/engine/engine.go
piotrm50: Done

packages/protocol/engine/congestioncontrol/congestioncontrol.go
piotrm50: > This seems confusing as we usually talk about BPS. Could we change this as well?

I talked to Andrew about this and this will need to be converted to Work Units Per Second ultimately. Let's leave it as it is for now as not to change it twice.

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
piotrm50: done

packages/protocol/engine/congestioncontrol/icca/scheduler/bufferqueue.go
piotrm50: done

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: I changed how orphaned and dropped blocks are handled by the scheduler. Scheduler no longer will mark blocks as orphaned, so take a look at how it's done now.

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -233,6 +230,14 @@ func (s *Scheduler) AddBlock(sourceBlock *virtualvoting.Block) {
 
 	block, _ := s.getOrRegisterBlock(sourceBlock)
 
+	if block.IsOrphaned() {
+		if !block.IsDropped() {
```
jonastheis: if you returned an `updated` from `SetDropped` you could inline this here

jonastheis (approved):

packages/protocol/engine/congestioncontrol/icca/scheduler/scheduler.go
piotrm50: 👍

PR #2429, opened 2022-09-12
Feat: started implementing CommitmentManager

hmoog:
# Description of change

This PR introduces a CommitmentManager that maintains an in-memory representation of the commitments and their formed chains and allows to determine a chain identifier for each epoch commitment.

This will allow us to dispatch received Blocks, to the correct protocol instance.

## Type of change

## Change checklist


packages/core/epoch/commitmentmanager/commitmentmanager.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/iotaledger/hive.go/core/generics/walker"
+
+	"github.com/iotaledger/goshimmer/packages/core/epoch"
+)
+
+type CommitmentManager struct {
+	Events             *Events
+	SnapshotCommitment *Commitment
+
+	commitmentsByEC map[epoch.EC]*Commitment
```
piotrm50, line 16: I found it a little confusing at first - "commitments by Epoch Commitment" sounds a little weird imo. Should we denote somehow when we use hash (merkle root) and when the contents of a commitment?

packages/core/epoch/commitmentmanager/commitmentmanager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	chain.addCommitment(child)
+
+	children := child.Children()
+	if len(children) == 0 {
+		return
+	}
+
+	return []*Commitment{children[0]}
```
piotrm50: Can we return a single value instead of a slice?

packages/core/epoch/commitmentmanager/testframework.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func NewTestFramework(test *testing.T, opts ...options.Option[TestFramework]) (testFramework *TestFramework) {
+	return options.Apply(&TestFramework{
+		ChainManager: New(0, epoch.ECR{0}, epoch.EC{0}),
+
+		test:               test,
+		commitmentsByAlias: make(map[string]*Commitment),
+	}, opts, func(t *TestFramework) {
+		t.commitmentsByAlias["genesis"] = t.ChainManager.SnapshotCommitment
```
piotrm50: Let's call it "Genesis" with capital G to be consistent with the tangle testframeworks?

packages/core/epoch/commitmentmanager/commitmentmanager.go
hmoog: Not really, then we can not unpack the result of that function in the PushAll of the walker and need extra code there.

But we could change it to return return children[:1]

Originally we always traversed all children in the walks but here we only push the chain identifier to the first child.

packages/core/epoch/commitmentmanager/commitmentmanager.go
hmoog: Actually I also think the naming is a bit weird (already in the epoch package).

For me the Commitment is the object including its properties (also the way we express it in the block) and the MerkleRoot would be sth like a CommitmentID.

But in the epoch package the EC is the ID and the full commitment is called ECRecord.

packages/core/epoch/commitmentmanager/testframework.go
hmoog: Done

piotrm50 (approved):

PR #2434, opened 2022-09-15
Refactor block creation

jonastheis:
# Description of change

This PR refactors the block creation which is split in two separate parts:
- block factory to create a block
- reference provider to provide references/votes based on the node's current opinion


packages/app/blockfactory/referenceprovider.go
```diff
@@ -0,0 +1,217 @@
...
+
+				committableEpoch := r.latestEpochIndexCallback()
+				if latestAttachment.ID().Index() <= committableEpoch {
+					continue
+				}
+
+				references.Add(models.WeakParentType, latestAttachment.ID())
+			}
+		}
+	}
```
piotrm50: Make this an utility function

PR #2436, opened 2022-09-20
DB Manager restoring from disk and LRU open DB instances

jonastheis:
This PR introduces a few things:
- pruning of the DB instances
- LRU cache of open DB instances, only a configurable max amount can be open
- restore DB manager from disk: gather information about epochs stored on disk and delete unhealthy epochs, returning the highest healthy epoch


packages/protocol/database/manager.go
```diff
@@ -95,16 +228,22 @@ func (m *Manager) Shutdown() {
...
 		db = m.createDBInstance(startingIndex)
 		m.dbs.Set(startingIndex, db)
 	}
+	db.lastAccessed = time.Now()
 
 	return db
 }
 
+func (m *Manager) computeDBBaseIndex(index epoch.Index) epoch.Index {
+	startingIndex := index / epoch.Index(m.optsGranularity) * epoch.Index(m.optsGranularity)
```
piotrm50: Return the value without assiging to a variable

packages/protocol/database/manager.go
```diff
@@ -95,16 +228,22 @@ func (m *Manager) Shutdown() {
...
 	if !exists {
 		db = m.createDBInstance(startingIndex)
 		m.dbs.Set(startingIndex, db)
 	}
+	db.lastAccessed = time.Now()
 
 	return db
 }
 
+func (m *Manager) computeDBBaseIndex(index epoch.Index) epoch.Index {
```
piotrm50: Should epoch.Index be used in the context of DB index? It could be slightly confusing

jonastheis: What else do you suggest?

piotrm50: Since it's an alias for `int64` maybe just create a new alias e.g. `DBBaseIndex`.

jonastheis: I tried to use it with this type but turned out to be awkward because lots of type conversions and we still need to handle `epoch.Index` in many places. Also the DBBaseIndex is technically only `index / epoch.Index(m.optsGranularity) * epoch.Index(m.optsGranularity)`. So for now I'll leave it with `epoch.Index`.

hmoog (approved):
LGBTQM

PR #2447, opened 2022-10-27
Epoch and Block Confirmation gadget

alexsporn:
Introduces Epoch and Block confirmation. Wires up dependencies including database pruning.


packages/core/eviction/state.go
```diff
@@ -23,8 +23,9 @@ func NewState[ID epoch.IndexedID]() (newState *State[ID]) {
 	var emptyID ID
 
 	return &State[ID]{
-		Events:          NewEvents(),
-		rootBlocks:      set.NewAdvancedSet(emptyID),
+		Events:     NewEvents(),
+		rootBlocks: set.NewAdvancedSet(emptyID),
+		// TODO: set max evicted epoch to the epoch from the snapshot or database when initializing
```
jonastheis, line 28: I suppose this needs to be done after #2457 ?

packages/protocol/engine/sybilprotection/activitytracker/activitytracker.go
```diff
@@ -50,14 +49,14 @@ func (a *ActivityTracker) Update(activeValidator *validator.Validator, activityT
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
 
-	fmt.Printf("Update activity %s: %s\n", activeValidator.ID(), activityTime)
+	//fmt.Printf("Update activity %s: %s\n", activeValidator.ID(), activityTime)
```
jonastheis: Maybe remove?

packages/core/votes/epochtracker/epochtracker.go
```diff
@@ -0,0 +1,137 @@
...
+		// We already tracked the voter for this epoch, so no need to update anything
+		return
+	}
+
+	votersVotes, _ := c.votesPerIdentity.RetrieveOrCreate(voterID, func() *latestvotes.LatestVotes[epoch.Index, EpochVotePower] {
+		return latestvotes.NewLatestVotes[epoch.Index, EpochVotePower](voter)
+	})
+
+	updated, previousHighestIndex := votersVotes.Store(epochIndex, power)
+	if !updated || previousHighestIndex == epochIndex {
```
jonastheis: Shouldn't it be `previousHighestIndex <= epochIndex`?

packages/protocol/engine/consensus/consensus.go
```diff
@@ -14,21 +16,25 @@ import (
 type Consensus struct {
 	Events *Events
 
-	*acceptance.Gadget
+	AcceptanceGadget        *acceptance.Gadget
```
jonastheis: I think the name `AcceptanceGadget` and package `acceptance` is not fitting anymore. It is actually misleading. I'm not sure what we should call it but now it suggest that only acceptance is handled within but also block confirmation is part of it.

packages/protocol/engine/consensus/acceptance/gadget.go
```diff
@@ -120,15 +165,33 @@ func (a *Gadget) RefreshSequenceAcceptance(sequenceID markers.SequenceID, newMax
 		marker := markers.NewMarker(sequenceID, markerIndex)
 
 		markerVoters := a.tangle.VirtualVoting.MarkerVoters(marker)
-		if a.tangle.ValidatorSet.IsThresholdReached(markerVoters.TotalWeight(), a.optsMarkerAcceptanceThreshold) && a.setMarkerAccepted(marker) {
-			queuedBlocks = append(queuedBlocks, a.propagateAcceptance(marker)...)
+
+		if validator.IsThresholdReached(totalWeight, a.tangle.ValidatorSet.TotalWeight(), a.optsConfirmationThreshold) {
+			// have enough weight to confirm based on total weight
+			if validator.IsThresholdReached(totalWeight, markerVoters.TotalWeight(), a.optsConfirmationThreshold) {
```
jonastheis: This could use a little bit more explanation or we could move it into a utility function `tryConfirm` or something

packages/protocol/engine/consensus/acceptance/gadget.go
```diff
@@ -158,29 +221,40 @@ func (a *Gadget) block(id models.BlockID) (block *Block, exists bool) {
 	return storage.Get(id)
 }
 
-func (a *Gadget) propagateAcceptance(marker markers.Marker) (queuedBlocks []*Block) {
+func (a *Gadget) propagateAcceptance(marker markers.Marker, confirmed bool) (acceptedBlocks, confirmedBlocks []*Block) {
```
jonastheis: `propagateAcceptance` is not accurate. It should be `propagateAcceptanceConfirmation`

packages/protocol/engine/consensus/acceptance/gadget.go
```diff
@@ -158,29 +221,40 @@ func (a *Gadget) block(id models.BlockID) (block *Block, exists bool) {
...
 		// this can happen when block was a root block and while processing this method, the root blocks method has already been replaced
 		return
 	}
 
 	pastConeWalker := walker.New[*Block](false).Push(block)
 	for pastConeWalker.HasNext() {
 		walkerBlock := pastConeWalker.Next()
-		if !walkerBlock.SetQueued() {
-			continue
+
```
jonastheis, line 252: We don't want to walk twice, that's why the confirmation is huddled together with the acceptance, right? But this makes the code a bit ugly and two distinct concepts are mixed together. Maybe we can rethink this a little bit so that the current `Gadget` becomes a wrapper component that houses the blocks, makes the walk etc but each actual `Gadget` for `acceptance` and `confirmation` lives in a separate struct. Each owning their own causal order, last*Marker, etc. These components seem so similar, maybe we can even make one implementation and make it configurable. Wdyt?

jonastheis (dismissed):

packages/core/eviction/state.go
piotrm50: Or not at all, as this file is going to be removed entirely and we won't have to store this value.

packages/core/votes/epochtracker/epochtracker.go
piotrm50: `previousHighestIndex >= epochIndex` - because we don't want to trigger an event the issuer already votes for the epoch.

packages/protocol/engine/consensus/consensus.go
piotrm50: :+1: I'll try to come up with something more general.

packages/protocol/engine/consensus/acceptance/gadget.go
piotrm50: Moved the whole if to a separate function with a comment and added more inline comments.

packages/protocol/engine/consensus/acceptance/gadget.go
piotrm50: > We don't want to walk twice, that's why the confirmation is huddled together with the acceptance 

This, and also because acceptance threshold depends on whether there is enough online weight to confirm. That makes the two concepts intertwined, as there is not much code that is unique to either confirmation or acceptance. It seems like separating those two concepts while retaining single walk would complicate the code.

piotrm50 (approved):

PR #2459, opened 2022-11-10
Feat/add commitment metrics

daria305:
# Description of change

Adds a new dashboard showing metrics from many nodes at once.
Mainly focused on the commitments data for now:
- the latest commited epoch, 
- commitment IDs per epoch
- number of blocks/tx/active validators per epoch
- number of blocks removed from epoch
- number of requested/received missing commitments

Dashboard name: Global Nodes View
image
image
image

## Type of change

docker-net


jkrvivian (approved):

packages/protocol/engine/notarization/manager.go
```diff
@@ -124,8 +124,17 @@ func (m *Manager) createCommitment(index epoch.Index) (success bool) {
 
 	m.pendingConflictsCounters.Delete(index)
 
-	stateRoot, manaRoot := m.storage.ApplyStateDiff(index, m.storage.LedgerStateDiffs.StateDiff(index))
+	blocksCount := m.EpochMutations.TotalAcceptedBlocks(index)
+	txsCount := m.EpochMutations.TotalAcceptedTransactions(index)
+	validatorsCount := m.EpochMutations.TotalActiveValidators(index)
+	epochDetails := &EvictedEpochDetails{
```
jonastheis: Instead of always doing this work just for the metrics, would it be possible to listen to the `EpochCommitted` event and just retrieve that information from the metrics plugin? In that way this is only done when the plugin is activated and we can get rid of the `EvictedEpochDetails` event as it is a bit misleading.

plugins/metrics/plugin.go
```diff
@@ -263,6 +264,15 @@ func registerLocalMetrics() {
 		}
 	}))
 
+	// Orphaned block counter that is removed successfully
+	deps.Protocol.Events.Engine.EpochMutations.AcceptedBlockRemoved.Attach(event.NewClosure(func(blkID models.BlockID) {
+		increaseRemovedBlockCounter(blkID)
+	}))
+
+	deps.Protocol.Events.Engine.NotarizationManager.EvictedEpochDetails.Attach(event.NewClosure(func(d *notarization.EvictedEpochDetails) {
```
jonastheis: Here I'd attach to `EpochCommitted` and just call the corresponding functions instead of this event.

plugins/metrics/epochcommittment.go
```diff
@@ -1,28 +1,168 @@
...
+
+func increaseRemovedBlockCounter(blkID models.BlockID) {
+	orphanedBlkRemovedMutex.Lock()
+	defer orphanedBlkRemovedMutex.Unlock()
+
+	if count, exists := orphanedBlkRemoved.Get(blkID.EpochIndex); exists {
+		count++
+		orphanedBlkRemoved.Set(blkID.EpochIndex, count)
+	} else {
+		orphanedBlkRemoved.Set(blkID.EpochIndex, 1)
```
jonastheis: even if it is not set the get should always return the default value (in this case 0). So the if shouldn't be necessary

plugins/metrics/epochcommittment.go
```diff
@@ -1,28 +1,168 @@
...
 )
 
 var (
-	lastCommittedEpoch      = new(commitment.Commitment)
+	lastCommitment          = new(commitment.Commitment)
 	lastCommittedEpochMutex sync.RWMutex
+
+	acceptedBlocksCount       atomic.Int32
+	acceptedTransactionsCount atomic.Int32
+	activeValidatorsCount     atomic.Int32
```
jonastheis: Why are these `atomic.Int32`? It seems that they are protected by a mutex anyway?

plugins/metrics/epochcommittment.go
```diff
@@ -1,28 +1,168 @@
...
+	currentEI                 epoch.Index
+	epochCountsMutex          sync.RWMutex
+
+	numberOfSeenOtherCommitments = 0
+	seenOtherCommitmentsMutex    sync.RWMutex
+
+	missingCommitments         = 0
+	missingCommitmentsReceived = 0
+	commitmentsMutex           sync.RWMutex
```
jonastheis: These could become atomics. Or you can just leave them with their mutexes.

jonastheis (commented):
I left a few comments. How do I make it display other nodes than `peer_master`, `peer_master2` and `faucet`?

daria305:
To show other instances they need to be listed in `tools/docker-network/prometheus.yml`. For sure we can add also peer no mana, and peer replica 1 by including:
```
- peer_nomana:9311
- peer_replica:9311
```
But I don't know how to add more replicas or if it's possible to do so automatically

packages/protocol/engine/notarization/manager.go
daria305: I was doing that in the first place, and the details were already forgotten, or at least it seemed so. That's I added this event. Hmm I'll try again, maybe something changed, or I did something incorrectly the first time

daria305: nah, it will not work in the current state of `commit` function. Maybe the eviction of all the data structures is delayed from the `epochCommited` event but the eviction of the epoch is now triggered right in `packages/protocol/engine/notarization/epochmutations.go:186`. Not sure why this is set this way now. 
I can remove this func from there and trigger eviction of notarization mgr epoch data on `EpochEvicted` event if there is no real purpose behind the current state

packages/protocol/engine/notarization/manager.go
```diff
@@ -45,7 +52,13 @@ func NewManager(storage *storage.Storage, sybilProtection *sybilprotection.Sybil
...
 		acceptanceTime:             storage.Settings.LatestCommitment().Index().EndTime(),
 		optsMinCommittableEpochAge: defaultMinEpochCommittableAge,
-	}, opts)
+
+		evictionState: evictionState,
+	}, opts, (*Manager).setupEvents)
+}
+
+func (m *Manager) setupEvents() {
+	m.evictionState.Events.EpochEvicted.Hook(event.NewClosure(m.evictUntil))
```
jonastheis: This doesn't seem right. Also if it's like this we probably call `evictUntil` twice: here and in `(m *EpochMutations) Commit`. I think the cleanest way would be to add this information to `Events.EpochCommitted` as we get all the information in `createCommitment` in line `acceptedBlocks, acceptedTransactions, activeValidators, cumulativeWeight, err := m.EpochMutations.Commit(index)`. Then you can also get rid of the getters in `EpochMutations`

tools/docker-network/prometheus.yml
```diff
@@ -6,3 +6,5 @@ scrape_configs:
           - peer_master:9311
           - faucet:9311
           - peer_master2:9311
+          - peer_nomana:9311
+          - peer_replica:9311
```
jonastheis: Instead of naming them statically maybe you can try something like this https://stackoverflow.com/questions/70803245/prometheus-service-discovery-with-docker-compose

jonastheis (changes requested):

packages/protocol/engine/notarization/manager.go
daria305: ok, extending EpochCommitted looks kinda nicer, and actually, the `acceptedBlocks`, and the rest still exist due to reference, which I took wrongly the last time

jonastheis (approved):

PR #2475, opened 2022-11-23
Refactor: Rollback support for ledger state

hmoog:
# Description of change

This PR refactors the ledger state to allow it to be rolled back to an arbitrary epoch in the past.

This feature can be used for snapshotting, crash-recovery or possible chain-reorgs in situations like eclipse attacks.

## Type of change

## Change checklist


packages/core/storable/types.go
```diff
@@ -19,7 +19,7 @@ type SerializableInt64 int64
 
 func (s SerializableInt64) Bytes() (data []byte, err error) {
 	data = make([]byte, binary.MaxVarintLen64)
-	binary.PutUvarint(data, uint64(s))
+	binary.LittleEndian.PutUint64(data, uint64(s))
```
jonastheis, line 22: isn't that dangerous? we write an int64 as a uint64. What if it is negative?

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -194,7 +194,7 @@ func (b *BlockDAG) canAttach(data *models.Block) (block *Block, canAttach bool,
 // this condition but exists as a missing entry, we mark it as invalid.
 func (b *BlockDAG) canAttachToParents(storedBlock *Block, data *models.Block) (block *Block, canAttach bool, err error) {
 	for _, parentID := range data.Parents() {
-		if b.EvictionState.InEvictedEpoch(parentID) && !b.EvictionState.IsRootBlock(parentID) {
+		if parentID != models.EmptyBlockID && b.EvictionState.InEvictedEpoch(parentID) && !b.EvictionState.IsRootBlock(parentID) {
```
jonastheis: Shouldn't this check `parentID != models.EmptyBlockID` be covered by `IsRootBlock` when `EmptyBlockID` is set at the beginning?

plugins/webapi/mana/nhighest.go
```diff
@@ -10,36 +10,36 @@ import (
 	"github.com/iotaledger/hive.go/core/identity"
 
 	"github.com/iotaledger/goshimmer/packages/app/jsonmodels"
-	"github.com/iotaledger/goshimmer/packages/protocol/engine/manatracker/manamodels"
+	manamodels2 "github.com/iotaledger/goshimmer/packages/protocol/engine/throughputquota/mana1/manamodels"
```
jonastheis: Rename import

plugins/faucet/connector.go
```diff
@@ -43,7 +43,7 @@ func (f *Connector) UnspentOutputs(addresses ...address.Address) (unspentOutputs
 								ConfirmationStateReached: outputMetadata.ConfirmationState().IsAccepted(),
 								Spent:                    false,
 								Metadata: wallet.OutputMetadata{
-									Timestamp: outputMetadata.CreationTime(),
+									Timestamp: outputMetadata.InclusionEpoch().EndTime(),
```
jonastheis, line 46: That means we should use this all over the place. At the same time a timestamp granularity == epochs for outputs and tx no?

plugins/dashboard/manafeed.go
```diff
@@ -15,8 +15,8 @@ import (
 	"github.com/mr-tron/base58"
 
 	"github.com/iotaledger/goshimmer/packages/core/shutdown"
-	"github.com/iotaledger/goshimmer/packages/protocol/engine/manatracker/manamodels"
 	"github.com/iotaledger/goshimmer/packages/protocol/engine/sybilprotection"
+	manamodels2 "github.com/iotaledger/goshimmer/packages/protocol/engine/throughputquota/mana1/manamodels"
```
jonastheis: Change import

packages/app/retainer/retainer.go
```diff
@@ -83,8 +84,8 @@ func (r *Retainer) Stream(index epoch.Index, callback func(id models.BlockID, me
 		return
 	}
 
-	r.blockStorage.Iterate(index, func(id models.BlockID, metadata *BlockMetadata) bool {
-		callback(id, metadata)
+	r.blockStorage.Iterate(index, func(id models.BlockID, metadata BlockMetadata) bool {
```
jonastheis: How exactly does this work with the generics here? Doesn't that create lots of copies of the objects when we change it to not receive a pointer anymore?

packages/app/retainer/retainer.go
```diff
@@ -57,7 +57,8 @@ func (r *Retainer) BlockMetadata(blockID models.BlockID) (metadata *BlockMetadat
 		return blockMetadata, blockExists
 	}
 
-	metadata, exists = r.blockStorage.Get(blockID)
+	metadata = new(BlockMetadata)
+	*metadata, exists = r.blockStorage.Get(blockID)
```
jonastheis, line 61: How exactly does this work with the generics here? Doesn't that create lots of copies of the objects when we change it to not receive a pointer anymore?

packages/protocol/engine/throughputquota/mana1/throughputquota.go
```diff
@@ -0,0 +1,195 @@
...
+	defer m.totalBalanceMutex.RUnlock()
+
+	return m.totalBalance
+}
+
+func (m *ThroughputQuota) ApplyCreatedOutput(output *ledger.OutputWithMetadata) (err error) {
+	if iotaBalance, exists := output.IOTABalance(); exists {
+		m.updateMana(output.AccessManaPledgeID(), int64(iotaBalance))
+
+		if !m.engine.LedgerState.UnspentOutputs.WasInitialized() {
```
jonastheis, line 125: Why do we only write through when `LedgerState.UnspentOutputs` was not initialized?

packages/network/protocol.go
```diff
@@ -140,7 +125,7 @@ func (p *Protocol) onBlock(blockData []byte, id identity.ID) {
 
 		return
 	}
-	block.DetermineIDFromBytes(blockData, blockHash)
+	block.DetermineID()
```
jonastheis: This basically means we hash twice

packages/protocol/engine/eviction/state.go
```diff
@@ -117,10 +116,14 @@ func (s *State) RemoveRootBlock(id models.BlockID) {
 
 // IsRootBlock returns true if the given block is a root block.
 func (s *State) IsRootBlock(id models.BlockID) (has bool) {
+	if id == models.EmptyBlockID {
+		return true
+	}
```
jonastheis, line 121: Why should we treat `EmptyBlockID` special?

packages/protocol/engine/eviction/state.go
```diff
@@ -30,16 +32,13 @@ type State struct {
...
-		if s.importRootBlocksFromStorage() == 0 {
-			s.AddRootBlock(models.EmptyBlockID)
-		}
-	})
+		Events:                      NewEvents(),
+		rootBlocks:                  memstorage.NewEpochStorage[models.BlockID, bool](),
+		latestRootBlock:             models.EmptyBlockID,
+		storage:                     storageInstance,
+		lastEvictedEpoch:            storageInstance.Settings.LatestCommitment().Index(),
+		optsRootBlocksEvictionDelay: 3,
```
jonastheis, line 40: Why this delay? Why 3 epochs by default?

packages/app/retainer/retainer.go
karimodm: We persist the object itself, not the pointer. This means that regardless of where you create the pointer to such deserialized object when retrieving it, you are still creating in-memory copies of the same object if you retrieve it multiple times. It seemed cleaner for the KVStore to return the value itself rather than a pointer to it though, at least you can more naturally support basic types.

packages/core/storable/types.go
karimodm: Not really: in the uint case you are treating the extra bit as another power of 2 instead of a sign; as long as you dump and retrieve the whole 8 bytes it is then just a matter of you intepret it (and the value is not expanded upon read and such ofc.).

You can try yourself:
```
func main() {
	var s SerializableInt64 = 42
	sRec := new(SerializableInt64)

	data, _ := s.Bytes()
	sRec.FromBytes(data)
	fmt.Println(s, *sRec)

	s = -101
	data, _ = s.Bytes()
	sRec.FromBytes(data)
	fmt.Println(s, *sRec)

	s = math.MaxInt64
	data, _ = s.Bytes()
	sRec.FromBytes(data)
	fmt.Println(s, *sRec)

	s = -math.MaxInt64
	data, _ = s.Bytes()
	sRec.FromBytes(data)
	fmt.Println(s, *sRec)

	s = -math.MaxInt64 - 1
	data, _ = s.Bytes()
	sRec.FromBytes(data)
	fmt.Println(s, *sRec)
}
```

packages/protocol/engine/eviction/state.go
karimodm: We wanted it to be a bit bigger than TSC; as TSC stands as 2 epochs (20 seconds), we made this 30s.

packages/protocol/engine/eviction/state.go
karimodm: Because otherwise blocks attaching to Genesis won't be considered solid, as one of the base conditions is that parents are checked to be cointained in the root blocks set.

packages/network/protocol.go
karimodm: Why twice? The blockData doesn't include the ID of the block, as it is never sent over the wire. It is true that we obtain the bytes of the object right after having deserialized it, but it is maybe better to compute the ID on the logical object's serialization rather than the raw bytes sent from the network.

packages/app/retainer/retainer.go
karimodm: See comment below!

plugins/dashboard/manafeed.go
karimodm: Goland strikes back ;)

packages/protocol/engine/tangle/blockdag/blockdag.go
karimodm: Yep, I think in this case is redundant!

packages/protocol/engine/throughputquota/mana1/throughputquota.go
karimodm: Because the total balance is the sum of all _initial_ outputs. Any other output happening after that is just a repledge, but the total does never change anymore.

plugins/faucet/connector.go
karimodm: Indeed! We might have missed some spots =). We had the granularity discussion with Hans last week, and we couldn't really come up with a use-case to have a more fine-grained granularity than epochs... but well: I'd be glad to know your opinion on the matter!

plugins/webapi/mana/nhighest.go
karimodm: Fixed

plugins/faucet/connector.go
hmoog: Actually, I am currently trying to write a sketch of the mana decay and demurrage logic (as research claims that this is hard to implement) and it looks like we need two times:

The `acceptanceTime` which is determined based on which epoch the tx was accepted in (epoch granularity) and the `creationTime` which is inherited from the timestamp of the transaction (full time granularity).

I think this makes much more sense and we chould most probably add the CreationTime back in and set it based on the tx timestamp.

hmoog: Okay, it looks like we are aligned around the decays which means that we need the CreationTime back in,

I will add it back, so we can use it in the corresponding throughput quota implementation.

packages/core/storable/types.go
alexsporn: why is it using `binary.MaxVarintLen64` if we are using uint64? This uses 10 bytes instead of 8

alexsporn (approved):

PR #2515, opened 2023-01-10
Improve web api

luca-moser:


packages/core/ads/map.go
```diff
@@ -141,3 +156,21 @@ func (m *Map[K, V, KPtr, VPtr]) Stream(callback func(key K, value VPtr) bool) (e
 
 	return
 }
+
+// has returns true if the key is in the map.
+func (m *Map[K, V, KPtr, VPtr]) has(keyBytes []byte) (has bool) {
```
jcburuchaga, line 161: can you reduce ifs nesting `

	has, err := m.tree.Has(keyBytes)
	if err == nil {
        return true
    } 
		
    if errors.Is(err, kvstore.ErrKeyNotFound) {
        return false
    }

    panic(err)
`

PR #2521, opened 2023-01-10
Fix global worker pool contention

### LINKED ISSUE #2507: Buffer/Consumer scheme for high-contention components
* Currently, event handlers are executed from the main workerpool workers: if the actions described in the event handler are not highly parallelizable (i.e. they lock on the same lock), all these workers are just stuck waiting for the lock.
* A better design would be to have a separate event consumer worker that deals with such tasks _sequentially_. The eventloop workers will then just push to a queue that the mentioned consumer consumes.
* We could pontentially change the `Attach` method to optionally pass a dedicated unbounded queue workerpool, using which, such event in consumed through.

jonastheis:
# Description of change

This PR solves the global worker pools contention problem, where all/many workers of the global event loop/worker pool were competing for the same lock. This is done by adding separate consumers to tasks that are prone to create contention. In code this reflects via specifying a distinct worker pool when attaching to an event.

It updates the Grafana dashboard `GoShimmer Local Metrics` to show all worker pools
image

Fixes #2507

### DIFF
--- a/packages/protocol/congestioncontrol/congestioncontrol.go
+++ b/packages/protocol/congestioncontrol/congestioncontrol.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/iotaledger/hive.go/core/generics/event"
 	"github.com/iotaledger/hive.go/core/generics/options"
+	"github.com/iotaledger/hive.go/core/workerpool"
 
 	"github.com/iotaledger/goshimmer/packages/protocol/congestioncontrol/icca/scheduler"
 	"github.com/iotaledger/goshimmer/packages/protocol/engine"
@@ -14,6 +15,7 @@ import (
 type CongestionControl struct {
 	Events *Events
 
+	workerPool     *workerpool.UnboundedWorkerPool
 	scheduler      *scheduler.Scheduler
 	schedulerMutex sync.RWMutex
 
@@ -21,9 +23,14 @@ type CongestionControl struct {
 }
 
 func New(opts ...options.Option[CongestionControl]) (congestionControl *CongestionControl) {
-	return options.Apply(&CongestionControl{
-		Events: NewEvents(),
+	congestionControl = options.Apply(&CongestionControl{
+		Events:     NewEvents(),
+		workerPool: workerpool.NewUnboundedWorkerPool(1),
 	}, opts)
+
+	congestionControl.workerPool.Start()
+
+	return congestionControl
 }
 
 func (c *CongestionControl) LinkTo(engine *engine.Engine) {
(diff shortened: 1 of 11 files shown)


packages/protocol/engine/clock/clock.go
```diff
@@ -74,28 +70,28 @@ func (c *Clock) SetConfirmedTime(confirmedTime time.Time) (updated bool) {
...
-	return c.lastConfirmedTime.Add(time.Since(c.lastConfirmedTimeUpdated))
+	return c.lastConfirmed.Time.Add(time.Since(c.lastConfirmed.Updated))
 }
 
-// updateTime updates the given time parameter if the given time larger than the current time.
-func (c *Clock) updateTime(now, newTime time.Time, param, updatedParam *time.Time) (updated bool) {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
+// updateTime updates the given Time parameter if the given Time larger than the current Time.
+func (c *Clock) updateTime(t *timeUpdate, now, newTime time.Time) (updated bool) {
```
alexsporn: move this to timeUpdate struct as a func

plugins/prometheus/workerpool.go
```diff
@@ -6,35 +6,46 @@ import (
 )
 
 var (
-	workerCount  prometheus.Gauge
-	pendingTasks prometheus.Gauge
+	// workerCount  prometheus.Gauge
```
alexsporn: remove commented out code

packages/protocol/congestioncontrol/congestioncontrol.go
```diff
@@ -14,16 +15,22 @@ import (
...
 
 func New(opts ...options.Option[CongestionControl]) (congestionControl *CongestionControl) {
-	return options.Apply(&CongestionControl{
-		Events: NewEvents(),
+	congestionControl = options.Apply(&CongestionControl{
+		Events:     NewEvents(),
+		workerPool: workerpool.NewUnboundedWorkerPool(1),
 	}, opts)
+
+	congestionControl.workerPool.Start()
```
karimodm, line 31: Only start the workerpool when you actually setup events. Like the retainer.

packages/protocol/engine/engine.go
```diff
@@ -346,17 +355,19 @@ func (e *Engine) initBlockStorage() {
...
+	}), 1)
+	e.Tangle.Events.BlockDAG.BlockOrphaned.AttachWithWorkerPool(event.NewClosure(func(block *blockdag.Block) {
 		if err := e.NotarizationManager.NotarizeOrphanedBlock(block.ModelsBlock); err != nil {
 			e.Events.Error.Trigger(errors.Errorf("failed to remove orphaned block %s from epoch: %w", block.ID(), err))
 		}
-	}))
+	}), wp)
+	e.workerPools["NotarizationManager.Blocks"] = wp
 
+	// TODO: Why is it hooked?
```
karimodm, line 370: If I recall correctly, it was initially to prevent the ledger from persisting Transactions in the statediffs out of order whenever the processed transaction spend from each other. This could be problematic in the case a transaction needs to be moved to another epoch, or deleted altogether. You have to be able to make sense of the outputs you are spending and creating in each epoch, and if such accepted transaction are not dealt with in order some inconsistencies might arise in the statediffs for each epoch: although unlikely, we wanted to play it safe!

karimodm (approved):

PR #2528, opened 2023-01-13
Fix TSC tip-fishing check

piotrm50:
# Description of change

Reimplementation of TSC check during tip fishining that does not employ markers, but always walks blocks. Block walk is optimized by a cache, so a correct part of the tangle is only walked once and then saved in cache.

## Type of change

Unit tests, docker network, feature network.

## Change checklist


karimodm (approved):

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -246,178 +255,29 @@ func (t *TipManager) isPastConeTimestampCorrect(block *booker.Block) (timestampV
...
-		}
-	}
-
-	// process markers from different sequences that are referenced by current marker's sequence, i.e., walk the sequence DAG
-	sequence.ReferencedMarkers(pair.Marker.Index()).ForEach(func(sequenceID markers.SequenceID, index markers.Index) bool {
-		// Ignore Marker(_, 0) as it sometimes occurs in the past marker cone. Marker mysteries.
-		// This should not be necessary anymore?
-		if index == 0 {
-			fmt.Println("(time: ", time.Now(), ") ignore Marker(_, 0) during tip fishing")
+func (t *TipManager) checkBlockRecursive(block *booker.Block, minSupportedTimestamp time.Time) (timestampValid bool) {
```
alexsporn, line 266: Is this recursion guaranteed to not overflow? Can we refactor this into a for-loop?

karimodm: It is not guaranteed, no. Implementing it iteratively is definitely possible, we will need to introduce a "breadcrumb" accumulator to remember and mark the path we are travelling from, while still exploring the graph in other directions whenever a certain path ends in a `true` return.
In go, the maximum allowed stack-depth is defined in bytes, rather than in # of frames. The limit is currently set at 1GB for 64 bit and 250MB for 32 bit architectures respectively.
We can expect very deep stacks whenever acceptance stops for a significant amount of time, but in those cases, the depth of this call stack is the least of our concerns: as we are not evicting any object from memory whenever acceptance stops.
For this reason I don't think this is an actual threat to our runtime, as we are way more likely to incur in OOM exceptions because of the objects cache before hitting the maximum call stack size.

PR #2529, opened 2023-01-13
Enforce commitment monotonicity in BlockDAG

### LINKED ISSUE #2526: Block filter to enfore monotonicity of epoch commitments
`curCommitIndex - parentcommitIndex >= 0`, no maximum distance enforced

karimodm:
# Description of change

Add block filter to enforce epoch commitment monotonicity.
Resolves #2526


alexsporn (approved):

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -131,6 +131,10 @@ func (b *BlockDAG) markSolid(block *Block) (err error) {
 		return err
 	}
 
+	if err := b.checkCommitmentMonotonicity(block); err != nil {
```
alexsporn: The code is cleaner by doing this checks one after the other, but you are re-iterating over the parents and loading them in both checks. It would be more efficient to do a `checkParents` function and do the steps inside the common loop one after the other.

alexsporn (approved):

PR #2530, opened 2023-01-13
TSA changes to account for epoch commitments

### LINKED ISSUE #2527: TSA adjustments to account for epoch commitments
* We park tips committing on a non-forking (but unknown) commitment.
* As soon as the epoch becomes committable we activate such parked tips by moving them to the main tippool.
* Do not select tips committing to an epoch < than current commitment index - 1.

karimodm:
# Description of change

Take into account epoch commitments in the TSA (see #2527).

Resolves #2527


packages/protocol/tipmanager/tipmanager.go
```diff
@@ -220,6 +214,84 @@ type markerPreviousBlockPair struct {
...
+			}
+		}
+	}
+
+	return false
+}
+
+// isStaleCommitment returns true if the block belongs to a commitment that is too far back in the past.
+func (t *TipManager) isStaleCommitment(block *scheduler.Block) (isOld bool) {
+	return block.Commitment().Index() < t.engine.Storage.Settings.LatestCommitment().Index()-1
```
alexsporn: Isn't `t.engine.Storage.Settings.LatestCommitment().Index()` 0 at the beginning?

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -220,6 +214,84 @@ type markerPreviousBlockPair struct {
...
+		)
+	}
+
+	return nil
+}
+
+// isRecentCommitment returns true if the commitment of the given block is not in the future and it is not older than 2
+// epoch with respect to our latest commitment.
+func (t *TipManager) isRecentCommitment(block *scheduler.Block) (isFresh bool) {
+	return !t.isFutureCommitment(block) && block.Commitment().Index() >= t.engine.Storage.Settings.LatestCommitment().Index()-1
```
alexsporn: same here, isn't `t.engine.Storage.Settings.LatestCommitment().Index()` 0 at the beginning?

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -216,19 +174,171 @@ func (t *TipManager) selectTips(count int) (parents models.BlockIDs) {
...
 
+func (t *TipManager) addTip(block *scheduler.Block) (added bool) {
+	if !t.tips.Has(block.ID()) {
+		t.tips.Set(block.ID(), block)
+		// t.tipsConflictTracker.AddTip(block)
+		t.Events.TipAdded.Trigger(block)
+
+		// skip removing tips if a width is set -> allows to artificially create a wide Tangle.
+		if t.TipCount() <= t.optsWidth {
+			return
```
piotrm50:
```suggestion
			return true
```

piotrm50: Tip was successfully added, but we just artificially don't remove parents to create wider tangle, so this should return true

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -216,19 +174,171 @@ func (t *TipManager) selectTips(count int) (parents models.BlockIDs) {
...
+	if futureEpochTips := t.futureTips.Get(cm.Index()); futureEpochTips != nil {
+		if tipsForCommitment, exists := futureEpochTips.Get(cm.ID()); exists {
+			tipsToPromote := make(map[models.BlockID]*scheduler.Block)
+			tipsForCommitment.ForEach(func(blockID models.BlockID, tip *scheduler.Block) bool {
+				tipsToPromote[blockID] = tip
+				return true
+			})
+
+			tipsToNotPromote := set.NewAdvancedSet[models.BlockID]()
+			for _, tip := range tipsToPromote {
```
piotrm50: Do we actually need this loop? I think we could populate `tipsToNotPromote` in the same loop as `tipsToPromote`?

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -216,19 +174,171 @@ func (t *TipManager) selectTips(count int) (parents models.BlockIDs) {
...
+			tip.IsScheduled(),
+			tip.IsOrphaned(),
+			t.engine.Consensus.BlockGadget.IsBlockAccepted(tip.ID()),
+		)
+	}
+
+	return nil
+}
+
+// isRecentCommitment returns true if the commitment of the given block is not in the future and it is not older than 2
```
piotrm50:
```suggestion
// isRecentCommitment returns true if the commitment of the given block is not in the future and it is not older than TSC threshold
```

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -216,19 +174,171 @@ func (t *TipManager) selectTips(count int) (parents models.BlockIDs) {
...
+			}
+		}
+	}
+
+	return false
+}
+
+// isStaleCommitment returns true if the block belongs to a commitment that is too far back in the past.
+func (t *TipManager) isStaleCommitment(block *scheduler.Block) (isOld bool) {
+	return block.Commitment().Index() < (t.engine.Storage.Settings.LatestCommitment().Index() - 1).Max(0)
```
piotrm50: Shouldn't that also use the epochDelay as defined in `isRecentCommitment`?  That way it could call `isStaleCommitment` instead of repeating the comparison.
```
epochDelay := epoch.Index(int64(t.optsTimeSinceConfirmationThreshold.Seconds()) / epoch.Duration)
```

piotrm50 (changes requested):

packages/protocol/tipmanager/tipmanager.go
karimodm: I don't think you actually can. Because in order to know what tips to avoid adding you need to check if in _any_ of the tips has some other tip as a parent; to do so you need first to collect all potential tips, and only then you can iterate through every single one of them and remove their parents from the potential tips set.

packages/protocol/tipmanager/tipmanager.go
karimodm: Yes indeed, I can change the order of the checks in `AddTip` and reuse the same check method.

packages/protocol/tipmanager/tipmanager.go
piotrm50: Yeah, but now we basically iterate twice over the same set of blocks. We could as well collect their parents in the same run and then this third loop stays the same
```
			for tipID, tip := range tipsToPromote {
				if !tipsToNotPromote.Has(tipID) {
					t.addTip(tip)
				}
			}
```

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -216,19 +180,165 @@ func (t *TipManager) selectTips(count int) (parents models.BlockIDs) {
...
+		t.Evict(cm.Index())
+	}()
+
+	if futureEpochTips := t.futureTips.Get(cm.Index()); futureEpochTips != nil {
+		if tipsForCommitment, exists := futureEpochTips.Get(cm.ID()); exists {
+			tipsToPromote := make(map[models.BlockID]*scheduler.Block)
+			tipsForCommitment.ForEach(func(blockID models.BlockID, tip *scheduler.Block) bool {
+				tipsToPromote[blockID] = tip
+				return true
+			})
```
piotrm50:
```suggestion
			tipsForCommitment.ForEach(func(blockID models.BlockID, tip *scheduler.Block) bool {
				tipsToPromote[blockID] = tip
				
				for _, parentID := range tip.Parents() {
					if _, exists := tipsToPromote[parentID]; exists {
						// Do not add this tips parent because this is not a tip anymore
						tipsToNotPromote.Add(parentID)
					}
				}
				// Remove all tips referenced by all future tips valid for this commitment
				t.RemoveStrongParents(tip.ModelsBlock)
				
				return true
			})
			      
```

piotrm50 (approved):

packages/protocol/tipmanager/tipmanager.go
karimodm: This you cannot really do, because you are filling `tipsToPromote` while you are checking membership in it. So the end result will depend in what order the outer ForEach will iterate! Hence why we need to fill `tipsToPromote` entirely first.

PR #2531, opened 2023-01-13
Metrics refactor: collector framework for metrics creation

daria305:
# Description of change

  - `prometheus` plugin was deleted
  - cleaned up old `metrics` plugin stays for now until we use dashboard, is renamed to `dashboardmetrics`
All prometheus metrics are now created with new `metrics` plugin. that handles both old `prometheus` and `metrics` plugin.
 - updated most of existing Grafana dashboards

## Type of change

Works on docker network and feature net


plugins/metrics/plugin.go
```diff
@@ -1,319 +1,156 @@
...
 
-	Protocol    *protocol.Protocol
-	BlockIssuer *blockissuer.BlockIssuer
+	Local                 *peer.Local
+	Protocol              *protocol.Protocol
+	BlockIssuer           *blockissuer.BlockIssuer
+	P2Pmgr                *p2p.Manager        `optional:"true"`
+	Selection             *selection.Protocol `optional:"true"`
+	Retainer              *retainer.Retainer  `optional:"true"`
+	AutoPeeringConnMetric *autopeering.UDPConnTraffic
```
jkrvivian:
```suggestion
	AutopeeringConnMetric *autopeering.UDPConnTraffic
```

tools/docker-network/docker-compose.yml
```diff
@@ -12,18 +12,16 @@ services:
...
       --protocol.snapshot.path=./snapshot.bin
       --blockIssuer.rateSetter.enable=false
-      --metrics.global=true
-      --metrics.local=true
-      --prometheus.bindAddress=0.0.0.0:9311
-      --prometheus.processMetrics=true
-      --prometheus.goMetrics=true
+      --metrics.bindAddress=0.0.0.0:9311
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
```
jkrvivian, line 23:
```suggestion
      --metricsCollector.bindAddress=0.0.0.0:9311
      --metricsCollector.processMetrics=true
      --metricsCollector.goMetrics=true
```

tools/docker-network/docker-compose.yml
```diff
@@ -50,15 +48,15 @@ services:
...
       --node.overwriteStoredSeed=true
       --protocol.snapshot.path=./snapshot.bin
       --blockIssuer.rateSetter.enable=false
-      --prometheus.bindAddress=0.0.0.0:9311
+      --metrics.bindAddress=0.0.0.0:9311
       --protocol.genesisTime=${GENESIS_TIME}
-      --prometheus.processMetrics=true
-      --prometheus.goMetrics=true
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
```
jkrvivian, line 57: metricsCollector

tools/docker-network/docker-compose.yml
```diff
@@ -75,18 +73,16 @@ services:
...
       --protocol.snapshot.path=./snapshot.bin
       --blockIssuer.rateSetter.enable=false
-      --metrics.global=true
-      --metrics.local=true
-      --prometheus.bindAddress=0.0.0.0:9311
-      --prometheus.processMetrics=true
-      --prometheus.goMetrics=true
+      --metrics.bindAddress=0.0.0.0:9311
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
```
jkrvivian, line 82:
```suggestion
      --metricsCollector.bindAddress=0.0.0.0:9311
      --metricsCollector.processMetrics=true
      --metricsCollector.goMetrics=true
```

tools/docker-network/docker-compose.yml
```diff
@@ -105,19 +101,20 @@ services:
...
-      --prometheus.bindAddress=0.0.0.0:9311
+      --metrics.bindAddress=0.0.0.0:9311
+      --metrics.bindAddress=0.0.0.0:9311
       --protocol.genesisTime=${GENESIS_TIME}
-      --prometheus.processMetrics=true
-      --prometheus.goMetrics=true
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
```
jkrvivian, line 113: Ditto. metricsCollector

tools/docker-network/docker-compose.yml
```diff
@@ -140,17 +137,15 @@ services:
...
-      --metrics.global=true
-      --metrics.local=true
       --protocol.snapshot.path=./snapshot.bin
       --blockIssuer.rateSetter.enable=false
-      --prometheus.bindAddress=0.0.0.0:9311
-      --prometheus.processMetrics=true
-      --prometheus.goMetrics=true
+      --metrics.bindAddress=0.0.0.0:9311
+      --metrics.processMetrics=true
+      --metrics.goMetrics=true
```
jkrvivian, line 143: metricsCollector

packages/app/collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Collector is responsible for creation and collection of metrics for the prometheus.
+type Collector struct {
+	Registry    *prometheus.Registry
+	collections map[string]*Collection
+}
+
+// New creates an instance of Manger, creates new prometheus registry for the protocol metrics collection.
```
jkrvivian:
```suggestion
// New creates an instance of Manager and creates a new prometheus registry for the protocol metrics collection.
```

packages/app/collector/metric.go
```diff
@@ -0,0 +1,260 @@
...
+
+// Update updates the metric value, for Gauge/GaugeVec values are set, for Counter/CounterVec values are added.
+// For CounterVec/GaugeVec metrics updates will be done only if each provided metric label was previously defined with WithLabels option.
+// To set metrics labels string values, enable use WithLabelValuesCollection option.
+func (m *Metric) Update(values map[string]float64) {
+	if m.labelValuesCollectionEnabled {
+		value := float64(0)
+		if len(values) != len(m.labels) {
+			fmt.Println("Warning! Nothing updated, label values and labels length mismatch when updating metric", m.Name)
+			return
```
jkrvivian, line 106: Return an error instead?

packages/app/collector/metric.go
```diff
@@ -0,0 +1,260 @@
...
+			m.PromMetric.(*prometheus.CounterVec).WithLabelValues().Inc()
+		}
+	} else {
+		switch m.Type {
+		case GaugeVec:
+			m.PromMetric.(*prometheus.GaugeVec).WithLabelValues(labels...).Inc()
+		case CounterVec:
+			m.PromMetric.(*prometheus.CounterVec).WithLabelValues(labels...).Inc()
+		}
+	}
```
jkrvivian, line 176: the entire section should be equivalent to this:
```go
switch m.Type {
	case Gauge:
		m.PromMetric.(prometheus.Gauge).Inc()
	case GaugeVec:
		m.PromMetric.(*prometheus.GaugeVec).WithLabelValues(labels...).Inc()
	case Counter:
		m.PromMetric.(prometheus.Counter).Inc()
	case CounterVec:
		m.PromMetric.(*prometheus.CounterVec).WithLabelValues(labels...).Inc()
	}
```

packages/protocol/engine/throughputquota/mana1/manamodels/apiutils.go
```diff
@@ -68,6 +69,17 @@ func Percentile(id identity.ID, m map[identity.ID]int64) (percentileValue float6
 	return (nBelow / float64(len(m))) * 100
 }
 
+// NeighborsAverageMana returns the average mana of the eighbors, based on provided mana vector and neighbors list.
```
jkrvivian:
```suggestion
// NeighborsAverageMana returns the average mana of the neighbors, based on provided mana vector and neighbors list.
```

packages/protocol/engine/throughputquota/mana1/manamodels/apiutils.go
```diff
@@ -68,6 +69,17 @@ func Percentile(id identity.ID, m map[identity.ID]int64) (percentileValue float6
...
+func NeighborsAverageMana(m map[identity.ID]int64, neighbors []*p2p.Neighbor) (avg float64) {
+	if len(m) == 0 {
+		return 0
+	}
+	for _, n := range neighbors {
+		avg += float64(m[n.Peer.ID()])
+	}
+	return avg / float64(len(m))
+}
```
jkrvivian, line 82: why is `neighbors` needed? we can iterate `m`, sum all mana and divide by `len(m)`

plugins/core.go
```diff
@@ -1,20 +1,21 @@
 package plugins
 
 import (
+	"github.com/iotaledger/goshimmer/plugins/metrics"
```
jkrvivian: rename the `metrics` to `metricsCollector`, which is the name of this plugin, or the other way around.

plugins/dashboardmetrics/block.go
```diff
@@ -0,0 +1,205 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/iotaledger/hive.go/core/syncutils"
+	"go.uber.org/atomic"
+)
+
+// BlockType defines the component for the different BPS metrics.
+type BlockType byte
```
jkrvivian: Reuse the type defined in `collector`? They're exactly the same.

plugins/dashboardmetrics/plugin.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/iotaledger/goshimmer/packages/network/p2p"
+	"github.com/iotaledger/goshimmer/packages/protocol"
+	"github.com/iotaledger/goshimmer/packages/protocol/congestioncontrol/icca/scheduler"
+	"github.com/iotaledger/goshimmer/packages/protocol/engine/tangle/blockdag"
+	"github.com/iotaledger/goshimmer/packages/protocol/engine/tangle/booker"
+	"github.com/iotaledger/goshimmer/packages/protocol/models"
+)
+
+// PluginName is the name of the metrics plugin.
+const PluginName = "Metrics"
```
jkrvivian: `dashboardMetrics`

plugins/metrics/parameters.go
```diff
@@ -1,24 +1,24 @@
...
+	PromhttpMetrics bool `default:"false" usage:"include promhttp metrics"`
 }
 
-// Parameters contains the configuration used by the metrics plugin.
+// Parameters contains the configuration used by the metrics collector plugin.
 var Parameters = &ParametersDefinition{}
 
 func init() {
-	config.BindParameters(Parameters, "metrics")
+	config.BindParameters(Parameters, "metricsCollector")
```
jkrvivian: need to decide which plugin name to use

packages/app/collector/metric.go
```diff
@@ -0,0 +1,260 @@
...
+// UpdateWithLabels allows to add new label values (string) and set/add value for them for correspondingly GaugeVec/CounterVec.
+// Provided label values should be ordered correspondingly to m.labels provided with WithLabels option.
+func (m *Metric) updateWithLabels(labelValues []string, val float64) {
+	switch m.Type {
+	case Gauge:
+	case GaugeVec:
+		m.PromMetric.(*prometheus.GaugeVec).WithLabelValues(labelValues...).Set(val)
+	case Counter:
+	case CounterVec:
+		m.PromMetric.(*prometheus.CounterVec).WithLabelValues(labelValues...).Add(val)
```
piotrm50, line 152: Does this correctly work if Counter is type-casted to CounterVec?

packages/protocol/congestioncontrol/icca/scheduler/scheduler.go
```diff
@@ -197,6 +197,10 @@ func (s *Scheduler) Deficit(issuerID identity.ID) *big.Rat {
 	return deficit
 }
 
+func (s *Scheduler) GetAccessManaMap() map[identity.ID]int64 {
```
piotrm50, line 203: That should not be exposed like that from here. Better to take the actual callback that the scheduler uses.

packages/protocol/engine/throughputquota/mana1/manamodels/apiutils.go
```diff
@@ -4,6 +4,7 @@ import (
 	"sort"
 	"time"
 
+	"github.com/iotaledger/goshimmer/packages/network/p2p"
```
piotrm50: Sort imports.

packages/protocol/engine/throughputquota/mana1/manamodels/apiutils.go
piotrm50: Can't do that, because it contains not only neighbors.

plugins/dashboard/ws.go
```diff
@@ -65,35 +65,26 @@ func configureWebSocketWorkerPool() {
...
-			Booker:     componentStatus[metrics.Booked],
+			Store:      componentStatus[dashboardmetrics.Attached],
+			Solidifier: componentStatus[dashboardmetrics.Solidified],
+			Scheduler:  componentStatus[dashboardmetrics.Scheduled],
+			Booker:     componentStatus[dashboardmetrics.Booked],
 		}
 		wsSendWorkerPool.TrySubmit(updateStatus)
 	})
-	updateRateSetterMetrics := event.NewClosure(func(metric *metrics.RateSetterMetric) {
-		wsSendWorkerPool.TrySubmit(&rateSetterMetric{
```
piotrm50, line 82: This is removed because ratesetter is not ready yet, right? That should be added eventually and visible on the dashboard.

plugins/dashboardmetrics/block.go
```diff
@@ -0,0 +1,205 @@
...
+		return "Booked"
+	default:
+		return fmt.Sprintf("Unknown (%d)", c)
+	}
+}
+
+// initial values at start of the node.
+var (
+	// number of solid blocks in the database at startup.
+	initialBlockCountPerComponentDB = make(map[ComponentType]uint64)
```
piotrm50: Is this populated anywhere? I think the idea behind it was so that metrics have correct values after node restart, but I don't think it makes sense now that we only store accepted blocks and the rest stays in memory.

plugins/dashboardmetrics/block.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// the same metrics as above, but since the start of a node.
+var (
+	// Number of blocks per component (store, scheduler, booker) type since start of the node.
+	// One for dashboard (reset every time is read), other for grafana with cumulative value.
+	blockCountPerComponentDashboard = make(map[ComponentType]uint64)
+	blockCountPerComponentGrafana   = make(map[ComponentType]uint64)
```
piotrm50: It seems to not be used anywhere.

plugins/dashboardmetrics/block.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// protect map from concurrent read/write.
+	blockCountPerComponentMutex syncutils.RWMutex
+
+	// current number of missing blocks in missingBlockStorage.
+	missingBlockCountDB atomic.Uint64
```
piotrm50: This metric seems to only be decreased, but never increased.

plugins/dashboardmetrics/block.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// sum of times since issued (since start of the node).
+	sumTimesSinceIssued = make(map[ComponentType]time.Duration)
```
piotrm50: It seems like this is not exposed anywhere even though the values are updated. Maybe we can remove the metric entirely, because I'm not sure what would be the correct way to display it and what would it tell us.

plugins/dashboardmetrics/plugin.go
```diff
@@ -0,0 +1,165 @@
...
+func run(_ *node.Plugin) {
+	log.Infof("Starting %s ...", PluginName)
+	registerLocalMetrics()
+
+	// create a background worker that update the metrics every second
+	if err := daemon.BackgroundWorker("Metrics Updater", func(ctx context.Context) {
+		// Do not block until the Ticker is shutdown because we might want to start multiple Tickers and we can
+		// safely ignore the last execution when shutting down.
+		timeutil.NewTicker(func() {
+			measureAttachedBPS()
```
piotrm50, line 61: This still seems to be over-engineered. The process looks as follows:
* Ticker runs measure funcs
* Measure methods trigger some event inside
* Dashboard is listening to the events and sends the metrics to the WebSocket.

Maybe the dashboard itself should have a sub-package that keeps track of metrics, and instead of triggering an event, we could use the ticker directly in the dashboard package and send metrics directly from the ticker callback? That way we wouldn't need to have dashboard metrics plugin at all - which kind of makes sense  because dashboard can't properly function without it, so it makes sense to just have one plugin.

plugins/metrics/metics_commitments.go
```diff
@@ -0,0 +1,99 @@
...
+			deps.Protocol.ChainManager().Events.ForkDetected.Attach(event.NewClosure(func(chain *chainmanager.Chain) {
+				deps.Collector.Increment(commitmentsNamespace, seenTotal)
+			}))
+		}),
+	)),
+	collector.WithMetric(collector.NewMetric(missingRequested,
+		collector.WithType(collector.Counter),
+		collector.WithHelp("Number of missing commitments requested by the node."),
+		collector.WithInitFunc(func() {
+			deps.Protocol.ChainManager().Events.CommitmentMissing.Attach(event.NewClosure(func(commitment commitment.ID) {
```
piotrm50, line 52: This whole plugin could use a separate worker pool, which was recently introduced, so the main workerpool is not bothered with low-priority tasks. It's done like that in the retainer plugin, so it can be used as a reference.

piotrm50 (changes requested):
Nice to finally see the metrics cleaned! Good work! Dashboard metrics could use a little cleanup as well, which I commented about.

plugins/core.go
daria305: `metrics` it is

plugins/dashboardmetrics/block.go
daria305: I assumed that we will remove dashboard soon anyway, so it doesn't matter, but you're right now we can use it

packages/app/collector/metric.go
daria305: It is not type-casted, for `Counter` case this function actually does nothing as what user tries to do is not possible, not sure if I should print some warning here. The same is for Gauge and GaugeVec

piotrm50: Ahh, ok. Forgot that in Go the switch cases do not fall through if you don't break :smile:

plugins/dashboardmetrics/block.go
daria305: It is used in Info API (plugins/webapi/info/plugin.go:164) for SolidBlockCount and TotalBlock Counts, I left it for now, as it does not bother there until metrics dashboard plugin exists.

plugins/dashboardmetrics/block.go
daria305: good catch, it's not needed

plugins/dashboardmetrics/block.go
daria305: yeah, it's unneeded leftover. Also I noticed that BPS on the dashboard was broken, ups

plugins/dashboard/ws.go
daria305: yes, I put it there again to not forget

plugins/dashboardmetrics/block.go
daria305: agree, removed

plugins/metrics/metics_commitments.go
daria305: will be addressed in separate PR

plugins/dashboardmetrics/plugin.go
daria305: will be addressed in separate PR

packages/protocol/congestioncontrol/icca/scheduler/scheduler.go
daria305: will be addressed in separate PR

piotrm50 (approved):
Ready to go :+1:

PR #2532, opened 2023-01-16
In memory ConflictDAG and transaction and attachment orphanage

### LINKED ISSUE #2513: Implement BranchDAG & ConflictDAG cache to avoid IO
* Possibly remove the ConflictResolver?

### LINKED ISSUE #2570: Fix conflict spamming
- [ ] ~conflicts should be [issued in parallel]~(https://github.com/iotaledger/goshimmer/blob/2d6495243d3190dfe41d0244bf38f5f80e432481/packages/app/blockissuer/blockfactory/factory.go#L192) otherwise we produce too many invalid blocks -> needs to be removed
- [x] we need to properly handle tx and conflict orphanage
- [x] we need to make sure that invalid votes are seen as such ([should be done, see commit](https://github.com/iotaledger/goshimmer/commit/15cb908dc0bf83c63b9343e7235499c0f5cd3d30)) and blocks marked as subjectively invalid. these should not be added to the tipset (and maybe not gossiped)
- [x] we need to make sure that the reference provider gives us the correct references, especially if a node already merged to master: never ever set a reference to something that is disliked -> do not pick up parents with rejected conflicts
- [ ] ~implement a proper eviction strategy for conflicts on epoch commit / epoch eviction event. (we need to remember disliked/rejected conflicts until they can't be referenced anymore naturally either below max evicted epoch or TSC)~
- [x] ensure commitment monotonicy on tip selection
- [x] revisit: only track attachments in notarization manager when a block is becoming accepted -> no need to handle orphanage (same as for blocks and transactions)
- [ ] ~revisit: drop unresolved conflicts after some time~
- [ ] ~revisit: orphan attachment on subjective invalidity & remove `firstValidAttachment` check subjectively invalid check~
- [ ] revisit: tipsconflicttracker too complicated? why not just always pick unresolved conflicts?
- [ ] ~revisit: introduce `confirmation.Orphaned` instead of setting conflicts to rejected~
- [ ] ~change booking logic to remove conflicting conflicts of the payload~
- [x] How do you pick up a TX that is in a subjectively invalid block?
- [x] How to prevent to pick up something that is too old?

---- 
----

- [x] conflict propagation in the tangle on fork seems to be still wrong
  - [x] we fixed the propagation via weak and like parents: we should double check whether that was everything that was needed. 
  - [x] can it be that propagation via subjectively invalid blocks might be weird?
- [x] always propagate acceptance forward in UTXO to avoid race conditions between inclusion and conflict acceptance
  - [x] we still need to make sure that TX are only monotonically accepted. that means we always need to walk into the future cone to check for further TXs becoming accepted once we accept a TX (but we do that already anyway I think)
- [ ] ~set conflicts that are created from previously accepted TX (via attachment weight, when it was not conflicting) to accepted right away. this should prevent the race condition in the last case on the jamboard where a tx is accepted -> conflict created (pending) -> eventually conflict accepted. but it might also elevate some other issues with nodes seemingly changing their opinion for a short period of time -> it also should increase performance as accepted things don't need to get propagated through the UTXO / tangle future cone~
- [x] do not track accepted / rejected conflicts anymore in conflict tracker
- [ ] ~blockfactory refuse block with disliked payload~
- [x] re-enable markers gap
- [ ] ~re-enable marker propagation~
- [ ] ~can we get rid of orphanage? or do we still need it protect memory and keep statistics for not included blocks?~
- [ ] ~evict conflicts from conflicttracker~
- [x] merge develop

jonastheis:
# Description of change

This PR introduces an in-memory ConflictDAG. The API is mostly the same to the previous object storage-based ConflictDAG except the models live in memory and cached objects are not necessary anymore.

Due to the nature of no inter epoch reorg, we don't need to evict the ConflictDAG. Whenever a conflict is accepted/rejected, we can delete its association with the conflict set. Once a conflict set has only resolved conflicts it can be deleted as well. 

Fixes #2570


packages/protocol/ledger/ledger_test.go
```diff
@@ -92,7 +92,7 @@ func TestLedger_BookInOrder(t *testing.T) {
 			"TX7*": {"TX1", "TX2", "TX3"},
 		})
 	}
-
+	return
```
piotrm50:
```suggestion

```

tools/double-spend/double-spend.go
```diff
@@ -2,115 +2,58 @@ package main
...
+		if !invalid(nextBlock) {
+			fmt.Println("Block is valid", nextBlock)
 
-	if _, err := clients[0].BroadcastFaucetRequest(myAddr.Address().Base58(), -1); err != nil {
-		fmt.Println(strings.ReplaceAll(err.Error(), "\n", ""))
-		return
-	}
+			block, err := c.GetBlock(prevBlock)
+			if err != nil {
+				panic(err)
```
piotrm50: Revert changes to this file.

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -3,94 +3,143 @@ package conflictdag
...
 
-	// options is a dictionary for configuration parameters of the ConflictDAG.
-	options *optionsConflictDAG
+	optsMergeToMaster bool
+}
 
-	// RWMutex is a mutex that prevents that two processes simultaneously update the ConfirmationState.
-	sync.RWMutex
+// New is the constructor for the BlockDAG and creates a new BlockDAG instance.
+func New[ConflictIDType, ResourceIDType comparable](evictionState *eviction.State, opts ...options.Option[ConflictDAG[ConflictIDType, ResourceIDType]]) (c *ConflictDAG[ConflictIDType, ResourceIDType]) {
```
piotrm50:
```suggestion
func New[ConflictIDType, ResourceIDType comparable](opts ...options.Option[ConflictDAG[ConflictIDType, ResourceIDType]]) (c *ConflictDAG[ConflictIDType, ResourceIDType]) {
```

packages/protocol/engine/consensus/conflictresolver/conflictresolver.go
```diff
@@ -69,53 +75,55 @@ func (o *ConflictResolver) ConflictLiked(conflictID utxo.TransactionID) (conflic
...
+		return
+	}
 
-		if preferred = !o.dislikedConnectedConflictingConflicts(conflictID).Has(conflictID); preferred {
-			for it := conflict.Parents().Iterator(); it.HasNext(); {
-				likeWalker.Push(it.Next())
+	if preferred = !o.dislikedConnectedConflictingConflicts(conflict).Has(conflict.ID()); preferred {
+		for it := conflict.Parents().Iterator(); it.HasNext(); {
+			parentConflict, exists := o.conflictDAG.Conflict(it.Next())
+			if !exists {
```
jcburuchaga, line 121: why not less code?

`if exists {
				likeWalker.Push(parentConflict)
			}
			`

packages/core/votes/conflicttracker/conflicttracker.go
```diff
@@ -56,40 +56,56 @@ func (c *ConflictTracker[ConflictIDType, ResourceIDType, VotePowerType]) Voters(
...
 	if !c.voterSupportsAllConflicts(voterID, parentConflictIDs) {
 		return
 	}
 
 	vote := votes.NewVote[ConflictIDType](voterID, power, votes.Like).WithConflictID(forkedConflictID)
-
 	votesObj, _ := c.votes.RetrieveOrCreate(forkedConflictID, votes.NewVotes[ConflictIDType, VotePowerType])
 	if added, opinionChanged := votesObj.Add(vote); added && opinionChanged {
-		c.Events.VoterAdded.Trigger(&VoterEvent[ConflictIDType]{Voter: voterID, ConflictID: forkedConflictID})
+		c.Events.VoterAdded.Trigger(&VoterEvent[ConflictIDType]{BlockID: fmt.Sprint(power), Voter: voterID, ConflictID: forkedConflictID})
```
jonastheis: this is only debug no?

packages/core/votes/conflicttracker/events.go
```diff
@@ -24,6 +24,7 @@ func NewEvents[ConflictIDType comparable](optLinkTargets ...*Events[ConflictIDTy
 }
 
 type VoterEvent[ConflictIDType comparable] struct {
+	BlockID    string
```
jonastheis: not needed anymore

packages/protocol/engine/notarization/events.go
```diff
@@ -9,17 +9,19 @@ import (
 
 // Events is a container that acts as a dictionary for the events of the notarization manager.
 type Events struct {
-	EpochCommitted *event.Linkable[*EpochCommittedDetails]
-	Error          *event.Linkable[error]
+	EpochCommitted  *event.Linkable[*EpochCommittedDetails]
+	ConflictDropped *event.Linkable[models.BlockID]
```
jonastheis: If we keep this we should also add the conflictID

packages/protocol/ledger/ledger.go
```diff
@@ -72,14 +74,16 @@ type Ledger struct {
 	optsConsumerCacheTime time.Duration
 
 	// optConflictDAG contains the optionsLedger for the ConflictDAG.
-	optConflictDAG []conflictdag.Option
+	optConflictDAG []options.Option[conflictdag.ConflictDAG[utxo.TransactionID, utxo.OutputID]]
 
 	// mutex is a DAGMutex that is used to make the Ledger thread safe.
 	mutex *syncutils.DAGMutex[utxo.TransactionID]
+
+	propagationMutex sync.RWMutex
```
jonastheis: Do we need this lock? Shouldn't the DAG mutex be enough?

packages/protocol/ledger/ledger.go
```diff
@@ -107,15 +108,17 @@ func New(chainStorage *storage.Storage, opts ...options.Option[Ledger]) (ledger
...
 
-	ledger.ConflictDAG.Events.ConflictAccepted.Attach(event.NewClosure(ledger.propagateAcceptanceToIncludedTransactions))
+	ledger.ConflictDAG.Events.ConflictAccepted.Attach(event.NewClosure(func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
+		ledger.PropagateAcceptanceToIncludedTransactions(conflict.ID())
+	}))
 
 	ledger.ConflictDAG.Events.ConflictRejected.Attach(event.NewClosure(ledger.propagatedRejectionToTransactions))
 
-	ledger.Events.TransactionBooked.Attach(event.NewClosure(func(event *TransactionBookedEvent) {
+	ledger.Events.TransactionBooked.Hook(event.NewClosure(func(event *TransactionBookedEvent) {
```
jonastheis: Do we need/want to hook here?

packages/protocol/ledger/ledger.go
```diff
@@ -107,15 +108,17 @@ func New(chainStorage *storage.Storage, opts ...options.Option[Ledger]) (ledger
...
 		ledger.processConsumingTransactions(event.Outputs.IDs())
 	}))
 
-	ledger.Events.TransactionInvalid.Attach(event.NewClosure(func(event *TransactionInvalidEvent) {
+	ledger.Events.TransactionInvalid.Hook(event.NewClosure(func(event *TransactionInvalidEvent) {
```
jonastheis: Do we need/want to hook here?

packages/protocol/ledger/ledger.go
```diff
@@ -269,20 +297,34 @@ func (l *Ledger) triggerRejectedEvent(txMetadata *TransactionMetadata) (triggere
...
+	defer l.propagationMutex.Unlock()
+
+	l.propagateAcceptanceToIncludedTransactions(txID)
+}
+
 // propagateAcceptanceToIncludedTransactions propagates confirmations to the included future cone of the given
 // Transaction.
 func (l *Ledger) propagateAcceptanceToIncludedTransactions(txID utxo.TransactionID) {
+	l.mutex.RLock(txID)
+	defer l.mutex.RUnlock(txID)
```
jonastheis: Why `RLock`? Shouldn't we `Lock` here as we modify the transactions?

packages/protocol/ledger/ledger.go
```diff
@@ -269,20 +297,34 @@ func (l *Ledger) triggerRejectedEvent(txMetadata *TransactionMetadata) (triggere
...
+
 	l.Storage.CachedTransactionMetadata(txID).Consume(func(txMetadata *TransactionMetadata) {
 		if !l.triggerAcceptedEvent(txMetadata) {
 			return
 		}
 
 		l.Utils.WalkConsumingTransactionMetadata(txMetadata.OutputIDs(), func(consumingTxMetadata *TransactionMetadata, walker *walker.Walker[utxo.OutputID]) {
-			if l.ConflictDAG.ConfirmationState(consumingTxMetadata.ConflictIDs()).IsAccepted() {
+			l.mutex.RLock(consumingTxMetadata.ID())
+			defer l.mutex.RUnlock(consumingTxMetadata.ID())
```
jonastheis: Why `RLock`? Shouldn't we `Lock` here as we modify the transactions?

packages/protocol/ledger/ledger.go
```diff
@@ -293,13 +335,16 @@ func (l *Ledger) propagateAcceptanceToIncludedTransactions(txID utxo.Transaction
 
 // propagateConfirmedConflictToIncludedTransactions propagates confirmations to the included future cone of the given
 // Transaction.
-func (l *Ledger) propagatedRejectionToTransactions(txID utxo.TransactionID) {
-	l.Storage.CachedTransactionMetadata(txID).Consume(func(txMetadata *TransactionMetadata) {
+func (l *Ledger) propagatedRejectionToTransactions(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
+	l.Storage.CachedTransactionMetadata(conflict.ID()).Consume(func(txMetadata *TransactionMetadata) {
```
jonastheis, line 324: We should lock the outer one too no?

packages/protocol/ledger/utils.go
```diff
@@ -35,17 +35,25 @@ func (u *Utils) ConflictIDsInFutureCone(conflictIDs utxo.TransactionIDs) (confli
 
 		if u.ledger.ConflictDAG.ConfirmationState(set.NewAdvancedSet(conflictID)).IsAccepted() {
 			u.ledger.Storage.CachedTransactionMetadata(conflictID).Consume(func(txMetadata *TransactionMetadata) {
-				u.WalkConsumingTransactionMetadata(txMetadata.OutputIDs(), func(txMetadata *TransactionMetadata, walker *walker.Walker[utxo.OutputID]) {
-					conflictIDsInFutureCone.AddAll(txMetadata.ConflictIDs())
+				u.WalkConsumingTransactionMetadata(txMetadata.OutputIDs(), func(consumingTxMetadata *TransactionMetadata, walker *walker.Walker[utxo.OutputID]) {
+					u.ledger.mutex.RLock(consumingTxMetadata.ID())
+					defer u.ledger.mutex.RUnlock(consumingTxMetadata.ID())
```
jonastheis, line 41: Is this correct? Don't we double lock here and where we use it?

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -63,12 +67,48 @@ func (r *ReferenceProvider) References(payload payload.Payload, strongParents mo
...
+	blockIDs = models.NewBlockIDs()
+	if amount == 0 {
+		return blockIDs
+	}
+
+	for it := r.protocol.TipManager.TipsConflictTracker.MissingConflicts(amount).Iterator(); it.HasNext(); {
+		conflictID := it.Next()
+
+		// TODO: make sure that timestamp monotonicity is not broken
+		attachment, err := r.latestValidAttachment(conflictID)
```
karimodm, line 90: latestValidAttachment should return a block that is not above our wall clock

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -98,6 +138,10 @@ func (r *ReferenceProvider) weakParentsFromUnacceptedInputs(payload payload.Payl
 				continue
 			}
 
+			if latestAttachment.Commitment().Index() > r.protocol.Engine().Storage.Settings.LatestCommitment().Index() {
```
karimodm: We should be able to remove this due to the futureBlocks buffer.

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -122,14 +167,32 @@ func (r *ReferenceProvider) addedReferencesForBlock(blockID models.BlockID, excl
...
+		fmt.Println(">> error adding references", blockID)
 		if schedulerBlock, schedulerBlockExists := r.protocol.CongestionControl.Scheduler().Block(blockID); schedulerBlockExists {
+			fmt.Println(">> tip removed", blockID)
 			r.protocol.TipManager.DeleteTip(schedulerBlock)
 		}
 		return nil, false
 	}
 
+	// We could not refer to any block to fix the opinion, so we add the tips' strong parents to the tip pool.
+	if addedReferences == nil {
```
karimodm, line 173: I think this could not happen anymore, as the only codepath to have this nil is when the conflict we like has the latest attachment as a future block (>= latest commitment). We now prevent to even see such block as an attachment, so we won't invalidate this tip and we won't therefore need to add its parents.

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
 		if adjust, referencedBlk, referenceErr := r.adjustOpinion(conflictID, excludedConflictIDs); referenceErr != nil {
 			return nil, errors.Wrapf(referenceErr, "failed to create reference for %s", conflictID)
 		} else if adjust {
-			referencesToAdd.Add(models.ShallowLikeParentType, referencedBlk)
+			if referencedBlk != models.EmptyBlockID {
+				referencesToAdd.Add(models.ShallowLikeParentType, referencedBlk)
+			} else {
+				// We could not find a block that we could reference to fix this strong parent, but we don't want to delete the tip.
```
karimodm, line 217: should never happen anymore, see above comment

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
+	// TODO: make sure that timestamp monotonicity is held
+	if err != nil {
+		return false, models.EmptyBlockID, err
+	}
 
-		// only walk deeper if we don't like "something else"
-		engineInstance.Ledger.ConflictDAG.Storage.CachedConflict(currentConflictID).Consume(func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
-			w.PushFront(conflict.Parents().Slice()...)
-		})
+	// Check if the attachment has a monotonic commitment.
```
karimodm: Remove

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
+	}
 
-			excludedConflictIDs.AddAll(engineInstance.Ledger.Utils.ConflictIDsInFutureCone(dislikedConflictIDs))
+	if likedConflictID == conflictID {
+		return false, models.EmptyBlockID, nil
+	}
 
-			return true, blkID, nil
-		}
+	attachment, err := r.latestValidAttachment(likedConflictID)
```
karimodm, line 240: latestValidAttachment should check for monotonicity against our wall clock

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
-		if likedConflictID, dislikedConflictIDs := engineInstance.Consensus.ConflictResolver.LikedConflictMember(currentConflictID); !likedConflictID.IsEmpty() {
-			// only triggers in first iteration
-			if likedConflictID == conflictID {
-				return false, models.EmptyBlockID, nil
-			}
-
-			if blkID, err = r.firstValidAttachment(likedConflictID); err != nil {
-				continue
-			}
+	if likedConflictID.IsEmpty() {
```
karimodm, line 232: `AdjustOpinion` now walks the ConflictDAG to the parents, can this happen then?

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
+	block = r.protocol.Engine().Tangle.Booker.GetLatestAttachment(txID)
+	if block == nil {
+		return nil, errors.Errorf("could not obtain latest attachment for %s", txID)
+	}
+
+	if acceptedTime := r.protocol.Engine().Clock.AcceptedTime(); block.IssuingTime().Before(acceptedTime.Add(-r.timeSinceConfirmationThreshold)) {
+		return nil, errors.Errorf("attachment of %s with %s is too far in the past relative to AcceptedTime %s", txID, block.ID(), acceptedTime.String())
+	}
+
+	// if block.IsSubjectivelyInvalid() {
```
karimodm: Remove

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
+	// 		})
+	// 	}
+	//
+	// 	if block == nil {
+	// 		return nil, errors.Errorf("no attachment for conflict members of %s are valid", txID)
+	// 		// return nil, errors.Errorf("attachment of %s with %s is subjectively invalid", txID, bookerBlock.ID())
+	// 	}
+	// }
 
 	if committableEpoch := r.latestEpochIndexCallback(); block.ID().Index() <= committableEpoch {
```
karimodm, line 262: Remove

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -150,53 +219,111 @@ func (r *ReferenceProvider) addedReferencesForConflicts(conflictIDs utxo.Transac
...
-func (r *ReferenceProvider) firstValidAttachment(txID utxo.TransactionID) (blockID models.BlockID, err error) {
-	block := r.protocol.Engine().Tangle.Booker.GetEarliestAttachment(txID)
+// latestValidAttachment returns the first valid attachment of the given transaction.
+func (r *ReferenceProvider) latestValidAttachment(txID utxo.TransactionID) (block *booker.Block, err error) {
+	block = r.protocol.Engine().Tangle.Booker.GetLatestAttachment(txID)
+	if block == nil {
+		return nil, errors.Errorf("could not obtain latest attachment for %s", txID)
+	}
+
+	if acceptedTime := r.protocol.Engine().Clock.AcceptedTime(); block.IssuingTime().Before(acceptedTime.Add(-r.timeSinceConfirmationThreshold)) {
```
karimodm, line 258: Check for timestamp monotonicity and error if failed

packages/protocol/engine/consensus/blockgadget/gadget.go
```diff
@@ -243,9 +248,25 @@ func (a *Gadget) setup() {
 	event.AttachWithWorkerPool(a.tangle.VirtualVoting.Events.SequenceTracker.VotersUpdated, func(evt *sequencetracker.VoterUpdatedEvent) {
 		a.RefreshSequence(evt.SequenceID, evt.NewMaxSupportedIndex, evt.PrevMaxSupportedIndex)
 	}, wp)
-	event.AttachWithWorkerPool(a.tangle.VirtualVoting.Events.ConflictTracker.VoterAdded, func(evt *conflicttracker.VoterEvent[utxo.TransactionID]) {
+
+	event.Hook(a.tangle.VirtualVoting.Events.ConflictTracker.VoterAdded, func(evt *conflicttracker.VoterEvent[utxo.TransactionID]) {
 		a.RefreshConflictAcceptance(evt.ConflictID)
+
+		conflictID := evt.ConflictID
```
karimodm: A leftover of debug

packages/protocol/engine/consensus/blockgadget/gadget.go
```diff
@@ -243,9 +248,25 @@ func (a *Gadget) setup() {
...
+		_, exists := a.tangle.Booker.Ledger.ConflictDAG.Conflict(conflictID)
+		if !exists {
+			return
+		}
+	})
+
+	event.AttachWithWorkerPool(a.tangle.VirtualVoting.Events.ConflictTracker.VoterRemoved, func(evt *conflicttracker.VoterEvent[utxo.TransactionID]) {
```
karimodm: Leftover from debug

packages/protocol/engine/consensus/conflictresolver/conflictresolver.go
```diff
@@ -34,32 +34,63 @@ func New(conflictDAG *conflictdag.ConflictDAG[utxo.TransactionID, utxo.OutputID]
...
 		}
-
 		return true
 	})
 
-	return
+	return likedConflict, dislikedConflicts
+}
+
+// AdjustOpinion returns the reference that is necessary to correct our opinion on the given conflict.
```
karimodm, line 60:
```suggestion
// AdjustOpinion returns the reference that is necessary to correct our opinion on the given conflict.
// It recursively walk to the parent conflicts to find the upmost liked conflict.
```

packages/protocol/engine/consensus/conflictresolver/conflictresolver.go
```diff
@@ -69,72 +100,75 @@ func (o *ConflictResolver) ConflictLiked(conflictID utxo.TransactionID) (conflic
...
 
 // conflictPreferred returns whether the conflict is the winner across its conflict sets.
-func (o *ConflictResolver) conflictPreferred(conflictID utxo.TransactionID, likeWalker *walker.Walker[utxo.TransactionID]) (preferred bool) {
+func (o *ConflictResolver) conflictPreferred(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID], likeWalker *walker.Walker[*conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]]) (preferred bool) {
 	preferred = true
-	if conflictID == utxo.EmptyTransactionID {
+	if conflict.ID().IsEmpty() {
 		return
 	}
+	// TODO: this should never happen?
```
karimodm: What does this refer to?

packages/protocol/engine/engine.go
```diff
@@ -93,10 +91,10 @@ func New(
 			optsBootstrappedThreshold: 10 * time.Second,
 			optsSnapshotDepth:         5,
 		}, opts, func(e *Engine) {
+			e.EvictionState = eviction.NewState(storageInstance)
```
karimodm: Initialize above instead? We don't need the engine instance.

packages/protocol/engine/engine.go
```diff
@@ -333,12 +331,13 @@ func (e *Engine) initClock() {
 func (e *Engine) initTSCManager() {
 	e.TSCManager = tsc.New(e.Consensus.BlockGadget.IsBlockAccepted, e.Tangle, e.optsTSCManagerOptions...)
 
-	wp := e.Workers.CreatePool("TSCManager", 1) // Using just 1 worker to avoid contention
+	// wp := e.Workers.CreatePool("TSCManager", 1) // Using just 1 worker to avoid contention
```
karimodm, line 334: This is the orphanage stuff right? Should we remove it all as part of this PR? We can also get it back if we need, but at least we can get rid of all the orphanage-related events.

packages/protocol/engine/engine.go
```diff
@@ -375,34 +374,82 @@ func (e *Engine) initNotarizationManager() {
...
-		}
-	}, wpBlocks)
-
-	// TODO: add transaction orphaned event
 
 	// Epochs are committed whenever ATT advances, start committing only when bootstrapped.
 	event.AttachWithWorkerPool(e.Clock.Events.AcceptanceTimeUpdated, func(event *clock.TimeUpdateEvent) {
 		e.NotarizationManager.SetAcceptanceTime(event.NewTime)
 	}, wpCommitments)
+
```
karimodm: Remove orphanage stuff

packages/protocol/engine/ledgerstate/ledgerstate.go
```diff
@@ -173,3 +173,17 @@ func (l *LedgerState) onTransactionInclusionUpdated(inclusionUpdatedEvent *ledge
 		l.StateDiffs.moveTransactionToOtherEpoch(inclusionUpdatedEvent.TransactionMetadata, inclusionUpdatedEvent.PreviousInclusionEpoch, inclusionUpdatedEvent.InclusionEpoch)
 	}
 }
+
+func (l *LedgerState) onTransactionOrphaned(event *ledger.TransactionEvent) {
```
karimodm: Remove

packages/protocol/engine/notarization/events.go
karimodm: We should still support it and log it

packages/protocol/engine/notarization/manager.go
```diff
@@ -34,7 +33,7 @@ type Manager struct {
 
 	storage                  *storage.Storage
 	ledgerState              *ledgerstate.LedgerState
-	pendingConflictsCounters *shrinkingmap.ShrinkingMap[epoch.Index, uint64]
+	pendingConflictsCounters *memstorage.EpochStorage[models.BlockID, *models.Block]
```
karimodm: Not needed if we remove orphanage

packages/protocol/engine/notarization/manager.go
```diff
@@ -53,43 +52,45 @@ func NewManager(storageInstance *storage.Storage, ledgerState *ledgerstate.Ledge
...
-
-	if newCounter := lo.Return1(m.pendingConflictsCounters.Get(index)) - 1; newCounter != 0 {
-		m.pendingConflictsCounters.Set(index, newCounter)
-	} else {
-		m.pendingConflictsCounters.Delete(index)
-
-		m.tryCommitEpoch(index, m.AcceptanceTime())
-	}
-}
+// // AddConflictingAttachment adds the conflicting attachment to a set of pending conflicts of an epoch.
```
karimodm: remove all

packages/protocol/engine/notarization/manager.go
```diff
@@ -201,7 +209,9 @@ func (m *Manager) createCommitment(index epoch.Index) (success bool) {
 		return false
 	}
 
-	m.pendingConflictsCounters.Delete(index)
+	// TODO: remove
+	// This should have been already evicted by dropPendingConflicts.
+	// m.pendingConflictsCounters.Evict(index)
```
karimodm: Not needed

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -34,19 +37,30 @@ type BlockDAG struct {
 	// solidifier contains the solidifier instance used to determine the solidity of Blocks.
 	solidifier *causalorder.CausalOrder[models.BlockID, *Block]
 
+	// commitmentFunc is a function that returns the commitment corresponding to the given epoch index.
+	commitmentFunc func(index epoch.Index) (*commitment.Commitment, error)
+
+	// futureBlocks contains blocks with a commitment in the future, that should not be passed to the booker yet.
+	futureBlocks *memstorage.EpochStorage[commitment.ID, *set.AdvancedSet[*Block]]
+
+	lastFuturePromotedIndex epoch.Index
```
karimodm: Add some comments

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -130,6 +144,9 @@ func (b *BlockDAG) evictEpoch(index epoch.Index) {
 	b.evictionMutex.Lock()
 	defer b.evictionMutex.Unlock()
 
+	// We want to deal with the synchronous BlockSolid events in a separate goroutine.
+	b.promoteFutureBlocksUntil(index)
```
karimodm: Spawn this process in a workerpool to prevent contention

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -140,11 +157,61 @@ func (b *BlockDAG) markSolid(block *Block) (err error) {
...
+				blocksStorage.ForEach(func(block *Block) (err error) {
+					b.Events.BlockSolid.Trigger(block)
+					return nil
+				})
+			}
+		}
+		b.futureBlocks.Evict(i)
+	}
+
+	b.lastFuturePromotedIndex = index + 1
```
karimodm: Maybe rename since we are adding +1? Is actually the next epoch we will promote

packages/protocol/engine/tangle/booker/attachments.go
```diff
@@ -1,36 +1,111 @@
...
 )
 
 type attachments struct {
-	attachments *memstorage.Storage[utxo.TransactionID, *memstorage.Storage[epoch.Index, *LockableSlice[*Block]]]
+	attachments *memstorage.Storage[utxo.TransactionID, *memstorage.Storage[epoch.Index, *memstorage.Storage[models.BlockID, *Block]]]
 	evictionMap *memstorage.Storage[epoch.Index, set.Set[utxo.TransactionID]]
+
+	// nonOrphanedCounter is used to count all non-orphaned attachment of a transaction,
+	// so that it's not necessary to iterate through all the attachments to check if the transaction is orphaned.
+	nonOrphanedCounter *memstorage.Storage[utxo.TransactionID, uint32]
```
karimodm, line 23: Still needed?

packages/protocol/engine/tangle/booker/attachments.go
```diff
@@ -1,36 +1,111 @@
...
+
+	prevValue, _ := a.nonOrphanedCounter.RetrieveOrCreate(txID, func() uint32 { return 0 })
+	a.nonOrphanedCounter.Set(txID, prevValue+1)
+
 	a.updateEvictionMap(block.ID().EpochIndex, txID)
+
+	return true
+}
+
+func (a *attachments) AttachmentOrphaned(txID utxo.TransactionID, block *Block) (attachmentBlock *Block, attachmentOrphaned, lastAttachmentOrphaned bool) {
```
karimodm, line 54: Remove

packages/protocol/engine/tangle/booker/attachments.go
```diff
@@ -1,36 +1,111 @@
...
+	}
+
+	return
+}
+
+func (a *attachments) GetAttachmentBlocks(txID utxo.TransactionID) (attachments *set.AdvancedSet[*Block]) {
+	attachments = set.NewAdvancedSet[*Block]()
+
+	if txStorage := a.storage(txID, false); txStorage != nil {
+		txStorage.ForEach(func(_ epoch.Index, blocks *memstorage.Storage[models.BlockID, *Block]) bool {
```
karimodm, line 104: Why we don't lock the `mutex` DAGMutex for reading but we do only for writing?

packages/protocol/engine/tangle/booker/attachments.go
```diff
@@ -43,57 +118,66 @@ func (a *attachments) Evict(epochIndex epoch.Index) {
...
+			})
 			return true
 		})
 	}
 
 	return
 }
 
-func (a *attachments) getLatestAttachment(txID utxo.TransactionID) (attachment *Block) {
+func (a *attachments) getLatestAttachment(txID utxo.TransactionID, returnOrphaned ...bool) (attachment *Block) {
```
karimodm: Check timestamp monotonicity for the attachment to be valid? The check should either be here or in `referenceProvider.latestValidAttachment`

karimodm: No need for the returnOrphaned argument anymore

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -434,6 +446,14 @@ func (b *Booker) setupEvents() {
 			panic(err)
 		}
 	})
+	event.Hook(b.BlockDAG.Events.BlockOrphaned, func(orphanedBlock *blockdag.Block) {
```
karimodm, line 460: nuke

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -450,61 +470,103 @@ func (b *Booker) setupEvents() {
 	}, b.workers.CreatePool("Booker", 2))
 }
 
+func (b *Booker) OrphanAttachment(block *Block) {
```
karimodm, line 484: nuke

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -450,61 +470,103 @@ func (b *Booker) setupEvents() {
...
+	for blockWalker.HasNext() {
+		block := blockWalker.Next()
+		propagateFurther, propagationErr := b.propagateToBlock(block, addedConflictID, removedConflictIDs)
+		if propagationErr != nil {
+			blockWalker.StopWalk()
+			return errors.Wrapf(propagationErr, "failed to propagate forked ConflictID %s to future cone of %s", addedConflictID, block.ID())
+		}
 		if propagateFurther {
-			blockWalker.PushAll(b.strongChildren(block)...)
+			// TODO: inherit weak and liked as well
```
karimodm: This is not true, right? Only for the first step but not after.

packages/protocol/protocol_test.go
```diff
@@ -1007,3 +1008,141 @@ func TestProtocol_EngineSwitching(t *testing.T) {
 		node1.AssertEqualChainsAtLeastAtEpoch(9, node4)
 	}
 }
+
+func TestEngine_GuavaConflict(t *testing.T) {
```
piotrm50, line 1012: Do we want to keep this as a test?

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -450,61 +470,103 @@ func (b *Booker) setupEvents() {
...
 	propagated = b.updateBlockConflicts(block, addedConflictID, removedConflictIDs)
 	// We only need to propagate further (in the block's future cone) if the block was updated.
 	return propagated, propagated, nil
 }
 
 func (b *Booker) updateBlockConflicts(block *Block, addedConflict utxo.TransactionID, parentConflicts utxo.TransactionIDs) (updated bool) {
-	if _, conflictIDs := b.blockBookingDetails(block); !conflictIDs.HasAll(parentConflicts) {
+	_, conflictIDs := b.blockBookingDetails(block)
+
+	if !conflictIDs.HasAll(parentConflicts) {
```
karimodm, line 579: Add a comment on this cryptic check

packages/protocol/engine/tangle/booker/events.go
```diff
@@ -9,7 +9,9 @@ import (
 )
 
 type Events struct {
-	BlockBooked         *event.Linkable[*Block]
+	BlockBooked         *event.Linkable[*BlockBookedEvent]
+	AttachmentCreated   *event.Linkable[*Block]
+	AttachmentOrphaned  *event.Linkable[*Block]
```
karimodm, line 14: Remove

packages/protocol/engine/tangle/virtualvoting/virtualvoting.go
```diff
@@ -42,10 +43,11 @@ type VirtualVoting struct {
 
 func New(workers *workerpool.Group, booker *booker.Booker, validators *sybilprotection.WeightedSet, opts ...options.Option[VirtualVoting]) (newVirtualVoting *VirtualVoting) {
 	return options.Apply(&VirtualVoting{
-		Validators: validators,
-		blocks:     memstorage.NewEpochStorage[models.BlockID, *Block](),
-		Booker:     booker,
+		Validators:    validators,
+		blocks:        memstorage.NewEpochStorage[models.BlockID, *Block](),
+		Booker:        booker,
 		Workers:    workers,
```
karimodm: File not properly formatted.

karimodm (changes requested):

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -139,53 +190,112 @@ func (b *ConflictDAG[ConflictIDType, ConflictingResourceID]) UnconfirmedConflict
...
-			})
+		conflict.ForEachConflictingConflict(func(conflictingConflict *Conflict[ConflictIDType, ResourceIDType]) bool {
+			conflictsToReject.Add(conflictingConflict)
+			return true
 		})
 	}
 
+	modified = c.rejectConflictsWithFutureCone(conflictsToReject) || modified
+
+	// // Delete all resolved ConflictSets (don't have a pending conflict anymore).
```
piotrm50, line 226: Delete this code as we'll not be evicting conflits upon resolution.

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -139,53 +190,112 @@ func (b *ConflictDAG[ConflictIDType, ConflictingResourceID]) UnconfirmedConflict
...
+		if confirmationState = confirmationState.Aggregate(c.confirmationState(conflictID)); confirmationState.IsRejected() {
+			return confirmation.Rejected
+		}
+	}
+
+	return confirmationState
+}
+
+// ConfirmationStateUnsafe returns the ConfirmationState of the given ConflictIDs.
+func (c *ConflictDAG[ConflictIDType, ResourceIDType]) ConfirmationStateUnsafe(conflictIDs *set.AdvancedSet[ConflictIDType]) (confirmationState confirmation.State) {
```
piotrm50: Remove this as it's not used anywhere anymore.

packages/protocol/engine/engine.go
piotrm50: Yeah, we could , but are we sure that we won't need those for transaction orphanage? I think we haven't concluded on what to do with it yet.

packages/protocol/engine/tangle/virtualvoting/virtualvoting.go
```diff
@@ -211,13 +213,22 @@ func (o *VirtualVoting) EvictEpochTracker(epochIndex epoch.Index) {
...
+	if !exists || block.IsSubjectivelyInvalid() {
+		return
+	}
+
 	o.conflictTracker.AddSupportToForkedConflict(forkedConflictID, parentConflictIDs, block.IssuerID(), votePower)
 }
 
 // take everything in future cone because it was not conflicting before and move to new conflict.
 func (o *VirtualVoting) processForkedMarker(marker markers.Marker, forkedConflictID utxo.TransactionID, parentConflictIDs utxo.TransactionIDs) {
+	fmt.Println("Begin processing forked marker", marker)
```
piotrm50:
```suggestion
```

packages/protocol/engine/tangle/virtualvoting/virtualvoting.go
```diff
@@ -1,11 +1,12 @@
 package virtualvoting
 
 import (
-	"sync"
+	"fmt"
```
piotrm50:
```suggestion
```

piotrm50:
```suggestion
```

packages/protocol/engine/tangle/booker/booker.go
piotrm50: exactly, for the first step. this should be removed.

piotrm50 (commented):
review in progress

packages/protocol/engine/tangle/booker/booker.go
jonastheis:
```suggestion
```

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -174,17 +175,19 @@ func (b *Booker) BlockFloor(marker markers.Marker) (floorMarker markers.Marker,
 }
 
 // GetEarliestAttachment returns the earliest attachment for a given transaction ID.
-func (b *Booker) GetEarliestAttachment(txID utxo.TransactionID) (attachment *Block) {
-	return b.attachments.getEarliestAttachment(txID)
+// returnOrphaned parameter specifies whether the returned attachment may be orphaned.
+func (b *Booker) GetEarliestAttachment(txID utxo.TransactionID, returnOrphaned ...bool) (attachment *Block) {
```
karimodm: We don't need returnOrphaned anymore

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -174,17 +175,19 @@ func (b *Booker) BlockFloor(marker markers.Marker) (floorMarker markers.Marker,
...
+	return b.attachments.getEarliestAttachment(txID, returnOrphaned...)
 }
 
 // GetLatestAttachment returns the latest attachment for a given transaction ID.
-func (b *Booker) GetLatestAttachment(txID utxo.TransactionID) (attachment *Block) {
-	return b.attachments.getLatestAttachment(txID)
+// returnOrphaned parameter specifies whether the returned attachment may be orphaned.
+func (b *Booker) GetLatestAttachment(txID utxo.TransactionID, returnOrphaned ...bool) (attachment *Block) {
```
karimodm: We don't need returnOrphaned anymore

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -229,31 +234,39 @@ func (b *Booker) block(id models.BlockID) (block *Block, exists bool) {
 	return storage.Get(id)
 }
 
-func (b *Booker) book(block *Block) (err error) {
+func (b *Booker) book(block *Block) (inheritingErr error) {
+	b.bookingMutex.Lock(block.ID())
```
karimodm, line 241: Add comment to explain that this is supposed to cover the synchronous execution of the "BlockBooked" event from the virtuavoting to add votes while mutually excluding the forking process.

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -450,61 +470,102 @@ func (b *Booker) setupEvents() {
...
 
-		updated, propagateFurther, forkErr := b.propagateForkedConflict(block, addedConflictID, removedConflictIDs)
-		if forkErr != nil {
-			blockWalker.StopWalk()
-			return errors.Wrapf(forkErr, "failed to propagate forked ConflictID %s to future cone of %s", addedConflictID, block.ID())
-		}
-		if !updated {
-			continue
-		}
+	for it := b.GetAllAttachments(transactionID).Iterator(); it.HasNext(); {
```
karimodm, line 506: Add a comment explaining the propagation to weak and liked for the first level of children.

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -450,61 +470,102 @@ func (b *Booker) setupEvents() {
...
+			blockWalker.PushAll(b.blocksFromBlockDAGBlocks(block.StrongChildren())...)
 		}
 	}
+
 	return nil
 }
 
-func (b *Booker) propagateForkedConflict(block *Block, addedConflictID utxo.TransactionID, removedConflictIDs utxo.TransactionIDs) (propagated, propagateFurther bool, err error) {
+func (b *Booker) propagateToBlock(block *Block, addedConflictID utxo.TransactionID, removedConflictIDs utxo.TransactionIDs) (propagateFurther bool, err error) {
 	b.bookingMutex.Lock(block.ID())
```
karimodm, line 535: Add a comment to explain the mutual exclusion between the `book` -> virtualvoting flow and this `propagateToBlock` -> virtualvoting `processForkedBlock`.
We have to be careful when uncommenting the propagation through markers: we might need something similar there as well to prevent race conditions.

packages/protocol/engine/tsc/manager.go
```diff
@@ -2,7 +2,6 @@ package tsc
```
karimodm, line 2: Remove the whole thing :smiley: ?

packages/protocol/ledger/conflictdag/conflictdag.go
karimodm: Can we maybe salvage it for eviction? We are not going to implement that now, but just not to rewrite this code.

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -139,53 +190,112 @@ func (b *ConflictDAG[ConflictIDType, ConflictingResourceID]) UnconfirmedConflict
...
-func (b *ConflictDAG[ConflictID, ConflictingResourceID]) ConfirmationState(conflictIDs *set.AdvancedSet[ConflictID]) (confirmationState confirmation.State) {
-	b.RLock()
-	defer b.RUnlock()
+func (c *ConflictDAG[ConflictIDType, ResourceIDType]) ConfirmationState(conflictIDs *set.AdvancedSet[ConflictIDType]) (confirmationState confirmation.State) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+
+	// TODO: simplify this method
+
+	confirmationState = confirmation.Confirmed
```
karimodm: Why defaulting to Confirmed instead of Undefined? If the conflict set received is empty you would return Confirmed.

packages/protocol/engine/tangle/booker/attachments.go
piotrm50: txStorage has a lock itself. We lock when writing because we had to store it and update the counters atomically.

packages/protocol/engine/tangle/blockdag/blockdag.go
```diff
@@ -167,6 +242,7 @@ func (b *BlockDAG) checkParents(block *Block) (err error) {
 }
 
 func (b *BlockDAG) markInvalid(block *Block, reason error) {
+	fmt.Println("marking block invalid", block.ID(), reason)
```
piotrm50: remove prints

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -233,99 +349,196 @@ func (b *ConflictDAG[ConflictID, ConflictingResourceID]) determineConflictsToAdd
...
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+
+	c.conflicts.ForEach(func(c2 ConflictIDType, conflict *Conflict[ConflictIDType, ResourceIDType]) bool {
+		consumer(conflict)
+		return true
 	})
+}
+
+func (c *ConflictDAG[ConflictIDType, ResourceIDType]) HandleOrphanedConflict(conflictID ConflictIDType) {
```
karimodm, line 469: Do we still need this?

packages/protocol/ledger/conflictdag/conflictdag.go
```diff
@@ -233,99 +349,196 @@ func (b *ConflictDAG[ConflictID, ConflictingResourceID]) determineConflictsToAdd
...
+		}
+	}
 }
 
-// registerConflictMember registers a Conflict in a Conflict by creating the references (if necessary) and increasing the
-// corresponding member counter.
-func (b *ConflictDAG[ConflictID, ConflictingResourceID]) registerConflictMember(resourceID ConflictingResourceID, conflictID ConflictID) {
-	b.Storage.CachedConflictMember(resourceID, conflictID, NewConflictMember[ConflictingResourceID, ConflictID]).Release()
+// getLastPendingConflict returns last pending Conflict from the ConflictSet or returns nil if zero or more than one pending conflicts left.
+func (c *ConflictDAG[ConflictIDType, ResourceIDType]) getLastPendingConflict(conflictSet *ConflictSet[ConflictIDType, ResourceIDType]) (pendingConflict *Conflict[ConflictIDType, ResourceIDType]) {
```
karimodm, line 508: If the above is removed this should go too.

packages/protocol/ledger/ledger.go
```diff
@@ -110,14 +107,16 @@ func New(workerPool *workerpool.UnboundedWorkerPool, chainStorage *storage.Stora
 	ledger.dataFlow = newDataFlow(ledger)
 	ledger.Utils = newUtils(ledger)
 
-	event.AttachWithWorkerPool(ledger.ConflictDAG.Events.ConflictAccepted, ledger.propagateAcceptanceToIncludedTransactions, workerPool)
+	event.AttachWithWorkerPool(ledger.ConflictDAG.Events.ConflictAccepted, func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
```
karimodm, line 111: Can these be racey if done asynchrnously?

packages/protocol/ledger/ledger.go
karimodm: It should not be needed: it is pretty much the same as `propagateAcceptanceToIncludedTransactions` above. We only lock this TX for writing when we set its state in the `triggerRejectedEventLocked` . 
But we have to make sure again that the booking with rejection does not race with a fork with rejection.

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -122,14 +167,32 @@ func (r *ReferenceProvider) addedReferencesForBlock(blockID models.BlockID, excl
...
+				return true
+			})
+		}
+		fmt.Println(">> could not fix opinion", blockID)
+		return nil, false
+	}
+
 	// A block might introduce too many references and cannot be picked up as a strong parent.
 	if _, success = r.tryExtendReferences(models.NewParentBlockIDs(), addedReferences); !success {
+		fmt.Println(">> tryExtendedReferences", blockID)
```
piotrm50: remove print

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -122,14 +167,32 @@ func (r *ReferenceProvider) addedReferencesForBlock(blockID models.BlockID, excl
...
+	if addedReferences == nil {
+		if block, exists := r.protocol.Engine().Tangle.Booker.Block(blockID); exists {
+			block.ForEachParentByType(models.StrongParentType, func(parentBlockID models.BlockID) bool {
+				if schedulerBlock, schedulerBlockExists := r.protocol.CongestionControl.Scheduler().Block(parentBlockID); schedulerBlockExists {
+					r.protocol.TipManager.AddTipNonMonotonic(schedulerBlock)
+				}
+				return true
+			})
+		}
+		fmt.Println(">> could not fix opinion", blockID)
```
piotrm50, line 182: remove print

plugins/p2p/p2p.go
```diff
@@ -21,25 +21,25 @@ func createManager(lPeer *peer.Local) *p2p.Manager {
 	// resolve the bind address
 	localAddr, err = net.ResolveTCPAddr("tcp", Parameters.BindAddress)
 	if err != nil {
-		Plugin.LogFatalfAndExit("bind address '%s' is invalid: %s", Parameters.BindAddress, err)
+		Plugin.LogFatalfAndExitf("bind address '%s' is invalid: %s", Parameters.BindAddress, err)
```
piotrm50, line 24: I think this should not be marked as change in our PR, as we merged this from develop, no?

packages/protocol/ledger/storage.go
```diff
@@ -258,32 +258,58 @@ func (s *Storage) initConsumers(outputIDs utxo.OutputIDs, txID utxo.TransactionI
 // pruneTransaction removes a Transaction (and all of its dependencies) from the database.
 func (s *Storage) pruneTransaction(txID utxo.TransactionID, pruneFutureCone bool) {
```
karimodm, line 259: This entire method is currently only used upon orphanage, so it should be removed. We could although use it on pruning: as currently we store in the ledger storage every output that ever existed, regardless of the fact that it was already spent by a (now pruned) transaction.

packages/protocol/ledger/storage.go
```diff
@@ -258,32 +258,58 @@ func (s *Storage) initConsumers(outputIDs utxo.OutputIDs, txID utxo.TransactionI
...
 		currentTxID := futureConeWalker.Next()
 
 		s.CachedTransactionMetadata(currentTxID).Consume(func(txMetadata *TransactionMetadata) {
 			s.CachedTransaction(currentTxID).Consume(func(tx utxo.Transaction) {
 				for it := s.ledger.Utils.ResolveInputs(tx.Inputs()).Iterator(); it.HasNext(); {
-					s.consumerStorage.Delete(byteutils.ConcatBytes(lo.PanicOnErr(it.Next().Bytes()), lo.PanicOnErr(currentTxID.Bytes())))
-				}
+					inputID := it.Next()
+
+					s.CachedOutput(inputID).Consume(func(output utxo.Output) {
```
karimodm, line 270: If we want to keep this method there is not need to collect stuff anymore to trigger the orphaned event, but just delete the inputs.

packages/protocol/ledger/utils.go
```diff
@@ -35,17 +36,25 @@ func (u *Utils) ConflictIDsInFutureCone(conflictIDs utxo.TransactionIDs) (confli
...
 			continue
 		}
 
-		u.ledger.ConflictDAG.Utils.ForEachChildConflictID(conflictID, func(childConflictID utxo.TransactionID) {
-			conflictIDWalker.Push(childConflictID)
+		conflict, exists := u.ledger.ConflictDAG.Conflict(conflictID)
+		if !exists {
+			continue
+		}
+		_ = conflict.Children().ForEach(func(childConflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) (err error) {
```
karimodm, line 55: Can we maybe .Slice() and PushAll to the walker?

packages/protocol/protocol.go
```diff
@@ -209,10 +209,10 @@ func (p *Protocol) initChainManager() {
...
 
 func (p *Protocol) initTipManager() {
-	p.TipManager = tipmanager.New(p.CongestionControl.Block, p.optsTipManagerOptions...)
+	p.TipManager = tipmanager.New(p.Workers.CreateGroup("TipManager"), p.CongestionControl.Block, p.optsTipManagerOptions...)
 	p.Events.TipManager = p.TipManager.Events
 
-	wp := p.Workers.CreatePool("TipManager", 1) // Using just 1 worker to avoid contention
+	wp := p.Workers.CreatePool("TipManagerAttach", 1) // Using just 1 worker to avoid contention
 
 	event.Hook(p.Events.Engine.Tangle.BlockDAG.BlockOrphaned, func(block *blockdag.Block) {
```
karimodm, line 217: Remove

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -141,7 +147,7 @@ func (t *TipManager) RemoveStrongParents(block *models.Block) {
 // RemoveStrongParents removes all tips that are parents of the given block.
 func (t *TipManager) removeStrongParents(block *models.Block) {
 	block.ForEachParent(func(parent models.Parent) {
-		// TODO: reintroduce TipsConflictTracker
+		// TODO: we remove the tip, but we reintroduce the pending liked conflict through the earliest attachment.
```
karimodm: This bit is actually obsolete: the new conflicttracker rescues conflicts, not specific tips.

packages/protocol/tipmanager/tipsconflicttracker.go
```diff
@@ -1,127 +1,148 @@
...
+
+func (c *TipsConflictTracker) Setup() {
+	wp := c.workers.CreatePool("TipsConflictTracker", 1)
+	event.AttachWithWorkerPool(c.engine.Ledger.ConflictDAG.Events.ConflictAccepted, func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
+		c.deleteConflict(conflict.ID())
+	}, wp)
+	event.AttachWithWorkerPool(c.engine.Ledger.ConflictDAG.Events.ConflictRejected, func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
+		c.deleteConflict(conflict.ID())
+	}, wp)
+	event.AttachWithWorkerPool(c.engine.Ledger.ConflictDAG.Events.ConflictNotConflicting, func(conflict *conflictdag.Conflict[utxo.TransactionID, utxo.OutputID]) {
```
karimodm, line 47: Does this event even make sense without orphanage?

packages/protocol/engine/notarization/events.go
piotrm50: removed the event along with the ConflictCounter in NotarizationManager

packages/protocol/ledger/ledger.go
piotrm50: changed to attach

packages/app/blockissuer/blockfactory/referenceprovider.go
piotrm50: To be resolved later.

packages/app/blockissuer/blockfactory/referenceprovider.go
piotrm50: panic if that happens

karimodm (approved):

PR #2537, opened 2023-01-18
Bag of bugfixes

piotrm50:
# Description of change

This PR contains fixes to a number of small issues.

## Type of change

## Change checklist


client/evilwallet/evilwallet.go
```diff
@@ -191,7 +193,7 @@ func (e *EvilWallet) requestFaucetFunds(wallet *Wallet) (outputID utxo.OutputID,
 	if err = RateSetterSleep(clt, true); err != nil {
 		return
 	}
-	err = clt.BroadcastFaucetRequest(addr.Base58())
+	err = clt.BroadcastFaucetRequest(addr.Base58(), 12)
```
karimodm, line 196: This is the faucet request PoW requirement right? Better maybe put it in a const

packages/app/blockissuer/blockfactory/factory.go
```diff
@@ -113,6 +113,7 @@ func (f *Factory) createBlockWithPayload(p payload.Payload, references models.Pa
 		return nil, errors.Wrap(err, "signing failed")
 	}
 	block.SetSignature(signature)
+	block.InvalidateBytesCache()
```
karimodm: This is quite ugly. You form a block, then you sign it by computing the block's content hash and therefore you are hashing the block's bytes without yet its signature. This results in the block's bytecache to be the serialized version of the block without its signature.
I think the `sign` method should invalidate the bytes cache, not this method. Put a comment in there as an explaination.

tools/docker-network/config.docker.json
```diff
@@ -79,5 +79,10 @@
   },
   "manaInitializer": {
     "FaucetAPI": "http://faucet:8080"
+  },
+  "blockIssuer": {
+    "rateSetter": {
+      "mode": "disabled"
```
karimodm, line 85: This overrides the other PR https://github.com/iotaledger/goshimmer/pull/2540/files#diff-3543443f8fa8c99ddae5adfe4ef6ba6b7231d434ce983f6fe5e72bf3e12a8ebe in which the rateSetter.enabled param is removed, thus activating the default "deficit" rateSetter.
What should we do here? Running the "defaults" or disabling it?

tools/docker-network/config.feature.docker.json
```diff
@@ -79,6 +79,10 @@
   },
   "manaInitializer": {
     "FaucetAPI": "http://faucet:8080"
+  },
+  "blockIssuer": {
+    "rateSetter": {
+      "mode": "disabled"
```
karimodm, line 85: Same as above

karimodm (changes requested):

tools/docker-network/config.feature.docker.json
piotrm50: I think locally we should disable it now, as it causes some problems with EvilSpammer. Later we can re-enable when other issues are resolved and we can focus on fine-tuning the rate-setter.

tools/docker-network/config.docker.json
piotrm50: I think locally we should disable it now, as it causes some problems with EvilSpammer. Later we can re-enable when other issues are resolved and we can focus on fine-tuning the rate-setter.

client/evilwallet/evilwallet.go
piotrm50: Yeah, that's a temporary fix. Ideally it should be put into a parameter and passed from outside, but we wanted to have something quick. We can add it as a parameter in a different PR.

packages/app/blockissuer/blockfactory/factory.go
piotrm50: Yeah, we talked about it and I think it makes more sense to put it after we actually modify the object, rather than after generating the signature.

karimodm (approved):

PR #2541, opened 2023-01-23
Add Engine Switching

### LINKED ISSUE #2510: Chain switching
### Flow:
- [ ] We receive a block from a source including a commit we disagree upon (we know is forking).
- [ ] The block will need to be (TSC epochs + 1) epochs old w.r.t the tip of the chain we are on.
- [ ] The block will contain alongside the commitment a CW.
- [ ] We compare the claimed CW against on our CW at the same point of the chain.
- [ ] If the claimed CW is higher than our own chain we request the attestations from source.
- [ ] When we recieve the attestations, we obtain back the block sent from this source.
- [ ] We have to obtain the mana vector at the forking point.
- [ ] With this mana vector we have to compute the _valid_ weight of the received attestations.
- [ ] This computed _valid_ weight needs to be == the claimed CW.
- [ ] If != trigger `peerBlacklisted` event for a future firewall component.
- [ ] If == then we have to dump a snapshot corresponding to the forking point.
- [ ] Instantiate and start candidateEngine with obtained snapshot.
- [ ] (maybe not needed) Seed the candidateEngine's solidifier with the block that originated the switch.
- [ ] Wait until the candidateEngine is bootstrapped (no warpsync).
- [ ] Compare candidate engine's CW of its latest commitment >= than the main chain CW.
- [ ] If >= shutdown and remove main engine and promote candidate engine to main, relink events.
- [ ] Prune dead engine storage.

alexsporn:


packages/core/commitment/commitment.go
```diff
@@ -30,11 +30,11 @@ func New(index epoch.Index, prevID ID, rootsID types.Identifier, cumulativeWeigh
 		PrevID:           prevID,
 		RootsID:          rootsID,
 		CumulativeWeight: cumulativeWeight,
-	})
+	}, false) //TODO: fix the data-race in immutable so that we can enable byte caching
```
hmoog: don't forget to revert this

packages/core/commitment/commitment.go
```diff
@@ -30,11 +30,11 @@ func New(index epoch.Index, prevID ID, rootsID types.Identifier, cumulativeWeigh
...
 }
 
 func NewEmptyCommitment() (newCommitment *Commitment) {
-	return model.NewImmutable[Commitment](&commitment{})
+	return model.NewImmutable[Commitment](&commitment{}, false)
```
hmoog: don't forget to revert this

packages/network/manualpeering/manualpeering.go
```diff
@@ -366,7 +367,7 @@ func (m *Manager) changeNeighborStatus(neighbor *p2p.Neighbor, connStatus Connec
 }
 
 func (m *Manager) connectionDirection(peerPK ed25519.PublicKey) (ConnectionDirection, error) {
-	result := bytes.Compare(m.local.PublicKey().Bytes(), peerPK.Bytes())
+	result := bytes.Compare(lo.PanicOnErr(m.local.PublicKey().Bytes()), lo.PanicOnErr(peerPK.Bytes()))
```
hmoog: since we already return an error in the method, we should maybe not use lo.PanicOnError

packages/protocol/chainmanager/manager.go
```diff
@@ -103,6 +132,21 @@ func (m *Manager) registerCommitment(commitment *commitment.Commitment) (isSolid
 		m.Events.CommitmentMissing.Trigger(parentCommitment.ID())
 	}
 
+	chainCommitment, created := m.Commitment(commitment.ID(), true)
+
+	if !created {
```
hmoog: I think this should be behind the PublishCommitment further down the line and only emitted if we store the block

packages/protocol/engine/clock/timeupdate.go
```diff
@@ -34,6 +34,7 @@ func (c *timeUpdate) Update(now, newTime time.Time) (updated bool) {
 	// the local wall clock should never be before the accepted Time unless we are eclipsed by malicious actors or our
 	// own Time is clearly in the past
 	if now.Before(newTime) {
+		//TODO: a block with a future time panics this
```
hmoog: We most probably should have a filter that doesn't allow to process future blocks, so they can also not get solid / confirmed

packages/protocol/engine/consensus/epochgadget/gadget.go
```diff
@@ -43,6 +43,14 @@ func (g *Gadget) LastConfirmedEpoch() epoch.Index {
 	return g.lastConfirmedEpoch
 }
 
+func (g *Gadget) setLastConfirmedEpoch(i epoch.Index) {
+	g.Lock()
+	defer g.Unlock()
+
+	g.lastConfirmedEpoch = i
+	g.Events.EpochConfirmed.Trigger(i)
```
hmoog: the trigger should maybe happen outside the scope of the lock to prevent possible deadlocks

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -78,12 +79,24 @@ func New(schedulerBlockRetrieverFunc blockRetrieverFunc, opts ...options.Option[
...
+	t.walkerCache = memstorage.NewEpochStorage[models.BlockID, types.Empty]()
 	t.tips = randommap.New[models.BlockID, *scheduler.Block]()
+	t.futureTips = memstorage.NewEpochStorage[commitment.ID, *memstorage.Storage[models.BlockID, *scheduler.Block]]()
+
 	t.engine = engine
 	t.blockAcceptanceGadget = engine.Consensus.BlockGadget
 }
 
 func (t *TipManager) AddTip(block *scheduler.Block) {
+	t.tipsMutex.RLock()
```
hmoog: Is a readlock correct here?

tools/integration-tests/tester/framework/network.go
```diff
@@ -401,7 +403,7 @@ func (n *Network) createPeers(ctx context.Context, numPeers int, networkConfig C
 	if networkConfig.Autopeering {
 		conf.AutoPeering.Enabled = true
 		conf.AutoPeering.EntryNodes = []string{
-			fmt.Sprintf("%s@%s:%d", base58.Encode(n.entryNode.Identity.PublicKey().Bytes()), n.entryNode.Name(), peeringPort),
+			fmt.Sprintf("%s@%s:%d", base58.Encode(lo.PanicOnErr(n.entryNode.Identity.PublicKey().Bytes())), n.entryNode.Name(), peeringPort),
```
hmoog, line 406: return error in methid instead of lo.PanicOnError

hmoog (changes requested):
A few comments ...

plugins/peer/plugin.go
```diff
@@ -134,8 +134,13 @@
 	if err != nil {
 		return errors.Wrapf(err, "unable to retrieve private key from peer database")
 	}
+	prvKeyDBBytes, err := prvKeyDB.Bytes()
+	if err != nil {
+		return err
+	}
 	prvKeyCfg := ed25519.PrivateKeyFromSeed(cfgSeed)
-	if !bytes.Equal(prvKeyCfg.Bytes(), prvKeyDB.Bytes()) {
+	prvKeyCfgBytes, err := prvKeyCfg.Bytes()
```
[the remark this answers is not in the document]
alexsporn: Already fixed

hmoog (dismissed):

tools/genesis-snapshot/main.go
```diff
@@ -143,7 +155,9 @@ func init() {
 
 func diagnosticPrintSnapshotFromFile(filePath string) {
 	s := createTempStorage()
-	e := engine.New(s, dpos.NewProvider(), mana1.NewProvider())
+	defer s.Shutdown()
+
+	e := engine.New(workerpool.NewGroup("Diagnostics"), s, dpos.NewProvider(), mana1.NewProvider())
```
jonastheis, line 160: add `defer e.Shutdown()`?

tools/genesis-snapshot/main.go
```diff
@@ -93,7 +97,7 @@ func main() {
 	manaDistribution := createManaDistribution(totalTokensToPledge)
 	initialAttestationsSlice := createInitialAttestations()
 
-	snapshotcreator.CreateSnapshot(protocol.DatabaseVersion, snapshotFileName, genesisTokenAmount, genesisSeed, manaDistribution, initialAttestationsSlice)
+	snapshotcreator.CreateSnapshot(workerpool.NewGroup("CreateSnapshot"), protocol.DatabaseVersion, snapshotFileName, genesisTokenAmount, genesisSeed, manaDistribution, initialAttestationsSlice, new(devnetvm.VM))
```
jonastheis: Having this call and passing in a workerpool group on first sight seems to be a non-blocking call. Can we remove the workerpool parameter here as it seems to be not really used (except to instantiate an engine) within this function?

packages/core/snapshotcreator/snapshotcreator.go
```diff
@@ -42,26 +46,32 @@ func CreateSnapshot(databaseVersion database.Version, snapshotFileName string, g
...
 	}
 
+	// Create outputs for nodes
 	engineInstance.NotarizationManager.Attestations.SetLastCommittedEpoch(-1)
-	for nodeID, value := range nodesToPledge {
+	for nodePublicKey, value := range nodesToPledge {
 		// pledge to ID but send funds to random address
-		output, outputMetadata = createOutput(devnetvm.NewED25519Address(ed25519.GenerateKeyPair().PublicKey), value, nodeID, 0)
+		nodeID := identity.NewID(nodePublicKey)
+		output, outputMetadata := createOutput(ledgerVM, nodePublicKey, value, nodeID, 0)
```
jonastheis, line 68: comment is not correct anymore. the funds are not send to a random address/burned. also the comment of the function is wrong.

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -423,27 +429,25 @@ func (b *Booker) strongChildren(block *Block) []*Block {
...
+	event.AttachWithWorkerPool(b.Ledger.Events.TransactionBooked, func(e *ledger.TransactionBookedEvent) {
 		contextBlockID := models.BlockIDFromContext(e.Context)
 
 		for _, block := range b.attachments.Get(e.TransactionID) {
 			if contextBlockID != block.ID() {
 				b.bookingOrder.Queue(block)
 			}
 		}
-	}))
+	}, b.Workers.CreatePool("Booker"))
```
jonastheis: This means we create a new unbounded workerpool with `2*runtime.NumCPU()` right? Isn't that too many idle goroutines for "just" handling the `TransactionBooked` event?

packages/protocol/engine/tangle/booker/testframework.go
```diff
@@ -99,38 +103,41 @@ func (t *TestFramework) AssertBlockConflictsUpdateCount(blockConflictsUpdateCoun
 }
 
 func (t *TestFramework) setupEvents() {
-	t.Booker.Events.BlockBooked.Hook(event.NewClosure(func(metadata *Block) {
+	t.Instance.Events.BlockBooked.Hook(event.NewClosure(func(metadata *Block) {
```
jonastheis: Why not the new style to hook to events?

packages/protocol/engine/tangle/testframework.go
```diff
@@ -5,127 +5,64 @@ import (
...
 	}
 }
 
-// WithLedgerOptions sets the ledger options that are used to create the ledger that is used by the Tangle.
-func WithLedgerOptions(opts ...options.Option[ledger.Ledger]) options.Option[TestFramework] {
-	return func(t *TestFramework) {
-		if t.optsLedger != nil {
-			panic("using the TestFramework with Ledger and LedgerOptions simultaneously is not allowed")
-		}
+func NewDefaultTestFramework(t *testing.T, workers *workerpool.Group, optsTangle ...options.Option[Tangle]) *TestFramework {
```
jonastheis, line 56: Having different constructors and combining them together is much easier to read and much less prone to errors due to unexpected behavior because of some default option parameters. I like!

packages/protocol/engine/throughputquota/mana1/throughputquota.go
```diff
@@ -179,7 +183,8 @@ func (m *ThroughputQuota) init() {
 
 	m.TriggerInitialized()
 
-	m.engine.Ledger.Events.TransactionAccepted.Attach(event.NewClosure(func(event *ledger.TransactionEvent) {
+	wp := m.workers.CreatePool("ThroughputQuota")
```
jonastheis: Do we need `2*runtime.NumCPU()` workers? I'd suggest to validate all calls to `CreatePool` in this regard

packages/protocol/engine/consensus/blockgadget/gadget.go
```diff
@@ -233,15 +239,15 @@ func (a *Gadget) EvictUntil(index epoch.Index) {
 }
 
 func (a *Gadget) setup() {
-	a.tangle.VirtualVoting.Events.SequenceTracker.VotersUpdated.Attach(event.NewClosure(func(evt *sequencetracker.VoterUpdatedEvent) {
-		a.RefreshSequence(evt.SequenceID, evt.NewMaxSupportedIndex, evt.PrevMaxSupportedIndex)
-	}))
+	wp := a.workers.CreatePool("Gadget")
```
jonastheis: Is it necessary to have in total 3 worker pools with `2*runtime.NumCPU()` each for the block gadget?

packages/protocol/engine/consensus/blockgadget/gadget.go
```diff
@@ -314,9 +320,12 @@ func (a *Gadget) markAsAccepted(block *Block) (err error) {
 	}
 
 	if block.SetAccepted() {
+		if block.ID().String() == "BlockID(Block5)" {
+			fmt.Println("foo")
+		}
```
jonastheis: Remove debug print

packages/protocol/engine/consensus/blockgadget/testframework.go
```diff
@@ -11,91 +11,84 @@ import (
...
+	)
 }
 
 func (t *TestFramework) setupEvents() {
 	t.Gadget.Events.BlockAccepted.Hook(event.NewClosure(func(metadata *Block) {
 		if debug.GetEnabled() {
 			t.test.Logf("ACCEPTED: %s", metadata.ID())
+			if metadata.ID().String() == "BlockID(Block5)" {
+				_ = ""
+			}
```
jonastheis: remove debug print

packages/protocol/engine/consensus/epochgadget/gadget.go
```diff
@@ -43,27 +47,33 @@ func (g *Gadget) LastConfirmedEpoch() epoch.Index {
...
+
+	g.lastConfirmedEpoch = i
+}
+
 func (g *Gadget) setup() {
-	g.tangle.VirtualVoting.Events.EpochTracker.VotersUpdated.Attach(event.NewClosure(func(evt *epochtracker.VoterUpdatedEvent) {
+	event.AttachWithWorkerPool(g.tangle.VirtualVoting.Events.EpochTracker.VotersUpdated, func(evt *epochtracker.VoterUpdatedEvent) {
 		g.refreshEpochConfirmation(evt.PrevLatestEpochIndex, evt.NewLatestEpochIndex)
-	}))
+	}, g.workers.CreatePool("Refresh"))
```
jonastheis: `2*runtime.NumCPU()` necessary?

packages/protocol/ledger/ledger.go
```diff
@@ -189,10 +190,8 @@ func (l *Ledger) processTransaction(tx utxo.Transaction) (err error) {
 func (l *Ledger) processConsumingTransactions(outputIDs utxo.OutputIDs) {
 	for it := l.Utils.UnprocessedConsumingTransactions(outputIDs).Iterator(); it.HasNext(); {
 		txID := it.Next()
-		event.Loop.Submit(func() {
```
jonastheis, line 192: Is this correct in the data flow? I.e. coming from the Tangle we book a tx and don't want to continue booking `UnprocessedConsumingTransactions` in the same goroutine but instead return to the tangle and continue booking our block. I think it should be queued in `l.wokerPool`

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -78,12 +79,24 @@ func New(schedulerBlockRetrieverFunc blockRetrieverFunc, opts ...options.Option[
...
 	t.tips = randommap.New[models.BlockID, *scheduler.Block]()
+	t.futureTips = memstorage.NewEpochStorage[commitment.ID, *memstorage.Storage[models.BlockID, *scheduler.Block]]()
+
 	t.engine = engine
 	t.blockAcceptanceGadget = engine.Consensus.BlockGadget
 }
 
 func (t *TipManager) AddTip(block *scheduler.Block) {
+	t.tipsMutex.Lock()
+	defer t.tipsMutex.Unlock()
```
jonastheis: What is this lock for exactly? Seems like it's protecting the instances, which means it could always be a read lock and is probably a superset of the `evictionMutex`. Can these be combined?

packages/storage/utils/directory.go
```diff
@@ -29,4 +30,36 @@ func (d *Directory) Path(relativePathElements ...string) (path string) {
...
 
+func (d *Directory) RemoveSubdir(name string) error {
+	return os.RemoveAll(d.Path(name))
+}
+
+func (d *Directory) SubDirs() ([]string, error) {
+	var dirs []string
+	basePath := filepath.Base(d.path)
+
+	err := filepath.WalkDir(d.path, func(path string, dirEntry os.DirEntry, err error) error {
```
jonastheis: Seems to be a bit complicated for what it is doing. `os.ReadDir()` as described [here](https://stackoverflow.com/questions/14668850/list-directory-in-go) should be simpler and probably faster

packages/protocol/tipmanager/testframework.go
```diff
@@ -107,34 +104,34 @@ func (t *TestFramework) setupEvents() {
...
 			t.test.Logf("TIP REMOVED: %s", block.ID())
 		}
 		atomic.AddUint32(&(t.tipRemoved), 1)
-	}))
+	})
 
-	t.mockAcceptance.BlockAcceptedEvent.Attach(event.NewClosure(func(block *blockgadget.Block) {
-		require.NoError(t.test, t.engine.NotarizationManager.NotarizeAcceptedBlock(block.ModelsBlock))
-	}))
+	event.Hook(t.mockAcceptance.BlockAcceptedEvent, func(block *blockgadget.Block) {
```
jonastheis, line 124: Is `event.Hook(` here expected? As it was `Attach` before.

packages/protocol/tipmanager/testframework.go
```diff
@@ -107,34 +104,34 @@ func (t *TestFramework) setupEvents() {
...
+		require.NoError(t.test, t.Engine.NotarizationManager.NotarizeAcceptedBlock(block.ModelsBlock))
+	})
 
-	t.engine.NotarizationManager.Events.EpochCommitted.Attach(event.NewClosure(func(details *notarization.EpochCommittedDetails) {
-		t.TipManager.PromoteFutureTips(details.Commitment)
-	}))
+	event.Hook(t.Engine.NotarizationManager.Events.EpochCommitted, func(details *notarization.EpochCommittedDetails) {
```
jonastheis, line 128: Is `event.Hook(` here expected? As it was `Attach` before.

packages/protocol/tipmanager/testframework.go
```diff
@@ -107,34 +104,34 @@ func (t *TestFramework) setupEvents() {
...
+		t.Instance.PromoteFutureTips(details.Commitment)
+	})
 
-	t.engine.EvictionState.Events.EpochEvicted.Attach(event.NewClosure(func(index epoch.Index) {
-		t.TipManager.Evict(index)
-	}))
+	event.Hook(t.Engine.EvictionState.Events.EpochEvicted, func(index epoch.Index) {
```
jonastheis, line 132: Is `event.Hook(` here expected? As it was `Attach` before.

packages/protocol/engine/sybilprotection/dpos/sybilprotection.go
```diff
@@ -68,12 +72,14 @@ func NewSybilProtection(engineInstance *engine.Engine, opts ...options.Option[Sy
...
 
-				s.engine.Events.Tangle.BlockDAG.BlockSolid.Attach(event.NewClosure(func(block *blockdag.Block) {
+				s.engine.SubscribeStopped(
+					s.stopInactivityManager,
+				)
+
+				event.AttachWithWorkerPool(s.engine.Events.Tangle.BlockDAG.BlockSolid, func(block *blockdag.Block) {
 					s.markValidatorActive(block.IssuerID(), block.IssuingTime())
-				}))
+				}, s.workers.CreatePool("SybilProtection"))
```
jonastheis: `2*runtime.NumCPU()` workers?

packages/protocol/engine/notarization/attestation.go
```diff
@@ -62,3 +63,20 @@ func (a Attestation) Bytes() (bytes []byte, err error) {
 func (a *Attestation) FromBytes(bytes []byte) (consumedBytes int, err error) {
 	return serix.DefaultAPI.Decode(context.Background(), bytes, a, serix.WithValidation())
 }
+
+func (a *Attestation) IssuerID() identity.ID {
+	return identity.NewID(a.IssuerPublicKey)
```
jonastheis, line 68: Seems like this computed value is used often. Maybe we can just create a non-serialized property on object creation to avoid calling this over and over again?

jonastheis (changes requested):

packages/protocol/engine/engine.go
```diff
@@ -280,176 +276,174 @@ func (e *Engine) initLedger() {
...
+	e.Tangle = tangle.New(e.Workers.CreateGroup("Tangle"), e.Ledger, e.EvictionState, e.SybilProtection.Validators(), e.LastConfirmedEpoch, e.FirstUnacceptedMarker, e.optsTangleOptions...)
 
-	e.Events.Filter.BlockAllowed.Attach(event.NewClosure(func(block *models.Block) {
-		if _, _, err := e.Tangle.Attach(block); err != nil {
+	event.AttachWithWorkerPool(e.Events.Filter.BlockAllowed, func(block *models.Block) {
+		if _, _, err := e.Tangle.BlockDAG.Attach(block); err != nil {
 			e.Events.Error.Trigger(errors.Wrapf(err, "failed to attach block with %s (issuerID: %s)", block.ID(), block.IssuerID()))
 		}
-	}))
+	}, e.Workers.CreatePool("Tangle.Attach", 1))
```
jonastheis: Sure we only want 1 worker here?

packages/protocol/ledger/ledger.go
alexsporn: We are already running in a workerpool inside the ledger

packages/protocol/tipmanager/testframework.go
alexsporn: It is easier to hook inside the tests

packages/protocol/tipmanager/testframework.go
alexsporn: It is easier to hook inside the tests

packages/protocol/tipmanager/testframework.go
alexsporn: It is easier to hook inside the tests

packages/protocol/engine/notarization/attestation.go
alexsporn: This does not seem to be called so often

packages/protocol/chainmanager/manager.go
```diff
@@ -3,24 +3,36 @@ package chainmanager
...
+	ErrCommitmentNotSolid = errors.New("commitment not solid")
 )
 
 type Manager struct {
 	Events              *Events
 	SnapshotCommitment  *Commitment
 	CommitmentRequester *eventticker.EventTicker[commitment.ID]
 
+	//TODO: change to memstorage.EpochStorage?
 	commitmentsByID      map[commitment.ID]*Commitment
```
jonastheis, line 31: This will grow forever no?

packages/protocol/chainmanager/manager.go
```diff
@@ -40,32 +53,27 @@ func NewManager(snapshot *commitment.Commitment) (manager *Manager) {
 	manager.SnapshotCommitment.publishChain(NewChain(manager.SnapshotCommitment))
 
 	manager.CommitmentRequester = eventticker.New(manager.optsCommitmentRequester...)
-	manager.Events.CommitmentMissing.Attach(event.NewClosure(manager.CommitmentRequester.StartTicker))
-	manager.Events.MissingCommitmentReceived.Attach(event.NewClosure(manager.CommitmentRequester.StopTicker))
+	manager.Events.CommitmentMissing.Hook(event.NewClosure(manager.CommitmentRequester.StartTicker))
+	manager.Events.MissingCommitmentReceived.Hook(event.NewClosure(manager.CommitmentRequester.StopTicker))
```
jonastheis: Use new style for events

packages/protocol/chainmanager/manager.go
```diff
@@ -77,23 +85,74 @@ func (m *Manager) ProcessCommitment(commitment *commitment.Commitment) (isSolid
...
+
+	// Do not trigger another event for the same forking point.
+	if commitmentsStorage.Has(forkingPoint.ID()) {
+		return
 	}
 
+	// The forking point should be at least 3 epochs in the past w.r.t this commitment index.
+	//TODO: replace 3 with values depending on the configuration
+	latestChainCommitment := chain.LatestCommitment()
+	if latestChainCommitment.ID().Index()-forkingPoint.ID().Index() < 3 {
```
jonastheis: Fix TODO and make it configurable

packages/protocol/chainmanager/manager.go
```diff
@@ -3,24 +3,36 @@ package chainmanager
...
 type Manager struct {
 	Events              *Events
 	SnapshotCommitment  *Commitment
 	CommitmentRequester *eventticker.EventTicker[commitment.ID]
 
+	//TODO: change to memstorage.EpochStorage?
 	commitmentsByID      map[commitment.ID]*Commitment
 	commitmentsByIDMutex sync.Mutex
 
+	commitmentsByForkingPoint *memstorage.EpochStorage[commitment.ID, *ForkDetectedEvent]
```
jonastheis: Seems to be a bit odd to store an event here. Maybe we can just rename this struct to something else?

packages/protocol/chainmanager/manager.go
```diff
@@ -77,23 +85,74 @@ func (m *Manager) ProcessCommitment(commitment *commitment.Commitment) (isSolid
...
+		Commitment:                   latestChainCommitment.Commitment(),
+		ForkingPointAgainstMainChain: forkingPoint.Commitment(),
+	}
+	commitmentsStorage.Set(forkingPoint.ID(), event)
+	m.Events.ForkDetected.Trigger(event)
+}
+
+func (m *Manager) registerCommitment(commitment *commitment.Commitment) (isNew bool, isSolid bool, wasForked bool, chainCommitment *Commitment) {
+	m.commitmentEntityMutex.Lock(commitment.ID())
+	defer m.commitmentEntityMutex.Unlock(commitment.ID())
```
jonastheis, line 161: Can this lead to a deadlock if another goroutine is currently in `processCommitment` to propagate solidity to the children. There we hold the lock of the parent in `processCommitment` and update the child in `propagateSolidity` where we lock the child. Here we do it the opposite way.

packages/protocol/chainmanager/manager.go
```diff
@@ -148,6 +217,105 @@ func (m *Manager) Commitments(id commitment.ID, amount int) (commitments []*Comm
...
+			childWalker.PushAll(m.deleteAllChildrenFromCache(childWalker.Next())...)
+		}
+	}
+
+	return nil
+}
+
+func (m *Manager) Evict(index epoch.Index) {
+	//TODO: do we need to evict more stuff?
+	m.commitmentsByForkingPoint.Evict(index)
```
jonastheis: We don't need an eviction lock here?

What about `commitmentsByID`?

packages/protocol/testframework.go
```diff
@@ -31,59 +38,47 @@ const genesisTokenAmount = 100
...
 }
 
 // region EngineTestFramework //////////////////////////////////////////////////////////////////////////////////////////
 
-type (
-	TangleTestFramework     = tangle.TestFramework
-	AcceptanceTestFramework = blockgadget.TestFramework
-)
-
 type EngineTestFramework struct {
```
jonastheis: Shouldn't this live in `engine/` ?

packages/protocol/chainmanager/manager.go
alexsporn: Yes, that is why I added that TODO. We need to decide what should happen here

packages/protocol/chainmanager/manager.go
alexsporn: No need for a lock, because the epochStorage and the Storage have locks already.
Regarding the `commitmentsByID` we need to discuss. I think the idea was to keep them all in memory, but maybe @karimodm and @hmoog can chime in

packages/protocol/chainmanager/manager.go
```diff
@@ -148,6 +219,109 @@ func (m *Manager) Commitments(id commitment.ID, amount int) (commitments []*Comm
...
+	}
+
+	return nil
+}
+
+func (m *Manager) Evict(index epoch.Index) {
+	//TODO: do we need to evict more stuff?
+
+	// Forget about the forks that we detected at that epoch so that we can detect them again if they happen
+	evictedForkDetections := m.forkingPointsByCommitments.Evict(index)
```
jonastheis, line 325: I still think this eviction process as a whole should be atomic (by introducing an evicitionMutex like in other components). What exactly happens if there's reads while this stuff is half evicted?

packages/protocol/chainmanager/manager_test.go
```diff
@@ -94,24 +104,159 @@ func TestManager(t *testing.T) {
...
-			lo.Return1(tf.Manager.Commitment(tf.EC("Genesis"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("8*"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("7*"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("6*"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("5*"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("4*"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("3"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("2"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("1"))),
+			lo.Return1(tf.Instance.Commitment(tf.EC("Genesis"))),
```
jonastheis: Maybe put this into the testframework so it's reusable? Same stuff in line 225

packages/protocol/enginemanager/enginemanager.go
```diff
@@ -0,0 +1,179 @@
...
+		instance.Shutdown()
+		_ = instance.RemoveFromFilesystem()
+		_ = os.Remove(snapshotPath)
+		return nil, err
+	}
+
+	return instance, nil
+}
+
+func (e *EngineInstance) InitializeWithSnapshot(snapshotPath string) error {
```
jonastheis, line 170: Separate `EngineInstance` and `EngineManager` at least with a region to make it more readable

packages/protocol/enginemanager/enginemanager.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (m *EngineManager) newEngineInstance() *EngineInstance {
+	dirName := lo.PanicOnErr(uuid.NewUUID()).String()
+	return m.loadEngineInstance(dirName)
+}
+
+func (m *EngineManager) ForkEngineAtEpoch(index epoch.Index) (*EngineInstance, error) {
+	// Dump a snapshot at the target index
+	snapshotPath := filepath.Join(os.TempDir(), fmt.Sprintf("snapshot_%d_%s.bin", index, lo.PanicOnErr(uuid.NewUUID()).String()))
+	if err := m.activeInstance.Engine.WriteSnapshot(snapshotPath, index); err != nil {
```
jonastheis, line 146: How expensive is this `WriteSnapshot`? Does it need to write the entire ledgerstate or is it some sort of incremental diff that is written? What if it was 1TB big?

packages/protocol/enginemanager/enginemanager_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// Block in epoch 2, not accepting anything new.
+	tf.BlockDAG.CreateBlock("2.D", models.WithStrongParents(tf.BlockDAG.BlockIDs("1.D")), models.WithIssuer(identitiesMap["D"]), models.WithIssuingTime(epoch2IssuingTime))
+	tf.BlockDAG.IssueBlocks("2.D")
+	workers.Wait()
+
+	// Block in epoch 11
+	tf.BlockDAG.CreateBlock("11.A", models.WithStrongParents(tf.BlockDAG.BlockIDs("2.D")), models.WithIssuer(identitiesMap["A"]))
+	tf.BlockDAG.IssueBlocks("11.A")
+	workers.Wait()
```
jonastheis: Should not be necessary as it waits in `tf.BlockDAG.IssueBlocks`

packages/protocol/enginemanager/enginemanager_test.go
```diff
@@ -0,0 +1,189 @@
...
+		"2.D":  true,
+		"11.A": true,
+		"11.B": false,
+		"11.C": false,
+	}))
+
+	// Time has advanced to epoch 10 because of A.5, rendering 10 - MinimumCommittableAge(6) = 4 epoch committable
+	require.Eventually(t, func() bool {
+		return tf.Instance.Storage.Settings.LatestCommitment().Index() == epoch.Index(4)
+	}, time.Second, 100*time.Millisecond)
```
jonastheis, line 104: 1 second wait time might be too short?

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
+	}, wpBlocks)
 
-	p.networkProtocol.Events.EpochCommitmentReceived.Attach(event.NewClosure(func(event *network.EpochCommitmentReceivedEvent) {
-		p.chainManager.ProcessCommitment(event.Commitment)
-	}))
-
-	p.networkProtocol.Events.EpochCommitmentRequestReceived.Attach(event.NewClosure(func(event *network.EpochCommitmentRequestReceivedEvent) {
-		if requestedCommitment, _ := p.chainManager.Commitment(event.CommitmentID); requestedCommitment != nil && requestedCommitment.Commitment() != nil {
-			p.networkProtocol.SendEpochCommitment(requestedCommitment.Commitment(), event.Source)
+	wpCommitments := p.Workers.CreatePool("NetworkEvents.EpochCommitments") // Using just 1 worker to avoid contention
```
jonastheis:
```suggestion
	wpCommitments := p.Workers.CreatePool("NetworkEvents.EpochCommitments", 1) // Using just 1 worker to avoid contention
```

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
-	p.networkProtocol.Events.AttestationsReceived.Attach(event.NewClosure(func(event *network.AttestationsReceivedEvent) {
-		p.ProcessAttestations(event.Source)
-	}))
-}
-
-func (p *Protocol) initMainEngine() {
-	p.engine = engine.New(p.storage, p.optsSybilProtectionProvider, p.optsThroughputQuotaProvider, p.optsEngineOptions...)
+	}, wpCommitments)
+
+	wpAttestations := p.Workers.CreatePool("NetworkEvents.Attestations") // Using just 1 worker to avoid contention
```
jonastheis:
```suggestion
	wpAttestations := p.Workers.CreatePool("NetworkEvents.Attestations", 1) // Using just 1 worker to avoid contention
```

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
+	}
+
+	p.networkProtocol.SendAttestations(forkingPoint, blockIDs, attestations, src)
 }
 
-func (p *Protocol) Engine() (instance *engine.Engine) {
-	p.activeEngineMutex.RLock()
-	defer p.activeEngineMutex.RUnlock()
+func (p *Protocol) ProcessAttestations(forkingPoint *commitment.Commitment, blockIDs models.BlockIDs, attestations *orderedmap.OrderedMap[epoch.Index, *set.AdvancedSet[*notarization.Attestation]], source identity.ID) {
+	fmt.Println("Received attestations for", forkingPoint.ID())
```
jonastheis: remove debug print

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
-	return p.engine
+	mainEngine := p.MainEngineInstance()
+
+	// Obtain mana vector at forking point - 1
+	snapshotTargetIndex := forkedEvent.ForkingPoint.Index() - 1
+	wb := sybilprotection.NewWeightsBatch(snapshotTargetIndex)
+
+	var calculatedCumulativeWeight int64
+	mainEngine.Engine.NotarizationManager.PerformLocked(func(m *notarization.Manager) {
+		// Calculate the difference between the latest commitment ledger and the ledger at the snapshot target index
```
jonastheis, line 404: Couldn't we calculate the weight diff for each epoch as part of the commitments and store it until it can be pruned. Then we only need to sum up those diffs here for the epochs instead of every time streaming through all the spent/created outputs. Or is it possible to get this information from the attestations of our current engine themselves?

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
+	if err != nil {
+		p.Events.Error.Trigger(errors.Wrap(err, "error creating new candidate engine"))
+		return
+	}
+
+	// Set the chain to the correct forking point
+	if err := candidateEngine.Engine.Storage.Settings.SetChainID(forkedEvent.ForkingPoint.ID()); err != nil {
+		p.Events.Error.Trigger(errors.Wrap(err, "error setting the ChainID on the forked engine"))
+		candidateEngine.Shutdown()
+		_ = candidateEngine.RemoveFromFilesystem()
```
jonastheis: Would be good to still log this error

packages/protocol/protocol.go
```diff
@@ -45,268 +43,520 @@ type Protocol struct {
...
+	oldEngine.Shutdown()
+
+	p.mainEngine = p.candidateEngine
+	p.candidateEngine = nil
+
+	p.activeEngineMutex.Unlock()
+
+	p.Events.MainEngineSwitched.Trigger(p.MainEngineInstance())
+
+	// TODO: copy over old epochs from the old engine to the new one
```
jonastheis, line 301: What about this TODO?

packages/protocol/protocol_test.go
```diff
@@ -639,27 +677,617 @@ func TestEngine_ShutdownResume(t *testing.T) {
...
+		),
+	)
 
+	tf2 := engine.NewTestFramework(t, workers.CreateGroup("EngineTestFramework2"), engine2)
+	require.NoError(t, tf2.Instance.Initialize(""))
+	workers.Wait()
 	tf2.AssertEpochState(0)
 }
+
+type NodeOnMockedNetwork struct {
```
jonastheis: This deserves it's own file no?

packages/protocol/protocol_test.go
```diff
@@ -639,27 +677,617 @@ func TestEngine_ShutdownResume(t *testing.T) {
...
-		identity.New(identitiesMap["B"]).ID(): 25,
-		identity.New(identitiesMap["C"]).ID(): 25,
-		identity.New(identitiesMap["D"]).ID(): 25,
-		identity.New(identitiesMap["Z"]).ID(): 0,
+	identitiesWeights := map[ed25519.PublicKey]uint64{
+		identity.New(identitiesMap["A"]).PublicKey(): 25,
+		identity.New(identitiesMap["B"]).PublicKey(): 25,
+		identity.New(identitiesMap["C"]).PublicKey(): 25,
+		identity.New(identitiesMap["D"]).PublicKey(): 25,
+		identity.New(identitiesMap["Z"]).PublicKey(): 0,
```
jonastheis, line 680: Not necessarily part of this PR but this code is repeated over and over again. It should be part of some testframework to make it easy to initialize a bunch of identities with alias and weights from which the snapshot is created

jonastheis (changes requested):

packages/protocol/protocol.go
alexsporn: This is imho not really needed for the MVP, this is to allow node operators to keep long history even after chain switches. The majority of nodes will follow the main chain anyway and probably never do a chain switch.
This is something that we should plan and implement as a separate feature, since it will probably mean that we need to merge the chainmanager, enginemanager and databasemanager.

packages/protocol/enginemanager/enginemanager.go
alexsporn: Yes, I had the same comment to @karimodm, but right now that is the only way we have to "clone" the state of an engine. It is good enough for the MVP but for production we need a way to stream it from one database to the other.

packages/protocol/enginemanager/enginemanager_test.go
alexsporn: 1 second is enough here,, we only need to wait for the issued blocks to be accepted and that will trigger an epoch commitment

packages/protocol/protocol.go
alexsporn: This is what was possible with the current available means. This is not called very often, so I don't see a problem in streaming it here. We should think about this whenever we tackle the ledger refactoring though.

jonastheis (approved):
Good stuff

packages/protocol/chainmanager/commitment.go
```diff
@@ -57,8 +58,10 @@ func (c *Commitment) Children() (children []*Commitment) {
 	c.RLock()
 	defer c.RUnlock()
 
-	children = make([]*Commitment, len(c.children))
-	copy(children, c.children)
+	children = make([]*Commitment, 0, len(c.children))
```
hmoog: Can be replaced by `return lo.Values(c.children)`

packages/protocol/chainmanager/commitment.go
```diff
@@ -116,13 +119,43 @@ func (c *Commitment) registerChild(child *Commitment) (isSolid bool, chain *Chai
...
+
+	return c.children[c.mainChildID]
+}
+
+func (c *Commitment) setMainChild(commitment *Commitment) error {
+	c.Lock()
+	defer c.Unlock()
+
+	if _, has := c.children[commitment.ID()]; !has {
+		return errors.Errorf("trying to set a main child %s before registering it as a children", commitment.ID())
```
hmoog:
```suggestion
		return errors.Errorf("trying to set a main child %s before registering it as a child", commitment.ID())
```

packages/protocol/chainmanager/fork.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/iotaledger/goshimmer/packages/core/epoch"
+)
+
+type Fork struct {
+	Source       identity.ID
+	Commitment   *commitment.Commitment
+	ForkingPoint *commitment.Commitment
+}
+
+func (e *Fork) EndEpoch() epoch.Index {
```
hmoog: Fork is a struct with public properties - having an additional getter feels unnecessary - we also use `forkedEvent.ForkingPoint.Index()` in other situations.

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -40,29 +42,33 @@ type Booker struct {
 
 	optsMarkerManager []options.Option[markermanager.MarkerManager[models.BlockID, *Block]]
 
-	*blockdag.BlockDAG
+	Workers *workerpool.Group
```
hmoog: should be private

packages/protocol/tipmanager/tipmanager.go
```diff
@@ -228,20 +261,26 @@ func (t *TipManager) PromoteFutureTips(cm *commitment.Commitment) {
...
 func (t *TipManager) Evict(index epoch.Index) {
-	t.evictionMutex.Lock()
-	defer t.evictionMutex.Unlock()
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	t.evict(index)
+}
 
+func (t *TipManager) evict(index epoch.Index) {
```
hmoog: Making a utility method for a single function call with the same signature seems unnecessary just replace calls to this func to call of the wrapped func directly.

hmoog (changes requested):
LGTM apart from these small issues

packages/protocol/tipmanager/tipmanager.go
alexsporn: the lowercase variant has no locking and is called from 2 places that already have the lock. The uppercase variant is public and called from the outside

alexsporn: I replaced it by a call to `t.futureTips.Evict` directly

hmoog: yeah thats what I meant :P

hmoog (approved):

PR #2543, opened 2023-01-24
Added partition support to the MockedNetwork

alexsporn:
# Description of change

This allows us to partition the network in unit tests


hmoog (approved):
LGBTQM

PR #2554, opened 2023-02-07
[Chore] cleanup worlkflows

alexsporn:
- Updated actions
- Using go 1.19
- Removing cache since we are not really using it

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,18 +18,7 @@ jobs:
     steps:
 
       - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Cache Go modules for tester container
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/go
-          # make sure concurrent runs (not really supported) do not match the same key but instead fall back to a reasonable cache
-          key: ${{ runner.os }}-it-go-${{ hashFiles('tools/integration-tests/tester/go.sum') }}-${{ github.job }}-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-it-go-${{ hashFiles('tools/integration-tests/tester/go.sum') }}-
-            ${{ runner.os }}-it-go-
+        uses: actions/checkout@v3
 
       - name: Build GoShimmer image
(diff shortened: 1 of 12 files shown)


.github/workflows/unit-test.yml
```diff
@@ -11,26 +11,16 @@ jobs:
...
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/go-build
-            ~/go/pkg/mod
-          # make sure concurrent runs (not really supported) do not match the same key but instead fall back to a reasonable cache
-          key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}-${{ github.event.action }}${{ github.event.after }}
-          restore-keys: |
-            ${{ runner.os }}-go-${{ hashFiles('go.sum') }}
-            ${{ runner.os }}-go-
```
jonastheis, line 33: The go build cache should still be useful. Because we were probably [trashing](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) all of our cache space on Github Actions with all the Docker stuff this wasn't effective. But if we get rid of this, it should be fine and speed up the unit tests

.github/workflows/integration-tests.yml
```diff
@@ -18,18 +18,7 @@ jobs:
     steps:
 
       - name: Checkout repository
-        uses: actions/checkout@v2
```
jonastheis, line 21: Maybe this might be useful too now? 

 > The go build cache should still be useful. Because we were probably [trashing](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) all of our cache space on Github Actions with all the Docker stuff this wasn't effective. But if we get rid of this, it should be fine.

alexsporn: just look at the caches generated, the fallbacks are not really working imho, so we are generating the same cache for each integration test. One run of the CI generates 8GB in cache.
I can take another look and see if stuff is properly being reused, but the steps of restoring and saving the caches are currently also taking very long

PR #2560, opened 2023-02-13
Implement epoch APIs/dashboard page

### LINKED ISSUE #2517: Implement epoch API (currently just stubs)
Fill in the epoch API stubs to return actual epoch data.
Stubs: https://github.com/iotaledger/goshimmer/commit/a22745db1401c338f8578f942925846ad373658a

* Possibly extend the retainer to also hold epoch-related data.

jkrvivian:
# Description of change

Close #2517 

## Type of change

- Breaking change
## Change checklist


packages/app/jsonmodels/epochs.go
```diff
@@ -5,16 +5,20 @@ import (
 )
 
 type EpochInfo struct {
-	EI     uint64 `json:"Index"`
-	ECR    string `json:"RootsID"`
-	PrevEC string `json:"prevEC"`
+	ID               string `json:"id"`
+	Index            uint64 `json:"index"`
+	ECR              string `json:"rootsID"`
```
jonastheis: Maybe just `roots` or `rootsID`? It should be consistent with the naming that we use inside the protocol, everything else is confusing

packages/app/retainer/retainer.go
```diff
@@ -177,11 +200,30 @@ func (r *Retainer) setupEvents() {
...
+				return true
+			})
+			_ = e.AcceptedTransactions.Stream(func(key utxo.TransactionID) bool {
+				txIDs.Add(key)
+				return true
+			})
+
+			cc.setCommitment(e.Commitment, blockIDs, txIDs, e.CreatedOutputs, e.SpentOutputs)
+		}
+	}), r.workerPool)
```
jonastheis: I think this could be a separate WP with only 1 worker specifically processing the `EpochCommitted` event

packages/app/retainer/commitment.go
```diff
@@ -0,0 +1,97 @@
...
+		panic(errors.Wrap(err, "error registering BlockIDs type settings"))
+	}
+	err = serix.DefaultAPI.RegisterTypeSettings(utxo.NewTransactionIDs(), serix.TypeSettings{}.WithLengthPrefixType(serix.LengthPrefixTypeAsUint16))
+	if err != nil {
+		panic(errors.Wrap(err, "error registering TransactionIDs type settings"))
+	}
+	err = serix.DefaultAPI.RegisterTypeSettings(utxo.NewOutputIDs(), serix.TypeSettings{}.WithLengthPrefixType(serix.LengthPrefixTypeAsUint16))
+	if err != nil {
+		panic(errors.Wrap(err, "error registering OutputIDs type settings"))
+	}
```
jonastheis: These type settings are global. Is this correct? I think you can override the default type settings when you call serialize / deserialize

packages/app/retainer/retainer.go
```diff
@@ -10,41 +10,56 @@ import (
...
+)
+
 type Retainer struct {
-	workerPool     *workerpool.UnboundedWorkerPool
-	cachedMetadata *memstorage.EpochStorage[models.BlockID, *cachedMetadata]
-	blockStorage   *database.PersistentEpochStorage[models.BlockID, BlockMetadata, *models.BlockID, *BlockMetadata]
+	workerPool        *workerpool.UnboundedWorkerPool
+	cachedMetadata    *memstorage.EpochStorage[models.BlockID, *cachedMetadata]
+	blockStorage      *database.PersistentEpochStorage[models.BlockID, BlockMetadata, *models.BlockID, *BlockMetadata]
+	cachedCommitment  *memstorage.EpochStorage[commitment.ID, *cachedCommitment]
```
jonastheis: Since commitments are not created too frequently and it's only 1 commitment per epoch I think a `cachedCommitment` is not necessary. We can probably simply create the `CommitmentDetails` directly and write them to disk. This simplifies the whole process massively, as there's only 1 storage, no eviction lock, no eviction and no separate cached commitment object

packages/protocol/engine/notarization/manager.go
```diff
@@ -241,11 +243,26 @@ func (m *Manager) createCommitment(index epoch.Index) (success bool) {
 		m.Events.Error.Trigger(errors.Wrap(err, "failed to store latest commitment"))
 	}
 
+	var (
+		createdOutputs = utxo.NewOutputIDs()
+		spentOutputs   = utxo.NewOutputIDs()
+	)
+	m.ledgerState.StateDiffs.StreamCreatedOutputs(index, func(owm *ledger.OutputWithMetadata) error {
```
jonastheis: This seems to be very expensive. Can't we put a pointer to `m.ledgerState.StateDiffs.StreamCreatedOutputs` and the consumer (in our case the retainer) can stream it in a separate thread?

jonastheis (changes requested):

plugins/metrics/metrics_workerpool.go
```diff
@@ -32,7 +32,8 @@ var WorkerPoolMetrics = collector.NewCollection(workerPoolNamespace,
 			}
 
 			if deps.Retainer != nil {
-				collected[retainerLabel] = float64(deps.Retainer.WorkerPool().WorkerCount())
+				collected[retainerLabel] = float64(deps.Retainer.BlockWorkerPool().WorkerCount() +
+					deps.Retainer.CommitmentWorkerPool().WorkerCount())
```
jonastheis: We could expose the `Workers` in the retainer and just do the same as for the protocol above.

```
for name, wp := range deps.Retainer.Workers.Pools() {
	collected[name] = float64(wp.WorkerCount())
	collected[name] = float64(wp.WorkerCount())
}
```

packages/protocol/engine/notarization/manager.go
```diff
@@ -201,10 +201,12 @@ func (m *Manager) createCommitment(index epoch.Index) (success bool) {
...
-		Commitment:                newCommitment,
-		AcceptedBlocksCount:       acceptedBlocks.Size(),
-		AcceptedTransactionsCount: acceptedTransactions.Size(),
-		ActiveValidatorsCount:     0,
+		Commitment:            newCommitment,
+		AcceptedBlocks:        acceptedBlocks,
+		AcceptedTransactions:  acceptedTransactions,
+		SpentOutputs:          m.ledgerState.StateDiffs.StreamSpentOutputs,
+		CreatedOutputs:        m.ledgerState.StateDiffs.StreamCreatedOutputs,
+		ActiveValidatorsCount: 0,
```
jonastheis, line 214:
```
m.Events.EpochCommitted.Trigger(&EpochCommittedDetails{
	Commitment:           newCommitment,
	AcceptedBlocks:       acceptedBlocks,
	AcceptedTransactions: acceptedTransactions,
	SpentOutputs: func(callback func(*ledger.OutputWithMetadata) error) error {
		return m.ledgerState.StateDiffs.StreamSpentOutputs(index, callback)
	},
	CreatedOutputs: func(callback func(*ledger.OutputWithMetadata) error) error {
		return m.ledgerState.StateDiffs.StreamCreatedOutputs(index, callback)
	},
	ActiveValidatorsCount: 0,
})
```
Something like this makes it easier to use as the receiver.

packages/protocol/engine/notarization/events.go
```diff
@@ -36,10 +40,14 @@ var NewEpochMutationsEvents = event.CreateGroupConstructor(func() (self *EpochMu
 	}
 })
 
+type OutputsProvider func(epoch.Index, func(*ledger.OutputWithMetadata) error) error
```
jonastheis:
```suggestion
type OutputsProvider func(callback func(*ledger.OutputWithMetadata) error) error
```

packages/app/retainer/retainer.go
```diff
@@ -77,20 +91,28 @@ func (r *Retainer) BlockMetadata(blockID models.BlockID) (metadata *BlockMetadat
...
 }
 
-func (r *Retainer) LoadAll(index epoch.Index) (ids *advancedset.AdvancedSet[*BlockMetadata]) {
-	r.evictionLock.RLock(index)
-	defer r.evictionLock.RUnlock(index)
+func (r *Retainer) Commitment(index epoch.Index) (c *CommitmentDetails, exists bool) {
+	return r.getCommitmentDetails(index)
+}
+
+func (r *Retainer) CommitmentbyID(id commitment.ID) (c *CommitmentDetails, exists bool) {
```
jonastheis:
```suggestion
func (r *Retainer) CommitmentByID(id commitment.ID) (c *CommitmentDetails, exists bool) {
```

packages/app/retainer/retainer.go
```diff
@@ -147,40 +174,77 @@ func (r *Retainer) setupEvents() {
...
+	}, event.WithWorkerPool(r.blockWorkerPool))
 
 	r.protocol.Events.Engine.Consensus.BlockGadget.BlockConfirmed.Hook(func(block *blockgadget.Block) {
 		if cm := r.createOrGetCachedMetadata(block.ID()); cm != nil {
 			cm.setConfirmationBlock(block)
 		}
-	}, event.WithWorkerPool(r.workerPool))
+	}, event.WithWorkerPool(r.blockWorkerPool))
 
 	r.protocol.Events.Engine.EvictionState.EpochEvicted.Hook(r.storeAndEvictEpoch)
```
jonastheis: This should also be done in the worker pool as this is slow and hooked!

packages/app/retainer/retainer.go
```diff
@@ -147,40 +174,77 @@ func (r *Retainer) setupEvents() {
...
+				spentOutputs.Add(owm.ID())
+				return nil
+			})
+
+			_ = e.CreatedOutputs(e.Commitment.Index(), func(owm *ledger.OutputWithMetadata) error {
+				createdOutputs.Add(owm.ID())
+				return nil
+			})
+
+			cd.setCommitment(e.Commitment, blockIDs, txIDs, createdOutputs, spentOutputs)
```
jonastheis: You can do all this in `setCommitment` directly on the respective fields

packages/app/retainer/retainer.go
```diff
@@ -147,40 +174,77 @@ func (r *Retainer) setupEvents() {
...
+	}, event.WithWorkerPool(r.blockWorkerPool))
 
 	r.protocol.Events.Engine.EvictionState.EpochEvicted.Hook(r.storeAndEvictEpoch)
+
+	r.protocol.Engine().NotarizationManager.Events.EpochCommitted.Hook(func(e *notarization.EpochCommittedDetails) {
+		if e.Commitment.Index() < r.protocol.Engine().EvictionState.LastEvictedEpoch() {
+			return
+		}
+
+		cd, _ := r.getCommitmentDetails(e.Commitment.Index())
```
jonastheis: Why get it from storage? We should only commit each epoch once so you should always be able to create a new `CommitmentDetails` and then just `setCommitment` (don't forget to set the ID in here) and finally store it.

packages/app/retainer/retainer.go
```diff
@@ -240,9 +303,19 @@ func (r *Retainer) storeBlockMetadata(metas []*BlockMetadata) {
 	}
 }
 
+func (r *Retainer) storeCommitmentDetails(c *CommitmentDetails) {
+	if c == nil {
+		return
+	}
```
jonastheis: this should never happen and can be removed

jonastheis (changes requested):

jonastheis (approved):

PR #2561, opened 2023-02-13
Commitment spammer

### LINKED ISSUE #2514: Design commitment spammers
* We need to extend the evil spammer to include scenarios that deal with epoch commitments

daria305:
# Description of change
Fixes  #2514 

Commitment Spammer implementation, fixes some known bugs in evil spammer too.
Without combination of correct network setup the fork will not be detected. 
Needs more future adjustment if it is supposed to succeed in chain switching/

Example usage: 
` go run . commitments --rate 10 --duration 5s --type random`
Available types:
 - latest - spammer node spams with valid commitment requested from the honest node
 - random - spam with the latest commitment altered by using fake rootsID, supposed to increase seen committment
 - fork - creates a forked chain, if blindly run on docker network will not work as block will be rejected for not solid chain error. To succeed needs to be combined with correctly set up docker network
 
Functions to create commitments can be found and added here: client/evilspammer/commitmentmanager.go:33


tools/evil-spammer/config.go
```diff
@@ -9,8 +9,8 @@ import (
 // Nodes used during the test, use at least two nodes to be able to doublespend.
 var (
 	// urls = []string{"http://bootstrap-01.feature.shimmer.iota.cafe:8080", "http://vanilla-01.feature.shimmer.iota.cafe:8080", "http://drng-01.feature.shimmer.iota.cafe:8080"}
-	//urls = []string{"http://localhost:8080", "http://localhost:8090", "http://localhost:8070", "http://localhost:8040"}
-	urls = []string{}
+	urls = []string{"http://localhost:8080", "http://localhost:8090", "http://localhost:8070", "http://localhost:8040"}
+	//urls = []string{}
```
piotrm50: We should not put any addresses by default as it caused some confusion in the past, when the command line parameter wasn't read correctly but the spammer still worked.

tools/evil-spammer/identity/identityconfig.go
```diff
@@ -0,0 +1,68 @@
...
+	"os"
+)
+
+type Params map[string]Identities
+type Identities map[string]privKeyURLPair
+type privKeyURLPair [2]string
+
+var Config = Params{}
+
+var identityConfigJSON = `{
```
piotrm50, line 15: Would it work better if consensus weight distribution in the snapshot was modified and evil spammer's identities had majority of weight? 

Do I understand correctly that currently the evil spammer spams commitments with identity of real nodes?

client/evilspammer/commitmentmanager.go
```diff
@@ -0,0 +1,245 @@
...
+	forkedComm := commitment.New(
+		validComm.Index(),
+		prevComm.ID(),
+		randomRoot(),
+		validComm.CumulativeWeight(),
+	)
+	return forkedComm, nil
+}
+
+func randomCommittmentChain(currSlot slot.Index) *commitment.Commitment {
```
piotrm50:
```suggestion
func randomCommitmentChain(currSlot slot.Index) *commitment.Commitment {
```

tools/evil-spammer/identity/identityconfig.go
daria305: Yes, you could use peer_master or whichever node you will put there, also some random fake id/seed should be ok, but there will be no mana assigned to it with the default snapshot.
Not sure if it would be much of a help to increase mana of the spammer. For sure we should disable activity blocks for spammer node, so we spam only the invalid/incorrect chain. The fork should be observed by the nodes, but without attestations proof it should not work to trigger the chain switch.
The random commitment was supposed to trigger more requests for attestations proofs, as the chain is totally different, but I would need to play with it more, not sure if the weight is the problem.

piotrm50 (approved):

PR #2565, opened 2023-02-14
Fixes in web API and frontend

### LINKED ISSUE #2562: Fix visualizer issues
Might be related to the way we handle root blocks. But often blocks seem to be not connected and thus the Tangle does not look proper

jkrvivian:
# Description of change

Close #2562 

## Type of change

- Bug fix

## Change checklist


daria305 (approved):
I tried it out and nice work, explorer works again under spam :tada: there are some floating blocks, but that's rather normal. 
And also I don't remember if ever we've seen `Faucet payload` in the explorer, but now there are only `Transaction` names showing up after funds request.
One more thing that probably is not even part of this PR: the tips tab refreshes only when the page is loaded it could be more useful if it works more like explorer

piotrm50:
> One more thing that probably is not even part of this PR: the tips tab refreshes only when the page is loaded it could be more useful if it works more like explorer

That would be nice visually, but I'm not sure if it would be more practical for debugging tbh.

alexsporn:
This PR needs updating due to the renaming from epoch to slot

alexsporn (approved):

PR #2568, opened 2023-02-17
Update to hive.go modules and cleanup

### LINKED ISSUE #462: Flatten & cleanup hive.go packages
* ... and move stuff from goshimmer into here.
* many generics are wrapper around non-generic types -> need to be native generics
* clean up APIs and make them more consistent
* check downstream dependencies (eg crypto libs?)
* move generic stuff from GoShimmer into hive
* move stuff from hive into GoShimmer / new node implementation if it's specific to the node
* unify / reuse between node implementations as much as possible (eg proper networking layer)
probably some more things

alexsporn:


packages/network/manualpeering/manualpeering.go
```diff
@@ -89,9 +90,8 @@ func NewManager(p2pm *p2p.Manager, local *peer.Local, log *logger.Logger) *Manag
 		log:               log,
 		reconnectInterval: defaultReconnectInterval,
 		knownPeers:        map[identity.ID]*knownPeer{},
+		workerPool:        workerpool.New("manualpeering.Manager", 1),
```
jonastheis: shouldn't this be part of the workergroup that is handed into the protocol upon creation? this way we can wait for it and it will show up automatically in the metrics

packages/network/manualpeering/manualpeering.go
```diff
@@ -174,24 +174,30 @@ func (m *Manager) GetPeers(opts ...GetPeersOption) []*KnownPeer {
...
 	}
 	m.stopOnce.Do(func() {
 		m.stopMutex.Lock()
 		defer m.stopMutex.Unlock()
 		m.isStopped = true
 		err = errors.WithStack(m.removeAllKnownPeers())
-		m.p2pm.NeighborGroupEvents(p2p.NeighborsGroupManual).NeighborRemoved.Detach(m.onGossipNeighborRemovedClosure)
-		m.p2pm.NeighborGroupEvents(p2p.NeighborsGroupManual).NeighborAdded.Detach(m.onGossipNeighborAddedClosure)
+		m.onGossipNeighborRemovedHook.Unhook()
+		m.onGossipNeighborAddedHook.Unhook()
```
jonastheis, line 199: Why do we unhook these events? Do we do that in other components as well?

packages/network/protocol.go
```diff
@@ -12,14 +12,15 @@ import (
...
-	"github.com/iotaledger/hive.go/core/generics/orderedmap"
-	"github.com/iotaledger/hive.go/core/generics/set"
-	"github.com/iotaledger/hive.go/core/generics/shrinkingmap"
 	"github.com/iotaledger/hive.go/core/identity"
 	"github.com/iotaledger/hive.go/core/types"
-	"github.com/iotaledger/hive.go/core/workerpool"
+	"github.com/iotaledger/hive.go/ds/advancedset"
+	"github.com/iotaledger/hive.go/ds/orderedmap"
+	"github.com/iotaledger/hive.go/ds/shrinkingmap"
+	dsTypes "github.com/iotaledger/hive.go/ds/types"
```
jonastheis, line 20: should we find another name for this? or rename`"github.com/iotaledger/hive.go/core/types"` ?

.goreleaser.yml
```diff
@@ -9,105 +9,62 @@ env:
...
-  - id: goshimmer
-    binary: goshimmer
-    flags:
-      - -tags=rocksdb
-    ldflags:
-      - -s -w -X github.com/iotaledger/goshimmer/plugins/banner.AppVersion={{.Version}}
-    goos:
-      - linux
-    goarch:
-      - amd64
```
jonastheis, line 22: We don't build GoShimmer?

plugins/p2p/plugin.go
```diff
@@ -50,12 +49,12 @@ func run(plugin *node.Plugin) {
 
 func configureLogging() {
 	// log the p2p events
-	deps.P2PMgr.NeighborGroupEvents(p2p.NeighborsGroupAuto).NeighborAdded.Attach(event.NewClosure(func(event *p2p.NeighborAddedEvent) {
+	deps.P2PMgr.NeighborGroupEvents(p2p.NeighborsGroupAuto).NeighborAdded.Hook(func(event *p2p.NeighborAddedEvent) {
```
jonastheis, line 43: Was attached before

plugins/p2p/plugin.go
```diff
@@ -50,12 +49,12 @@ func run(plugin *node.Plugin) {
...
 		n := event.Neighbor
 		Plugin.LogInfof("Neighbor added: %s / %s", p2p.GetAddress(n.Peer), n.ID())
-	}))
-	deps.P2PMgr.NeighborGroupEvents(p2p.NeighborsGroupAuto).NeighborRemoved.Attach(event.NewClosure(func(event *p2p.NeighborRemovedEvent) {
+	})
+	deps.P2PMgr.NeighborGroupEvents(p2p.NeighborsGroupAuto).NeighborRemoved.Hook(func(event *p2p.NeighborRemovedEvent) {
```
jonastheis, line 48: Was attached before

plugins/remotelog/plugin.go
```diff
@@ -80,28 +78,22 @@ func configure(_ *node.Plugin) {
 
 	getGitInfo()
 
-	workerPool = workerpool.NewNonBlockingQueuedWorkerPool(func(task workerpool.Task) {
-		deps.RemoteLogger.SendLogMsg(task.Param(levelIndex).(logger.Level), task.Param(nameIndex).(string), task.Param(blockIndex).(string))
-
-		task.Return(nil)
-	}, workerpool.WorkerCount(runtime.GOMAXPROCS(0)), workerpool.QueueSize(1000))
+	workerPool = workerpool.New("RemoteLog", 1)
```
jonastheis: why not use the worker pool integrated in the plugin?

jonastheis (changes requested):

.goreleaser.yml
alexsporn: Correct. We should focus on 1 setup (Docker).

plugins/remotelog/plugin.go
alexsporn: Good catch

packages/network/protocol.go
alexsporn: This is temporary until we refactor the `core/types` package

packages/network/manualpeering/manualpeering.go
alexsporn: This is a package and the Stop should clean up.

packages/network/manualpeering/manualpeering.go
alexsporn: The problem I had is that this manager has nothing to do with the protocol. It is instantiated by the manualpeering plugin in the provide phase. I refactored it now to pass an external workerpool and passed the one from the plugin in the provide func

jonastheis (approved):

PR #2578, opened 2023-02-21
Fix protocol not running on node startup

jonastheis:
If not done synchronously there might be a race condition where https://github.com/iotaledger/goshimmer/blob/1cfa01c301f15cdfc7b04ad09109695265a9589c/plugins/metrics/metrics_tangle.go#L77 is nil and the node panics if metrics are enabled


alexsporn:
The issue is a different one. The metrics plugin should not be accessing `Network().Events` and instead use `Events.Network`.
I created a PR with a fix https://github.com/iotaledger/goshimmer/pull/2580

PR #2588, opened 2023-02-27
(e|E)poch -> (s|S)lot and Time Provider

### LINKED ISSUE #462: Flatten & cleanup hive.go packages

alexsporn:
- Renamed (e|E)poch -> (s|S)lot
- Refactored the epoch.GenesisTime and epoch.Duration into a slot.TimeProvider that can be injected into classes requiring time calculations.

This is in preparation to move everything slot-related to hive.go


hmoog (approved):
LGTM

PR #2591, opened 2023-02-28
Properly compute individually mapped conflicts from new structure details

### LINKED ISSUE #2590: Node carries a "Rejected" conflict forward not booked by other nodes

karimodm:
# Description of change

Whenever assigning past markers to a new block from a set of strong parents, only the highest ranking marker for each sequence is kept. Although, in order to compute the individually-mapped added and subtracted conflicts, the whole list of conflict mapped on past markers from the strong parents was used. As the new structure details do not necessarily carry the same information of this latter list - as the past markers got "normalized" -, some conflict where not represented anymore.

Bonus content: 
 * fixed deadlock between futureblocks and solidifier.
 * conflict mapping on markers looks racey: introduce a DAGMutex on Sequences in the booker.
 * when inheriting structure details: do not normalize markers already referenced by other markers; it is computationally heavy and pretty much useless.  

Fixes #2590


packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -57,6 +58,7 @@ func New(workers *workerpool.Group, blockDAG *blockdag.BlockDAG, ledger *ledger.
 		attachments:       newAttachments(),
 		blocks:            memstorage.NewEpochStorage[models.BlockID, *virtualvoting.Block](),
 		bookingMutex:      syncutils.NewDAGMutex[models.BlockID](),
+		sequenceMutex:     syncutils.NewDAGMutex[markers.SequenceID](),
```
piotrm50, line 61: Having a SequenceMutex in the booker, is the one in the MarkersManager still necessary?

packages/protocol/engine/tangle/booker/booker_test.go
```diff
@@ -1149,3 +1149,40 @@ func TestOTV_Track(t *testing.T) {
 		tf.Ledger.Transaction("Tx4").ID(): tf.VirtualVoting.Votes.ValidatorsSet("A", "B", "C", "D", "E"),
 	}))
 }
+
+func TestInheritancePrecedence(t *testing.T) {
```
piotrm50: This doesn't test anything. Should either add some asserts or remove it.

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -348,42 +376,69 @@ func (b *Booker) inheritConflictIDs(block *virtualvoting.Block) (inheritedConfli
...
+	dislikedFutureCone := b.Ledger.Utils.ConflictIDsInFutureCone(dislikedConflictIDs)
+	inheritedConflictIDs.DeleteAll(dislikedFutureCone)
 
 	// block always sets Like reference its own conflict, if its payload is a transaction, and it's conflicting
 	if selfConflictID, selfDislikedConflictIDs, isTransaction := b.PayloadConflictID(block); isTransaction && !selfConflictID.IsEmpty() {
+		selfFutureDislikedFuture := b.Ledger.Utils.ConflictIDsInFutureCone(selfDislikedConflictIDs)
 		inheritedConflictIDs.Add(selfConflictID)
 		// if a payload is a conflicting transaction, then remove any conflicting conflicts from supported conflicts
-		inheritedConflictIDs.DeleteAll(b.Ledger.Utils.ConflictIDsInFutureCone(selfDislikedConflictIDs))
+		inheritedConflictIDs.DeleteAll(selfFutureDislikedFuture)
```
piotrm50: Does not need a separate variable. Probably some print leftover

```suggestion
		inheritedConflictIDs.DeleteAll(b.Ledger.Utils.ConflictIDsInFutureCone(selfDislikedConflictIDs))
```

packages/protocol/engine/tangle/booker/booker.go
```diff
@@ -348,42 +376,69 @@ func (b *Booker) inheritConflictIDs(block *virtualvoting.Block) (inheritedConfli
...
 	inheritedConflictIDs.AddAll(weakPayloadConflictIDs)
 	inheritedConflictIDs.AddAll(likedConflictIDs)
-	inheritedConflictIDs.DeleteAll(b.Ledger.Utils.ConflictIDsInFutureCone(dislikedConflictIDs))
+
+	dislikedFutureCone := b.Ledger.Utils.ConflictIDsInFutureCone(dislikedConflictIDs)
+	inheritedConflictIDs.DeleteAll(dislikedFutureCone)
 
 	// block always sets Like reference its own conflict, if its payload is a transaction, and it's conflicting
 	if selfConflictID, selfDislikedConflictIDs, isTransaction := b.PayloadConflictID(block); isTransaction && !selfConflictID.IsEmpty() {
+		selfFutureDislikedFuture := b.Ledger.Utils.ConflictIDsInFutureCone(selfDislikedConflictIDs)
```
piotrm50:
```suggestion
```

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -225,7 +225,9 @@ func (r *ReferenceProvider) adjustOpinion(conflictID utxo.TransactionID, exclude
 	likedConflictID, dislikedConflictIDs := engineInstance.Consensus.ConflictResolver.AdjustOpinion(conflictID)
 
 	if likedConflictID.IsEmpty() {
-		panic("likedConflictID empty when trying to adjust opinion!")
+		return false, models.EmptyBlockID, errors.Errorf("likedConflictID empty when trying to adjust opinion for %s", conflictID)
+		// TODO: re-introduce this, as it should never happen, especially without conflict eviction
+		//panic("likedConflictID empty when trying to adjust opinion!")
```
piotrm50: I think it would be best to simply return an error in this case. It should never happen, but currently we don't create the both conflicts atomically, so it may happen that we try to evaluate it before the second one is created, which could possibly cause this. We need to discuss if and how to fix that.

packages/protocol/engine/tangle/booker/booker.go
piotrm50: It is, never mind.

piotrm50 (approved):

PR #2598, opened 2023-03-02
Load genesis timestamp and slot duration from snapshot

alexsporn:


packages/protocol/engine/notarization/manager.go
```diff
@@ -106,7 +108,7 @@ func (m *Manager) NotarizeOrphanedBlock(block *models.Block) (err error) {
 		return errors.Wrap(err, "failed to remove accepted block from slot mutations")
 	}
 
-	if _, err = m.Attestations.Delete(NewAttestation(block, m.SlotTimeProvider)); err != nil {
+	if _, err = m.Attestations.Delete(NewAttestation(block, m.storage.Settings.SlotTimeProvider())); err != nil {
```
hmoog, line 111: Shouldn't the Attestations also use a SlotTimeProvider func?

hmoog (changes requested):
LGTM

packages/protocol/engine/notarization/manager.go
alexsporn: They just use it in the initializer once

hmoog (approved):

PR #2601, opened 2023-03-03
Feat: Move clock to self contained module

hmoog:
# Description of change

This PR moves the Clock to a self-contained module to make the wiring between different components easier to understand.

## Type of change

## Change checklist


packages/protocol/engine/clock/blocktime/clock.go
```diff
@@ -0,0 +1,75 @@
...
+					c.acceptedTime.Set(e.SlotTimeProvider().EndTime(e.Storage.Settings.LatestCommitment().Index()))
+					c.confirmedTime.Set(e.SlotTimeProvider().EndTime(e.Storage.Settings.LatestCommitment().Index()))
+
+					c.TriggerInitialized()
+				})
+
+				e.Events.Clock.AcceptedTimeUpdated.LinkTo(c.acceptedTime.OnUpdated)
+				e.Events.Clock.ConfirmedTimeUpdated.LinkTo(c.confirmedTime.OnUpdated)
+
+				async := event.WithWorkerPool(e.Workers.CreatePool("Clock", 1))
```
alexsporn: don't call the workerpool `async`, this makes it confusing

packages/protocol/engine/clock/events.go
```diff
@@ -6,31 +6,22 @@ import (
 	"github.com/iotaledger/hive.go/runtime/event"
 )
 
-// Events is a collection of events that can be triggered by the Clock.
+// Events contains a dictionary of events that are triggered by the Clock.
 type Events struct {
-	// AcceptanceTimeUpdated is triggered when a Acceptance Tangle Time advances.
-	AcceptanceTimeUpdated *event.Event1[*TimeUpdateEvent]
+	// AcceptedTimeUpdated is triggered when the accepted time is updated.
+	AcceptedTimeUpdated *event.Event1[time.Time]
```
alexsporn, line 12: Maybe pass the `RelativeTime` in the events instead of time.Time?

packages/protocol/mockednetwork/node.go
```diff
@@ -202,8 +201,8 @@ func (n *Node) attachEngineLogs(instance *enginemanager.EngineInstance) {
 		fmt.Printf("%s > [%s] Tangle.VirtualVoting.SequenceTracker.VotersUpdated: %s %s %d -> %d\n", n.Name, engineName, event.Voter, event.SequenceID, event.PrevMaxSupportedIndex, event.NewMaxSupportedIndex)
 	})
 
-	events.Clock.AcceptanceTimeUpdated.Hook(func(event *clock.TimeUpdateEvent) {
-		fmt.Printf("%s > [%s] Clock.AcceptanceTimeUpdated: %s\n", n.Name, engineName, event.NewTime)
+	events.Clock.AcceptedTimeUpdated.Hook(func(newTime time.Time) {
+		fmt.Printf("%s > [%s] Clock.AcceptanceTimeUpdated: %s\n", n.Name, engineName, newTime)
```
alexsporn: update printf to match the new event name

packages/protocol/engine/clock/events.go
hmoog: I don't think that is necessary as hooking to the event of the clock already implies having access to the entire clock object in the closure anyway and we always only react to the updated time value anyway (as the relative time advances constantly and would rather require some tick callback):

```
clock.myEvent.Hook(func(...) {
    // do sth with clock despite not being in the even param
   clock.RelativeTime()
})
```

packages/protocol/engine/clock/blocktime/clock.go
hmoog: It's not the workerpool that is called async but the "option" to "execute things asyncronously in a workerpool"

I personally think it makes the code more readable to name the variable in such a qualitative way rather than "workerPoolOption" or sth. which would be more aligned with the low-level concepts behind the "async hooking / execution" of the callbacks.

alexsporn: The problem I have with `async` is that this name is a qualifier in lots of languages. A variable name should imho explain what it contains. The only variables where I would really name them more explaining the behaviour than the content are boolean flags.

alexsporn (approved):

PR #2608, opened 2023-03-10
Fixes for ConflictDAG Locking, UTXO Commitment Divergence, and Acceptance Propagation 

piotrm50:
# Description of change

This PR contains fixes for:
* prevents creation of subjectively invalid blocks by locking the ConflictDAG during block creation, forking and adding a new vote.
* prevents commitment divergence on UTXOs by correctly setting the InclusionSlot to which an output is notarized.
* fixed acceptance propagation to the past cone of a marker. Currently acceptance/confirmation is not propagated past weak/like parents, but those parents are accepted. This requires having a separate `weaklyAccepted` flag, to later allow correctly propagating acceptance through a block using causal orderer. 

## Type of change

## Change checklist


packages/protocol/engine/ledger/utxoledger/statediffs.go
```diff
@@ -288,10 +288,14 @@ func (s *StateDiffs) storeTransaction(transaction utxo.Transaction, metadata *me
 	return nil
 }
 
-func (s *StateDiffs) outputWithMetadata(outputID utxo.OutputID) (outputWithMetadata *mempool.OutputWithMetadata) {
+func (s *StateDiffs) outputWithMetadata(outputID utxo.OutputID, index ...slot.Index) (outputWithMetadata *mempool.OutputWithMetadata) {
```
karimodm: Do we need this initializer then? We could just maybe set the index from the outside on the created object. It is just a very specific usecase now: and probably not needed as the outputs metadata should be already set.

piotrm50: Done

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/iotaledger/hive.go/app v0.0.0-20230301135914-603c4dac79e1
 	github.com/iotaledger/hive.go/autopeering v0.0.0-20230301135914-603c4dac79e1
 	github.com/iotaledger/hive.go/constraints v0.0.0-20230301135914-603c4dac79e1
-	github.com/iotaledger/hive.go/core v1.0.0-rc.3.0.20230301135914-603c4dac79e1
+	github.com/iotaledger/hive.go/core v1.0.0-rc.3.0.20230307163144-1809be6dde56
```
alexsporn: Please do a PR on hive.go with that commit (`feat/causal-parents-callback`)

packages/app/blockissuer/blockfactory/referenceprovider.go
```diff
@@ -39,6 +39,9 @@ func (r *ReferenceProvider) References(payload payload.Payload, strongParents mo
 
 	excludedConflictIDs := utxo.NewTransactionIDs()
 
+	r.protocol.Engine().Ledger.MemPool().ConflictDAG().WeightsMutex.Lock()
```
alexsporn, line 42: Can we solve this differently? Locking this from the outside is not such a good idea

packages/protocol/engine/ledger/mempool/conflictdag/conflictdag.go
```diff
@@ -24,6 +25,8 @@ type ConflictDAG[ConflictIDType, ResourceIDType comparable] struct {
 	// mutex is a mutex that prevents that two processes simultaneously update the ConflictDAG.
 	mutex *syncutils.StarvingMutex
 
+	WeightsMutex sync.RWMutex
```
alexsporn, line 32: This mutex is very weird. It says it is locking access to weights but is not used in this pkg but locked/unlocked from the outside.

packages/app/blockissuer/blockfactory/referenceprovider.go
piotrm50: It's because conflict tracking is done across multiple different components, and putting the lock in the ConflictDAG was easiest for access because the components that use the lock can access the shared lock. It  will be soon refactored anyway.

packages/protocol/engine/ledger/mempool/conflictdag/conflictdag.go
piotrm50: Yes, it will be refactored along with the whole ConflictDAG/ConflictTracker/ConflictResolver rework.

alexsporn (approved):

PR #2614, opened 2023-03-16
Modularize notarization

alexsporn:


packages/protocol/engine/sybilprotection/dpos/sybilprotection.go
```diff
@@ -102,6 +102,8 @@ func (s *SybilProtection) initializeActiveValidators() {
 	}); err != nil {
 		panic(err)
 	}
+
+	s.TriggerInitialized()
```
jonastheis, line 106: if this is in here, maybe rename the function to `initialize` ?

packages/protocol/engine/engine.go
```diff
@@ -104,13 +103,12 @@ func New(
 			e.HookInitialized(lo.Batch(
 				e.Storage.Settings.TriggerInitialized,
 				e.Storage.Commitments.TriggerInitialized,
-				e.NotarizationManager.TriggerInitialized,
+				e.Notarization.TriggerInitialized,
```
jonastheis: Why is this triggered here and not in the module itself?

packages/core/snapshotcreator/snapshotcreator.go
```diff
@@ -104,7 +106,7 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 	if err != nil {
 		return err
 	}
-	if _, _, err = engineInstance.NotarizationManager.Attestations.Commit(0); err != nil {
+	if _, _, err = engineInstance.Notarization.Attestations().(*slotnotarization.Attestations).Commit(0); err != nil {
```
jonastheis, line 109: Why not add these to the interface? To keep it lean as it's only needed here and in some tests?

packages/core/snapshotcreator/snapshotcreator.go
```diff
@@ -114,7 +116,7 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 }
 
 func (m *Options) attest(engineInstance *engine.Engine, nodePublicKey ed25519.PublicKey) error {
-	if _, err := engineInstance.NotarizationManager.Attestations.Add(&notarization.Attestation{
+	if _, err := engineInstance.Notarization.Attestations().(*slotnotarization.Attestations).Add(&notarization.Attestation{
```
jonastheis, line 119: Why not add these to the interface? To keep it lean as it's only needed here and in some tests?

jonastheis (approved):

packages/core/snapshotcreator/snapshotcreator.go
alexsporn: Yes. The setters are currently only used in tests and during snapshot creation.

packages/core/snapshotcreator/snapshotcreator.go
alexsporn: Yes, see other comment above.

packages/protocol/engine/sybilprotection/dpos/sybilprotection.go
alexsporn: That file has 3 initialize* doing different setup steps, so not sure it will make it clearer to rename it

PR #2616, opened 2023-03-22
Modularize Tangle

alexsporn:


packages/protocol/engine/consensus/blockgadget/tresholdblockgadget/gadget.go
```diff
@@ -58,25 +64,27 @@ type Gadget struct {
...
+		e.SybilProtection.HookInitialized(func() {
+			g.Initialize(e.SlotTimeProvider(), e.SybilProtection.Validators(), e.SybilProtection.Weights().TotalWeightWithoutZeroIdentity)
 		})
 
 		return g
 	})
 }
 
-func New(workers *workerpool.Group, tangle *tangle.Tangle, evictionState *eviction.State, opts ...options.Option[Gadget]) *Gadget {
+func New(workers *workerpool.Group, booker booker.Booker, blockDAG blockdag.BlockDAG, memPool mempool.MemPool, evictionState *eviction.State, opts ...options.Option[Gadget]) *Gadget {
```
karimodm, line 77: Good untangling of the Tangle :+1:

plugins/protocol/plugin.go
```diff
@@ -135,7 +135,7 @@ func configureLogging(plugin *node.Plugin) {
 	// 	Plugin.LogDebugf("Block %s booked", block.ID())
 	// }))
 	//
-	// deps.Protocol.Events.Engine.Tangle.VirtualVoting.BlockTracked.Attach(event.NewClosure(func(block *virtualvoting.Block) {
+	// deps.Protocol.Events.Engine.Tangle.VirtualVoting.BlockTracked.Attach(event.NewClosure(func(block *booker.Block) {
```
karimodm, line 138: +1 for refactoring comments ;)

karimodm (approved):
I have skimmed through it all but focused on bits that were not renames: looks great!
One thing that popped to the eye: the `Attestations` type is still in the notarization implementation slotnotarization, while pretty much every other type is sitting next to the interface. Should we move this type as well? It is pretty much fundamental as we commit to it and we send it back and forth to prove fork weights.

alexsporn:
> I have skimmed through it all but focused on bits that were not renames: looks great!
> One thing that popped to the eye: the `Attestations` type is still in the notarization implementation slotnotarization, while pretty much every other type is sitting next to the interface. Should we move this type as well? It is pretty much fundamental as we commit to it and we send it back and forth to prove fork weights.

We don't really pass this object around. `Attestations` is actually a wrapper around the storage. Currently the notarization interface has an `Attestations` interface that provides a getter to fetch certain attestations as a set (this set is what we actually pass around). Maybe it makes more sense to try to break down stuff into more interfaces and move types implementing logic (not just data containers) into the concrete implementations

PR #2618, opened 2023-03-22
Start node from live snapshot

karimodm:
# Description of change

* Snapshot creation WebAPI.
* Correctly set ChainID when starting from snapshot.
* Make ChainManager evictable and do not process commitments below "rootCommitment".
* Introduced rootCommitment to determine earliest forkable and processing point when loading a snapshot.


packages/protocol/protocol.go
```diff
@@ -202,7 +203,21 @@ func (p *Protocol) initNetworkEvents() {
 }
 
 func (p *Protocol) initChainManager() {
-	p.chainManager = chainmanager.NewManager(p.Engine().Storage.Settings.LatestCommitment(), p.optsChainManagerOptions...)
+	genesisCommitment := p.Engine().Storage.Settings.LatestCommitment()
+	p.chainManager = chainmanager.NewManager(genesisCommitment, p.optsChainManagerOptions...)
```
jonastheis: This is useless and I think we should change the `ChainManager`'s constructor to not receive this commitment but instead rename `InitRootCommitment` to `Initialize` and set up everything properly. This will already be in line with future modularization of the component

packages/protocol/protocol.go
```diff
@@ -202,7 +203,21 @@ func (p *Protocol) initNetworkEvents() {
...
+
+	p.Engine().HookInitialized(func() {
+		earliestRootCommitment := p.Engine().EvictionState.EarliestRootCommitment()
```
jonastheis, line 217: Maybe we could add a comment here why this is necessary: "We need to make sure that no block can reference an older commitment than the root commitment. Otherwise the chain might seem not solid and the corresponding blocks can never be processed. We effectively achieve this through the fact that block's commitments need to be monotonically increasing (or the same) and the fact that root blocks are the only blocks that can be referenced."

packages/protocol/protocol.go
```diff
@@ -202,7 +203,21 @@ func (p *Protocol) initNetworkEvents() {
...
+		rootCommitment, err := p.Engine().Storage.Commitments.Load(earliestRootCommitment.Index())
+		if err != nil {
+			panic(fmt.Sprintln("could not load earliest commitment after engine initialization", err))
+		}
+		if err := p.Engine().Storage.Settings.SetChainID(rootCommitment.ID()); err != nil {
```
jonastheis, line 224: Similarly, would be good to add a comment here as to why we set the chainID to the rootcommitment and its implications.

packages/protocol/protocol.go
```diff
@@ -212,9 +227,12 @@ func (p *Protocol) initChainManager() {
...
 		p.chainManager.CommitmentRequester.EvictUntil(index)
-	}, event.WithWorkerPool(wp))
-	p.Events.Engine.EvictionState.SlotEvicted.Hook(func(index slot.Index) {
-		p.chainManager.Evict(index)
+		newRootCommitment, err := p.Engine().Storage.Commitments.Load(index)
+		if err != nil {
+			panic(fmt.Sprintln("could not load latest confirmed commitment", err))
+		}
+		p.chainManager.SetRootCommitment(newRootCommitment)
+		p.chainManager.EvictUntil(index - 1)
```
jonastheis, line 249: Add a comment why we evict until `index - 1`

packages/protocol/protocol.go
```diff
@@ -212,9 +227,12 @@ func (p *Protocol) initChainManager() {
 	}, event.WithWorkerPool(wp))
 	p.Events.Engine.Consensus.SlotGadget.SlotConfirmed.Hook(func(index slot.Index) {
 		p.chainManager.CommitmentRequester.EvictUntil(index)
-	}, event.WithWorkerPool(wp))
-	p.Events.Engine.EvictionState.SlotEvicted.Hook(func(index slot.Index) {
-		p.chainManager.Evict(index)
+		newRootCommitment, err := p.Engine().Storage.Commitments.Load(index)
```
jonastheis, line 239: Does this mean we always load it from disk? The commitment should also be in the chainmanager no? Can't we just load it from there?

packages/protocol/engine/engine.go
```diff
@@ -217,8 +217,6 @@ func (e *Engine) Import(reader io.ReadSeeker) (err error) {
 		return errors.Wrap(err, "failed to import settings")
 	} else if err = e.Storage.Commitments.Import(reader); err != nil {
 		return errors.Wrap(err, "failed to import commitments")
-	} else if err = e.Storage.Settings.SetChainID(e.Storage.Settings.LatestCommitment().ID()); err != nil {
-		return errors.Wrap(err, "failed to set chainID")
```
jonastheis, line 220: Does not setting the `ChainID` on export have any unforeseen consequences? We set it on node startup and probably when we switch the chain right?

packages/protocol/engine/eviction/state.go
```diff
@@ -76,6 +78,21 @@ func (s *State) LastEvictedSlot() slot.Index {
 	return s.lastEvictedSlot
 }
 
+// EarliestRootCommitment returns the earliest commitment across all rootblocks.
+func (s *State) EarliestRootCommitment() (earliestCommitment commitment.ID) {
+	s.rootBlocks.ForEach(func(index slot.Index, storage *shrinkingmap.ShrinkingMap[models.BlockID, commitment.ID]) {
+		storage.ForEach(func(id models.BlockID, commitmentID commitment.ID) bool {
+			if commitmentID.Index() < earliestCommitment.Index() {
```
jonastheis: Isn't this always false? `earliestCommitment.Index()` is the empty value `0` and any `commitmentID.Index()` is at least `0` as well

packages/protocol/engine/eviction/state.go
```diff
@@ -165,35 +186,22 @@ func (s *State) Export(writer io.WriteSeeker, evictedSlot slot.Index) (err error
...
 
-		s.AddRootBlock(rootBlockID)
+		s.AddRootBlock(rootBlockID, commitmentID)
 
 		return nil
 	})
 }
 
-// importRootBlocksFromStorage imports the root blocks from the storage into the cache.
-func (s *State) importRootBlocksFromStorage() (importedBlocks int) {
```
jonastheis, line 181: While this function is currently unused, it might come in handy for our next PR where we want to look at starting a node from disk. So maybe leave it?

packages/protocol/chainmanager/testframework.go
```diff
@@ -120,26 +161,31 @@ func (t *TestFramework) AssertChainState(chains map[string]string) {
...
-
-		for _, commitmentAlias := range commitmentAliases {
-			chainCommitment := chain.Commitment(t.SlotIndex(commitmentAlias))
-
-			require.NotNil(t.test, chainCommitment)
-			require.EqualValues(t.test, t.SlotCommitment(commitmentAlias), chainCommitment.ID())
-			require.EqualValues(t.test, t.commitment(commitmentAlias), chainCommitment.Commitment())
-		}
-	}
+	// for chainAlias, commitmentAliases := range commitmentsByChainAlias {
```
jonastheis: Maybe remove?

packages/protocol/protocol.go
jonastheis: Can this lead to a race condition where we update the root commitment and only evict later? In this order I don't think so, just want to make sure though. If so, we should add a comment that we first need to set the rootcommitment and only then can evict.

packages/protocol/chainmanager/manager.go
```diff
@@ -41,32 +41,180 @@ type Manager struct {
...
+		return
+	}
+
+	for currentIndex := m.lastEvictedSlot + 1; currentIndex <= index; currentIndex++ {
+		m.evict(currentIndex)
+	}
+	m.lastEvictedSlot = index
+}
+
+func (m *Manager) InitRootCommitment(c *commitment.Commitment) {
```
jonastheis: Maybe rename to `Initialize`

jonastheis (changes requested):
I think we should add a test to verify the correct chainID on startup/snapshot loading from a snapshot which does not start at genesis.

packages/protocol/engine/engine.go
alexsporn: This way we start with the same chain that is inside the snapshot we used. So this change is actually correct imho

packages/protocol/engine/eviction/state.go
```diff
@@ -145,11 +167,15 @@ func (s *State) LatestRootBlocks() models.BlockIDs {
...
 	return stream.WriteCollection(writer, func() (elementsCount uint64, err error) {
 		for currentSlot := s.delayedBlockEvictionThreshold(evictedSlot) + 1; currentSlot <= evictedSlot; currentSlot++ {
-			if err = s.storage.RootBlocks.Stream(currentSlot, func(rootBlockID models.BlockID) (err error) {
+			if err = s.storage.RootBlocks.Stream(currentSlot, func(rootBlockID models.BlockID, commitmentID commitment.ID) (err error) {
 				if err = stream.WriteSerializable(writer, rootBlockID, models.BlockIDLength); err != nil {
 					return errors.Wrapf(err, "failed to write root block %s", rootBlockID)
 				}
 
+				if err = stream.WriteSerializable(writer, commitmentID, commitmentID.Length()); err != nil {
+					return errors.Wrapf(err, "failed to write root block %s", rootBlockID)
```
alexsporn: maybe add a different error message in this case?

packages/protocol/engine/eviction/state.go
```diff
@@ -165,35 +191,22 @@ func (s *State) Export(writer io.WriteSeeker, evictedSlot slot.Index) (err error
...
 func (s *State) Import(reader io.ReadSeeker) (err error) {
 	var rootBlockID models.BlockID
+	var commitmentID commitment.ID
 
 	return stream.ReadCollection(reader, func(i int) error {
 		if err = stream.ReadSerializable(reader, &rootBlockID, models.BlockIDLength); err != nil {
 			return errors.Wrapf(err, "failed to read root block id %d", i)
 		}
+		if err = stream.ReadSerializable(reader, &commitmentID, commitmentID.Length()); err != nil {
+			return errors.Wrapf(err, "failed to read root block id %d", i)
```
alexsporn: maybe add a different error message in this case?

alexsporn (approved):

PR #2620, opened 2023-04-01
Starting node from disk

karimodm:
# Description of change

Node starts from disk!


jonastheis (approved):

packages/protocol/engine/notarization/slotnotarization/manager.go
```diff
@@ -51,6 +50,7 @@ func NewProvider(opts ...options.Option[Manager]) module.Provider[*engine.Engine
 	return module.Provide(func(e *engine.Engine) notarization.Notarization {
 		return options.Apply(&Manager{
 			events:                    notarization.NewEvents(),
+			acceptedTimeFunc:          e.Clock.Accepted().Time,
```
piotrm50: I think this should be put into `e.HookConstructed` or `e.HookConstructed` callback, in case Notarization Manager is initialized before the clock, and clock is not available yet.

piotrm50 (changes requested):

piotrm50 (approved):

